PR #21, opened 2026-03-01
Refactor: pass BatterySettings reference instead of bare float capacity

pookey:
## Summary

- `SensorCollector`, `EnergyFlowCalculator`, and `HistoricalDataStore` now accept a shared `BatterySettings` reference instead of a bare `float` capacity
- Components always read the live value from the shared settings object, eliminating manual capacity propagation in `update_settings()`
- Matches the pattern already used by `DailyViewBuilder` and `GrowattScheduleManager`
- Bumps version to 7.1.1

## Context

Addresses the review feedback on the original SOC >100% fix — instead of manually propagating a float to three components, pass the `BatterySettings` reference so components always read the live value.

## Test plan

- [x] `pytest core/bess/tests/unit/ core/bess/tests/integration/` — 190 passed, 2 skipped
- [x] `black` and `ruff check` clean
- [ ] Deploy to HA install — verify SOC values stay 0-100%

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to BESS Battery Manager will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [7.1.1] - 2026-03-02
+
+### Fixed
+
+- Battery SOC no longer shows impossible values (e.g. 168%) when battery capacity differs from the 30 kWh default. `SensorCollector`, `EnergyFlowCalculator`, and `HistoricalDataStore` were initialised with the default capacity and only received the configured value via manual propagation in `update_settings()`. They now hold a shared `BatterySettings` reference so the configured capacity is always used for SOC-to-SOE conversion.
+
 ## [7.1.0] - 2026-03-01
 
 Thanks to [@pookey](https://github.com/pookey) for contributing this fix (PR #20).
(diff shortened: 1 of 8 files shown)


johanzander (commented):
Thanks for the clear bug report and fix! The diagnosis is spot-on — this is invisible to anyone with the default 30 kWh battery.

One design concern before merging: the fix propagates capacity by directly mutating attributes on three components (`sensor_collector.battery_capacity`, `energy_flow_calculator.battery_capacity`, `historical_store.total_capacity`). This works, but it creates a maintenance hazard — if a fourth component ever gets its own capacity copy, it's easy to miss updating `update_settings()`.

The existing pattern in the codebase avoids this problem: `DailyViewBuilder` and `_schedule_manager` both receive the `BatterySettings` object directly and read `battery_settings.total_capacity` on use, so they automatically see any updates.

Would you be open to applying the same pattern to the three affected components? Specifically:
- `SensorCollector.__init__` and `EnergyFlowCalculator.__init__` could accept a `BatterySettings` reference instead of a bare `float`
- `HistoricalDataStore.__init__` likewise

Then `update_settings()` needs no capacity-propagation logic at all — the components always read the live value.

Happy to discuss if there's a reason the scalar copy was intentional.

johanzander (approved):
Thanks for updating the approach — the shared  reference is much cleaner. No more manual propagation needed and no risk of missing a fourth component in the future. LGTM.

johanzander (approved):
Thanks for updating the approach — the shared BatterySettings reference is much cleaner. No more manual propagation needed and no risk of missing a fourth component in the future. LGTM.

PR #23, opened 2026-03-04
WIP: Add ML energy consumption predictor

pookey:
## Summary

- Standalone ML module (`ml/`) using XGBoost to predict 24h energy consumption at 15-minute resolution
- CLI tool with train, predict, evaluate, baseline, report, and fetch-data commands
- `report` command runs the full pipeline (retrain + predict + weather + history) and generates a timestamped HTML dashboard for tracking model evolution
- Configurable feature flags in `ml_config.yaml` — DOW and daylight features disabled by default until sufficient training data accumulates
- Integrates with BESS battery optimizer via `consumption_strategy: ml_prediction` setting

## Test plan

- [x] `ml/.venv/bin/python -m ml report` completes without errors
- [x] Generated HTML chart renders correctly in browser
- [x] Feature flags (day_of_week, daylight_hours) properly exclude features when disabled
- [x] MAE improves when removing noise features with limited training data
- [ ] Verify prediction integration with battery optimizer end-to-end


johanzander (commented):
This looks great and really promising — the ML module is well-structured, the strategy dispatch pattern is clean, and the direct multi-output prediction approach is the right call. A few things need fixing before merge, ranging from blocking issues to minor polish. See inline comments below.

ml/data_fetcher.py
```diff
@@ -0,0 +1,584 @@
...
+from datetime import date, datetime, timedelta
+from zoneinfo import ZoneInfo
+
+import pandas as pd
+import requests
+
+_LOGGER = logging.getLogger(__name__)
+
+LOCAL_TZ = ZoneInfo("Europe/Stockholm")
+UTC = ZoneInfo("UTC")
```
johanzander: **[Blocking]** Timezone is hardcoded in a module intended to be standalone and reusable. This breaks for UK users (Octopus Energy support was just added in v7.3.0) and anyone else outside Stockholm. The timezone should come from `ml_config.yaml` — e.g. `config["location"]["timezone"]` — and be passed into `fetch_history_context()`.

ml/feature_engineer.py
```diff
@@ -0,0 +1,272 @@
...
+
+def _compute_daylight_hours(
+    target_date: date, latitude: float, longitude: float
+) -> float:
+    """Compute daylight hours for a given date and location using astral."""
+    location = LocationInfo(
+        name="location",
+        region="",
+        timezone="Europe/Stockholm",
+        latitude=latitude,
```
johanzander: Same issue as `data_fetcher.py` — timezone should come from config, not be hardcoded. The `location` section in `ml_config.yaml` already has `latitude`/`longitude`; adding a `timezone` key there and reading it here would fix both this and the coordinate/timezone mismatch in `ml_config.yaml`.

PR #25, opened 2026-03-05
Fix terminal value using last known price instead of daily average

pookey:
## Summary

- When tomorrow's prices aren't yet published (Octopus Agile doesn't publish the next day's import rates until around 16:00), the optimizer falls back to a terminal value to prevent dumping battery charge at end of day.
- Previously the terminal value was calculated as the average of all remaining buy prices for the day. This included tonight's expensive peak periods, inflating the terminal value and causing the optimizer to refuse to discharge during those same peaks — a circular, self-defeating calculation.
- Now uses the last known non-zero price in the horizon (typically the 22:30 off-peak rate, since Octopus hasn't yet published 23:00+). This is a far more accurate proxy for what overnight recharging would cost tomorrow, and allows the optimizer to correctly schedule evening discharge during peak-price periods.

## Test plan

- [ ] Verify that with tomorrow's import prices unavailable (before ~16:00), the battery correctly schedules discharge during evening peak periods
- [ ] Verify that when tomorrow's prices are available, terminal value remains 0.0 (unchanged behaviour — the last-known-price path is not reached)
- [ ] Verify that a full-day run from midnight still behaves correctly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to BESS Battery Manager will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [7.3.1] - 2026-03-05
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


johanzander:
@claude-bot please review PR

PR #29, opened 2026-03-06
Fix chart data positioning, X-axis, and price step function

pookey:
## Summary

- **Remove fake zero-point** in EnergyFlowChart: data is now positioned at period start instead of offset by 1, eliminating the empty data point at index 0 that caused misalignment
- **Fix X-axis rendering**: add `type="number"` with explicit ticks and `interval={0}` for reliable label placement on both charts
- **Price line uses stepAfter**: matches the actual step-function nature of electricity pricing instead of smooth monotone interpolation
- **Replace broken `<rect>` overlay** with Recharts `ReferenceArea` for predicted hours shading (the raw SVG rect never rendered at correct chart coordinates)
- **Normalize tomorrow period numbers**: handle API returning periods as 96-191 (continuation) vs 0-95 (relative)
- **Fix tomorrow data X positioning** to align with period start consistently
- **Fix maxHour calculation** (+1) to give stepAfter room to render the last period
- **Use modulo 24** for tick formatting to show clean hour labels across the day boundary
- **Remove midnight separator lines** (redundant visual noise given the tick labels)

## Test plan

- [ ] Verify energy flow chart shows data starting at hour 00, not offset by 1
- [ ] Verify price line renders as a step function (flat segments) not smooth curves
- [ ] Verify predicted hours shading overlay appears at the correct position
- [ ] Verify X-axis labels show 00-23 (and repeat for tomorrow data)
- [ ] Verify battery level chart data points align to period boundaries
- [ ] Test with and without tomorrow data present
- [ ] Test with quarter-hourly resolution

### DIFF
--- a/frontend/src/components/BatteryLevelChart.tsx
+++ b/frontend/src/components/BatteryLevelChart.tsx
@@ -83,22 +83,20 @@ export const BatteryLevelChart: React.FC<BatteryLevelChartProps> = ({ hourlyData
       throw new Error(`MISSING DATA: batteryAction is required but missing at index ${index}`);
     }
     const batteryAction = getValue(hour.batteryAction);
-    const batterySocPercent = getValue(hour.batterySocEnd); // Extract from FormattedValue
-    const price = getValue(hour.buyPrice); // Extract from FormattedValue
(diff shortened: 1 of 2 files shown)


pookey:
image

PR #61, opened 2026-03-17
Fix IDLE intent blocking battery discharge for home consumption

HesselHam:
IDLE periods had discharge_rate=0, which prevented the battery from discharging to cover home load even while importing from grid. Since IDLE means "no strategic action" (no arbitrage, no forced export), the battery should still naturally discharge in load_first mode.

Set discharge_rate=100 for IDLE so the inverter can draw from the battery to meet home consumption as intended by load_first behavior.


HesselHam:
idle discharge rate 0 > 100

PR #71, opened 2026-04-16
fix: model Growatt hardware solar behavior in DP optimizer

pookey:
## Summary

The DP optimizer's energy model did not match how the Growatt inverter actually handles solar in its operating modes, causing incorrect schedules:

- **battery_first mode**: Hardware absorbs ALL excess solar, but the DP only modeled the requested charge power — allowing it to plan impossible cross-temporal arbitrage (export solar cheap, buy grid expensive)
- **Discharge override**: When solar >= consumption, the Growatt overrides discharge commands and charges from excess solar instead. The DP had no knowledge of this and planned discharge during sunny periods
- **Dashboard mismatch**: `create_decision_data()` derived intent and `batteryAction` from the raw DP power, ignoring hardware overrides — the dashboard showed "Powering Home" with phantom discharge during solar-covered periods

### Changes

- `_state_transition`, `_compute_reward`, `_build_period_data`: three-way branching (charge / effective-discharge / idle-or-override) matching actual Growatt hardware behavior
- `_compute_idle_solar_charging`: shared helper for implicit solar charging (excess solar clamped by charge rate and available capacity)
- `_create_idle_schedule`: propagates SOE with implicit solar charging instead of flat SOE
- `create_decision_data`: derives strategic intent and `battery_action_kwh` from post-override energy flows (`energy_data.battery_charged`/`battery_discharged`) instead of raw DP power
- Updated 5 scenario test expected values
- Added `test_battery_first_solar_charging.py` (8 tests) and `test_idle_solar_charging.py` (16 tests)

## Test plan

- [x] All 277 unit tests pass
- [x] All 50 integration tests pass (3 skipped)
- [ ] Verify on dashboard: during SOLAR_STORAGE periods, excess solar goes to battery (not exported)
- [ ] Verify on dashboard: during periods where solar covers load, no phantom discharge shown
- [ ] Verify schedule bar shows correct mode labels (IDLE not LOAD_SUPPORT when solar covers load)

### DIFF
--- a/core/bess/decision_intelligence.py
+++ b/core/bess/decision_intelligence.py
@@ -443,16 +443,16 @@ def create_decision_data(
     Returns:
         Enhanced DecisionData with all fields populated including advanced flow patterns
     """
-    # Determine strategic intent based on actual energy flows
-    # CRITICAL: Intent controls hardware behavior via Growatt schedule
-    # Must accurately reflect actual action, not just economic conditions
-    if power < -0.1:  # Discharging
-        # Check actual flows: are we exporting to grid?
+    # Determine strategic intent based on actual energy flows (post-override)
+    # CRITICAL: Intent controls hardware behavior via Growatt schedule.
+    # Must use energy_data (which reflects hardware overrides like discharge
+    # suppression when solar >= consumption) rather than raw DP power.
+    if energy_data.battery_discharged > 0.1:  # Effective discharge
         if energy_data.battery_to_grid > 0.1:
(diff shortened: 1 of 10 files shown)


johanzander:
One thing you could do is export and share your BESS debug log. There is a
script that mocks your settings and current state of export so one can run
it and reproduce.

Also useful when changing implementation, where you can replay the exact
past scenario using newer version of the code and inspect changed behavior
for better or worse.

Complements the tests.


tors 16 apr. 2026 kl. 13:31 skrev Ian ***@***.***>:

PR #141, opened 2026-06-17
Stop morning intent-mode dither (min-dwell hysteresis) — investigation + design

johanzander:
## Summary

A user reported the schedule timeline rapidly flipping **Storing Solar** ⇄ **Selling to Grid** all morning. This PR carries the **investigation + design spec**; the code fix follows in an implementation plan. No production code yet.

> Note: branch name (`...-deadband`) predates the conclusion — the dead-band idea was **disproven** (see below). Kept to avoid churn.

## Root cause

`±0.1 kWh/quarter ≈ 0.4 kW`. The optimal control near the SoE floor is *fractional*; the 0.1 kWh DP grid renders it as an alternating ±0.1 pattern whose running average is the fractional target. Each flip crosses the intent classifier threshold → flips intent → churns the inverter `grid_first`/`load_first` mode + flickers the UI. **Not an economic bug** — the trajectory is cost-optimal.

## What was ruled out (measured, not argued)

- **Remove actions ≤0.1 kWh:** catastrophic, −26.4 SEK (savings halved). The ~0.4 kW morning *drain* has the same magnitude as the dither — magnitude can't separate signal from noise.
- **Tie-break in the DP argmax:** no-op — the jagged path is the strict grid optimum, not a tie.
- **Constrain optimizer to ≥ min controllable power (Option A):** *worse* — mode-changes 10 → 14 and −0.41 SEK (coarser control overshoots and corrects with bigger swings).

## Chosen solution: downstream min-dwell hysteresis (Option C)

Stabilize the committed-mode stream between the optimizer and the inverter/UI; **optimizer, reward function, and reported economics untouched**. A mode change commits only when it persists (≥ dwell) or is backed by a clearly-controllable action; noise-level flips inherit the prior mode. Persistence is the discriminator magnitude can't give.

**Measured on the real day:** mode-changes **10 → 4**, savings **byte-identical (62.48 SEK)**, morning stabilizes to a single `grid_first` band then `load_first`. Insensitive to the exact dwell (3 ≡ 4).

## In this PR

- `docs/investigations/intent-dither-morning.md` — corrected analysis + anonymized reproduction log.
- `docs/superpowers/specs/2026-06-18-intent-mode-hysteresis-design.md` — design, grounded parameters, edge cases, validation plan.

## Next

Implementation plan → TDD the `stabilize_intents` function + integration in `battery_system_manager`, validated against the full suite with a byte-identical-economics guard test.


johanzander:
Blocked on #144 (closed-loop savings simulator). This fix changes applied inverter modes, and we have no way today to verify its realized economic impact — that verification capability is #144. Implementation will not start until the simulator can A/B the modes and prove no material savings loss.

johanzander:
Superseded by #145/#146. This PR proposed a downstream **hysteresis** to hide the cosmetic morning intent-dither. Investigation (and the closed-loop simulator, #144) showed the dither was a *symptom* of a real economic bug — the optimizer over-credited solar export it couldn't execute (#145). Fixing the root cause (binary store/export disposition + continuous reconstruction, #146) **dissolves the dither on its own**, so the hysteresis is no longer needed. Closing.

PR #242, opened 2026-07-06
Remove ad hoc DP guardrails in favor of pure backward induction

johanzander:
## Summary

Replaces the DP battery-optimizer's ad hoc profitability floors and whole-day rejection gate with pure backward induction, per Bellman's principle of optimality: `IDLE` is always a feasible action, so the value function's own `max` already makes the hold-vs-discharge call correctly — a separate veto on top of that is redundant at best, actively harmful at worst (issue #234).

Design spec: `docs/superpowers/specs/2026-07-06-dp-bellman-guardrail-removal-design.md`
Implementation plan (12 tasks, including 4 discovered mid-execution): `docs/superpowers/plans/2026-07-06-dp-remove-guardrails.md`

- Removed the `cost_basis` profitability floor and anti-cycling special case from `_compute_reward`'s discharge branch.
- Removed the whole-day `min_action_profit_threshold` rejection gate, replaced with a trivial idle-vs-DP-cost numerical safety net (justified by SoE-grid discretization noise, not economic distrust).
- Removed dead code: the `C` cost-basis-threading grid (#234's root cause — a loop-order bug meant it was never read back), the unreachable "no valid action found" fallback, and `stored_period_data` (already discarded by its sole caller).
- Replaced the continuous-path replay's grid-snapping approximation with a one-step recompute using interpolated `V[t+1,:]` (reduces, doesn't fully eliminate, a small discretization residual — the safety net covers the rest).
- Fixed #240: a discharge overshooting `home_consumption` by a small amount is no longer credited as export revenue (load-first hardware self-throttles and never actually delivers it).

**Two real, unplanned bugs found and fixed along the way:**
- `decision_intelligence.classify_strategic_intent` / `models.py`'s `infer_intent_from_flows` used a `0.1` kWh `BATTERY_EXPORT` threshold, ten times coarser than every other flow check in the same functions (`0.01`). This misclassified small export-only discharges as `LOAD_SUPPORT` (a hardware mode that physically cannot export), causing planned-vs-realized cost gaps up to 18.7 SEK on quarter-hourly fixtures. Fixed both files.
- `_compute_reward`'s own export-credit threshold didn't match the classification threshold after the above fix (found by the final whole-branch review) — reconciled to `0.01` everywhere.

## Validation

- All 26 pinned fixtures: equal-or-better economics except 2 fixtures with an accepted sub-öre residual (documented, R==P-verified), consistent with the design doc's empirical validation.
- `R == P` plan-faithfulness invariant (#145/#240) holds for all 26 fixtures.
- A permanent regression test locks in the property the whole redesign rests on: the DP's own schedule is never worse than an all-IDLE baseline, across every fixture.
- Fast suite: 724 passed, 12 skipped. Slow suite: 26/26 scenario fixtures passing.

## Rollout

Changes nearly every pinned fixture's expected schedule — ships to beta before stable, per the design doc, since it's easy to self-validate against fixtures but hard to validate against real hardware/pricing behavior ahead of time.

`BatterySettings.min_action_profit_threshold` and the corresponding `config.yaml` schema entry are deliberately left in place but unused (backward-compatible for existing installs); full removal is a separate follow-up issue.

## Test plan

- [x] Fast suite (`pytest core/bess -m "not slow"`): 724 passed
- [x] Slow scenario suite (`pytest core/bess/tests/unit/test_scenarios.py -m slow`): 26/26 passed
- [x] `./scripts/quality-check.sh`: clean
- [ ] Beta deployment + real-world validation before promoting to stable


johanzander:
Closing as superseded — this shipped to beta directly as 526656d per the PR's own rollout note, and that beta commit has since been forward-ported into main via #253 (part of the release-workflow migration reconciling beta-only commits). Content is already on main; merging this would be redundant/conflicting.

PR #245, opened 2026-07-07
fix: base DP terminal value on sell price, not buy price

johanzander:
## Summary

- `_calculate_terminal_value`'s "today only" fallback (fires on every hourly re-optimization, since the DP's rolling horizon only extends through today) priced leftover battery charge using the **median buy price**. On contracts where buy and sell diverge a lot (e.g. Belgian dynamic tariffs), this systematically overvalued *holding* charge past a known, better real export price, causing the optimizer to sit on a near-full battery through the evening peak instead of discharging.
- Now uses **median sell price** — the only value of that kWh actually known and realizable at the horizon boundary.
- Also updates the independently-duplicated terminal-value estimate in `simulation/verification.py` for consistency.

## Test plan

- [x] Updated `TestCalculateTerminalValue` in `test_extended_horizon.py`, added a new case (`test_based_on_sell_price_not_buy_price`) asserting the estimate tracks sell price and is strictly less than the old buy-price-based estimate on an asymmetric-spread scenario
- [x] Full fast suite (`pytest -m "not slow"`): 1018 passed
- [x] Relevant slow suites (`test_extended_horizon.py`, `test_terminal_value.py`): 23 passed
- [x] `test_plan_faithfulness.py` (exercises the `verification.py` helper touched here): 5 passed
- [x] `black` / `ruff` clean

Closes #244


johanzander:
Closing without merging — the sell-price swap in this PR is the wrong fix.

Had bess-analyst review it before merge (should have done this before opening the PR in the first place, not after). Verdict: this "fixes" the Belgian case by an arithmetic accident while introducing a genuine full-drain regression for ordinary (Nordic/UK-style) markets.

Mechanism: cycle cost is only ever charged on charging, never on discharge (`_compute_reward`), so a sell-median-based terminal value is structurally biased *below* the ordinary in-horizon export value for roughly half of any normal day — not a Belgium-only artifact. Reproduced empirically: a realistic non-Belgian evening-peak scenario drains the battery to near-empty where the current (buggy-for-Belgium) code correctly holds a substantial reserve.

Replacement approach (arbitrage-consistency cap on the existing buy-based estimate, rather than swapping the input series) will follow in a new issue + PR. See #244 for the original bug report.

PR #251, opened 2026-07-09
fix: cap DP terminal-value estimate at best in-horizon export price (replaces #245)

johanzander:
## Summary
- `_calculate_terminal_value()` priced leftover battery charge at horizon end using only the median buy price, which can exceed the real achievable evening export price on wide buy/sell-spread contracts (Belgian ENTSO-e/Belpex), causing the DP to hold charge through the evening peak instead of exporting (#126, #244).
- Adds an arbitrage-consistency cap: `terminal_value = min(buy_based, sell_cap)`, where `sell_cap` is derived from the best sell price already visible in the DP's own horizon. Self-calibrating — no market-specific threshold.
- Same fix applied to the duplicated estimator in `core/bess/simulation/verification.py` (`realized_under_solar_error`).

A prior attempt (#245) swapped buy price for sell price outright. bess-analyst review found this structurally biased *below* ordinary in-horizon export value for roughly half of any normal day (cycle cost is only ever charged on charging, never discharge), over-draining Nordic-shaped markets — the exact regression this PR's `test_nordic_shaped_market_retains_reserve` test now covers.

## Fix
- `core/bess/battery_system_manager.py`: `_calculate_terminal_value` now takes `sell_prices` and caps `buy_based` at `sell_cap = max(0, max(sell_prices)*efficiency_discharge - cycle_cost)`.
- `core/bess/simulation/verification.py`: identical cap applied to `realized_under_solar_error`'s terminal-value estimate.
- Tests: extended `TestCalculateTerminalValue` (binding-cap and non-binding-cap unit cases with hand-computed numbers), new `TestTerminalValueCapRegression` in `test_terminal_value.py` (Belgian-shaped scenario proving the DP now exports at the real evening peak; Nordic-shaped scenario proving reserve is still retained when the cap doesn't bind).

## Test plan
- [x] `./scripts/quality-check.sh` passes locally (fast suite, Black, Ruff)
- [x] Full fast suite (`pytest -m "not slow"`): 1018 passed
- [x] Full slow suite (`pytest -m slow`): 336 passed, including `test_extended_horizon.py`, `test_terminal_value.py`, `test_plan_faithfulness.py`
- [x] `/code-review` (medium effort, 8 finder angles + verify pass): 2 confirmed cleanup findings (hardcoded test index violating rules.md's "no algorithm-specific boundaries" rule; redundant `max()` recomputation) — both fixed, re-verified green
- [x] Replayed #126/#244's actual reported debug bundle (`bess-debug-2026-07-07-143926.md`, Frank Leysen's real Belpex prices/config/battery state) through the real backend via `mock-run.sh`. Confirmed in logs: `Terminal value: 0.103/kWh (buy_based=0.257, sell_cap=0.103, median_buy=0.313, max_sell=0.150, ...)` — the cap binds exactly as expected on this real data. The resulting schedule shows the battery actively discharging (`LOAD_SUPPORT`, -0.9 kWh/period) through the real 0.15 EUR/kWh evening peak (periods 84-92) instead of sitting `IDLE` through it, which is the exact symptom reported in #126/#244.

Supersedes #245 (closed). Root issue: #244.

Closes #246

### DIFF
--- a/core/bess/battery_system_manager.py
+++ b/core/bess/battery_system_manager.py
@@ -1702,19 +1702,33 @@ class BatterySystemManager:
         return optimization_period, optimization_data
 
     def _calculate_terminal_value(
-        self, buy_prices: list[float], optimization_period: int
+        self,
+        buy_prices: list[float],
+        sell_prices: list[float],
+        optimization_period: int,
     ) -> float:
         """Calculate terminal value per kWh for the DP optimization.
 
         When the horizon already extends past today (i.e. tomorrow's prices are
         included), return 0.0 since the DP has explicit future data. Otherwise,
-        estimate value from the median buy price adjusted for efficiency
-        and cycle cost.
+        estimate value from the median buy price adjusted for efficiency and
+        cycle cost ("avoid tomorrow's purchase"), capped at the best achievable
+        in-horizon export value ("arbitrage-consistency cap").
 
-        Using the median avoids inflating the terminal value with peak prices.
+        The cap is required because cycle cost is only ever charged on
+        charging, never on discharge (see `_compute_reward`): an uncapped
+        buy-median terminal value can exceed the best real, known export price
+        already visible inside today's horizon, which makes the DP hold charge
+        to chase a fictitious future bonus instead of exporting now (#126,
+        #244). The cap is self-calibrating from data the DP already has, so it
+        collapses on wide-spread contracts (e.g. Belgian ENTSO-e/Belpex)
+        without needing a market-specific threshold, and stays inert on
+        ordinary/Nordic-shaped markets where the best in-horizon peak is
+        already above the buy-median estimate (#246, supersedes #245).
 
         Args:
             buy_prices: Full buy price array (from optimization_period onwards)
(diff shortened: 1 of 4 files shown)


Frank-Leysen:
Hi Johan,

This looks like the right fix for what we saw on **9.9.0b9** — full battery, good evening injection prices (~EUR 0.13–0.16/kWh on our Luminus HA sensors), but Energy Flow planning only minimal evening discharge while our HA scripts exported normally to the meter.

Good to see you replayed our **`bess-debug-2026-07-07-143926.md`** bundle and the cap binds as expected (`buy_based=0.257`, `sell_cap=0.103`, evening LOAD_SUPPORT through the real peak). That is exactly the symptom we reported in #126 / #244.

We will install the beta from `bess-manager-beta` as soon as it lands and report back:
- evening discharge/export plan vs actual meter on a sunny Luminus day
- whether Today's Costs / savings still behave sensibly alongside the plan change

Demo mode stays on until we are confident the plan is stable across re-optimizations — but this is the main blocker we had for trusting evening export on wide buy/sell-spread contracts.

Frank

PR #305, opened 2026-07-14
feat: inverter AC output cap — solar-clipping-aware optimization

pookey:
## Problem

On a hybrid inverter with DC-coupled PV, total AC output is capped (5 kW on my Growatt MIN 5000TL-XH) but the array peaks higher (~7 kW). Power above the cap can only be captured by charging the battery DC-side — and only while the battery has room. The optimizer has no concept of this cap: export is an unconstrained energy balance, so a full battery under high solar costs the plan nothing. The battery therefore fills as early as surplus exists (binary STORE + passive IDLE absorption both store surplus), and on strong days the entire above-cap midday peak is clipped — hours of generation flat-lined at 5 kW.

## What this adds (opt-in, default off)

Two new `BatterySettings` fields, plumbed through store/API/settings UI:

- `inverter_ac_capacity_kw` — total inverter AC output cap; **0 (default) disables the feature entirely** — optimizer, controller, and simulator behave exactly as today.
- `inverter_ac_capacity_margin` (default 0.05) — model-side haircut on the cap, compensating for hourly Solcast forecasts flattening sub-period peaks. Never written to hardware.

With the cap set:

- **Clipping is a real loss in the DP.** Per period, solar not stored DC-side is deliverable only up to `cap × (1 − margin) × dt`; the excess is `clipped_solar` with zero credit. Reserving headroom for the above-cap window then falls out of the DP's own economics — no heuristics. `clipped_solar` is reported per period on `EnergyData` and as `clippedSolar` in the dashboard API.
- **A HOLD pseudo-action** joins the search (vectorized backward pass + continuous replay). This is required, not optional: every non-discharge action absorbs surplus today, so a reward change alone cannot express "export the surplus, keep the room". HOLD periods classify as **SOLAR_EXPORT** via the existing flow classifier — no new intent, no `INTENT_TO_*` table changes.
- **Apply side:** with the cap set, SOLAR_EXPORT writes `charge_rate=0` (per-period register on Growatt MIN), which is what physically stops load_first from passively filling the battery. On a genuinely full battery (today's only source of SOLAR_EXPORT) it's a no-op. The shadow-price discharge gate is untouched and orthogonal.
- **Discharge shares the cap:** battery discharge is limited to the AC headroom solar leaves (`cap − min(solar, cap)`) in DP feasibility, replay candidates, and the simulator.

The scalar and vectorized reward paths share the new AC-side flow helper (`_ac_flows` / `ac_flows_grid`), and `_create_idle_schedule` uses it too so the idle-fallback safety net can't credit clipped export. The plan-faithfulness simulator mirrors the semantics.

## Feature-off safety

With the default `0`, `_effective_ac_cap_kwh` returns `None` and every branch reduces algebraically to the current formulas; the HOLD action isn't added to the search and `charge_rate` mapping is unchanged. Full suite passes unmodified (1167 fast + 362 slow).

## Tests

- `core/bess/tests/unit/test_ac_cap_clipping.py` — cap math per disposition, HOLD semantics + SOLAR_EXPORT classification, discharge headroom, charge-rate gate, simulator hold behavior, feature-off equivalences.
- New scenario `synthetic_clear_sky_ac_clipping.json` (96 quarterly periods, 7 kW clear-sky bell on a 5 kW cap): the plan keeps headroom until the above-cap window, clips <0.3 kWh of a 5.3 kWh overflow, and shows SOLAR_EXPORT holds before the window; executing the cap-aware plan beats the cap-blind plan under cap reality, while the passive-absorption hardware default clips 3.6 kWh. Runs through `test_all_scenarios` including its R == P check.

## Known limitations (v1, by design)

- HOLD is all-or-nothing per 15-min period; "charge exactly the overflow at 1.7 kW while exporting 5 kW" isn't expressible under binary-store semantics — the DP approximates by mixing absorb and hold periods. A follow-up could add action-derived charge rates for SOLAR_STORAGE.
- Platforms without per-period charge-rate control (SPH, SolaX native) get a cap-aware plan but HOLD isn't hardware-enforceable there.
- Assumes `charge_rate=0` blocks DC-side PV→battery charging on the MIN (the register limits charge from any source); the 15-min re-optimization loop bounds any deviation.


johanzander:
Hi @pookey, interesting! I was not aware of this limitation (I have a MID-30KTL3-XH). 
Is it because it is a 5000 model (hence 5kW)?

pookey:
> Hi @pookey, interesting! I was not aware of this limitation (I have a MID-30KTL3-XH). Is it because it is a 5000 model (hence 5kW)?

I have a 7.6kw solar array plugged into a 5kw inverter. (as you suspect, 5000 is exactly that, there's a 3000, and a 6000 version too!).

This time of year, there's quite a lot of clipping unless I 'manage' it buy disabling BESS, and force exporting for the first half of the day (until. this PR that is!).

see below for the last week where I was away so not micro-managing my own settings. 

image

johanzander:
Thanks for this — the AC-cap/clipping problem is a real gap for hybrid DC-coupled inverters, and the writeup plus the synthetic clear-sky scenario make the reasoning easy to follow. The HOLD pseudo-action and the shared `_ac_flows` helper are a clean way to express "export the surplus, keep the room" without touching the intent table, and the feature-off equivalence + discharge-headroom sharing across DP/replay/simulator all check out.

Two things I'd like addressed before merge:

**1. IDLE-branch cost basis doesn't get the same clipping discount STORE does.**

In `_compute_reward`'s IDLE branch (`dp_battery_algorithm.py`, ~line 584):

```python
else:  # IDLE — passive solar charging
    battery_charged, _ = _idle_battery_flows(soe, next_soe, battery_settings)
    ...
    solar_opportunity_cost = battery_charged * current_sell_price   # no cap discount
```

The STORE branch a few lines up explicitly discounts this — "absorbing energy that would have been clipped anyway is free" — computing opportunity cost as `export_without_storing - grid_exported` (both AC-cap-aware) instead of a flat `solar_to_battery * sell_price`. IDLE's passive absorption doesn't get that treatment, so under AC-cap conditions it always charges the full sell price against `cost_basis`, even when the absorbed energy would otherwise have clipped (zero real opportunity cost).

This doesn't affect the schedule itself — the reward returned is correctly `_ac_flows`-based in both branches — but it does inflate `cost_basis` (discharge profitability bookkeeping / reporting) specifically in the passive-IDLE + clipping case, which is the opposite of what the PR is trying to fix. `test_ac_cap_clipping.py`'s IDLE test only covers an already-full battery (`battery_charged == 0`, branch is a no-op), so this path isn't exercised. Worth either applying the same discount here or confirming it's intentionally out of scope for v1.

**2. Naming: "capacity" collides with the battery's existing "capacity."**

`inverter_ac_capacity_kw` (and the UI label "Inverter AC Output Cap") uses "capacity" for a power rating (kW), while `BatterySettings.total_capacity` already uses "capacity" for the battery's energy rating (kWh) — same word, two different physical quantities, in the same settings object. Manufacturers spec this number as "rated AC power" / "max output power" (e.g. the MIN 5000TL-XH datasheet), so matching that language would also make the field more recognizable to someone reading their inverter's spec sheet to fill it in.

Suggest:
- `inverter_ac_capacity_kw` → `inverter_max_ac_power_kw`
- `inverter_ac_capacity_margin` → `inverter_ac_power_margin`
- UI label "Inverter AC Output Cap" → "Inverter Max AC Power" (or "Max Output Power")

Thanks again for the contribution and support!

pookey:
Thanks for the review! Both points addressed, and the branch is rebased onto current `main` (85940b7).

**1. IDLE-branch cost basis now gets the same clipping discount as STORE.** Passive absorption under the cap charges cost basis only for the export it actually displaces (`export_without_absorbing - grid_exported`), so energy that would have clipped anyway is free. Covered by a new test (`test_idle_cost_basis_discounts_clipped_absorption`) that exercises exactly the passive-IDLE + clipping path you flagged as untested.

**2. Renamed throughout** (settings, store defaults, API camelCase, frontend, docs, changelog):
- `inverter_ac_capacity_kw` → `inverter_max_ac_power_kw`
- `inverter_ac_capacity_margin` → `inverter_ac_power_margin`
- UI label → "Inverter Max AC Power"

**Bonus from the rebase: the HOLD pseudo-action is gone.** #313/#315 (which landed after this PR was based) added the SOLAR_EXPORT-below-max bypass candidate — power 0 with SoE forced unchanged — in both the backward pass and the continuous replay. With this PR's cap-aware reward functions, that bypass computes exactly HOLD's economics ("export the surplus, keep the room"), and #313 already writes `charge_rate=0` for SOLAR_EXPORT unconditionally, so the hardware side needs no cap gating either. Keeping HOLD would have left a dead candidate that could never win a tie, so the diff is now noticeably smaller: no new action, no `hold` plumbing — the cap-aware reward plus your existing bypass is enough. The whole-day scenario tests (clipping avoided, headroom preserved, cap-blind plan costs more) all still pass.

pookey:
This seems to work great - where it's delaying charging, and also charging from grid to get charged ASAP as export is 12p, and import is <12p - exactly what I'd want.


image
image

johanzander:
Great! Its still a draft, let me know when its ready for review again!

pookey:
This has been behaving as I'd hope for, and I am no longer micro-managing my setup during high-solar output days.

johanzander:
### Code review

Found 2 issues. Neither was flagged in the earlier manual review round, apologies for missing them then.

1. `_calculate_detailed_flows` computes `solar_to_home` from raw, pre-cap solar production rather than the AC-cap-limited amount, while `solar_to_grid` on the very next lines does subtract `clipped_solar`.

https://github.com/johanzander/bess-manager/blob/9877a3326e8858e9b2f0ab4246a8aed967995b00/core/bess/models.py#L106-L116

If clipping and home-serving ever coexist in the same period (e.g. 10 kWh solar, 5 kWh AC cap, 6 kWh home load → `clipped_solar=5`, actual AC output capped at 5), this attributes 6 kWh to `solar_to_home` when only 5 kWh could have physically reached the AC stage — home gets over-credited and grid import under-credited for that period. `validate_energy_balance()` only checks top-level totals, so it won't catch a detailed-flow split like this.

2. `EconomicData.from_energy_data`'s `solar_only_cost` baseline (the "solar but no battery" counterfactual) is computed from raw `solar_production`/`home_consumption` with no reference to `clipped_solar`, even though a solar-only system would be subject to the same physical AC cap.

https://github.com/johanzander/bess-manager/blob/9877a3326e8858e9b2f0ab4246a8aed967995b00/core/bess/models.py#L252-L261

With the AC cap enabled, this overstates the baseline's usable solar and inflates `hourly_savings`/`solar_savings` for capped periods — somewhat undercutting the point of a "solar-clipping-aware" feature. Pre-existing method, not a regression from this PR, but the PR is the natural place to extend it since it introduces `clipped_solar`.

PR #319, opened 2026-07-16
fix: extend SOLAR_EXPORT shadow-price discharge gate to SOLAR_STORAGE

johanzander:
## Summary
- `SOLAR_STORAGE` periods hardcoded `discharge_rate=0` unconditionally, with no economic gate — any intra-period deficit not covered by forecast solar (e.g. an unforecast EV-charging session) went 100% to grid import even at high SOC where the DP's own economics (`shadow_price`) would allow covering it from the battery.
- Generalized the `SOLAR_EXPORT` gate from #187 (renamed `solar_export_discharge_rate` → `intra_period_discharge_gate`, already intent-agnostic) and broadened its scope in `_apply_period_schedule` to also cover `SOLAR_STORAGE`.

## Root cause
`core/bess/inverter_controller.py`'s `INTENT_TO_CONTROL["SOLAR_STORAGE"]` and `_map_intent_to_rates` hardcode `discharge_rate=0` for `SOLAR_STORAGE` with no gate — unlike `SOLAR_EXPORT`, which is gated via the shadow-price formula in `_apply_period_schedule`, applied only when `strategic_intent == "SOLAR_EXPORT"`. `shadow_price` is already computed unconditionally for every period regardless of intent (`dp_battery_algorithm.py`), so no DP-side change was needed — the gap was purely in which intents the hardware-write gate covered.

## Fix
- `core/bess/battery_system_manager.py`: broadened the gate condition to `if strategic_intent in ("SOLAR_EXPORT", "SOLAR_STORAGE"):`, reusing the same shadow-price/buy-price/discharge-efficiency lookup. Renamed the gate function to `intra_period_discharge_gate` since it now serves two intents.
- New `core/bess/tests/unit/test_solar_storage_discharge_gate.py`: unit tests exercising `_apply_period_schedule` directly (the actual production hardware-write path) with a populated schedule store, proving the gate now applies to `SOLAR_STORAGE`; plus two `@pytest.mark.slow` DP-scenario tests mirroring the `SOLAR_EXPORT` gate test pattern (high-shadow-price/early-accumulation → gate stays closed; low-opportunity-value → gate opens).
- Ruled out reintroducing the #261 stale-register bug: `solax_modbus_growatt_controller.py` writes `discharge_rate` unconditionally post-#261 — this fix only changes the *value* computed upstream, not the write-skip logic #261 removed. Cross-platform tests (`test_cross_platform.py`, including `test_solar_storage_no_grid_charge`) all still pass unaffected.

## Test plan
- [x] `./scripts/quality-check.sh` passes locally (Python: 1159 passed, 14 skipped, Black/Ruff clean; frontend lint/test failures are a pre-existing environment gap — `frontend/node_modules` was never installed in this worktree — unrelated to this Python-only diff)
- [x] Full slow suite (`.venv/bin/pytest -m slow`) passes: 360 passed, 0 failures, including cross-platform and both new SOLAR_STORAGE gate scenario tests
- [x] `/code-review` run at medium effort against the diff: zero CONFIRMED and zero PLAUSIBLE findings
- [x] Directly observed the fix via a before/after script driving the real `_apply_period_schedule` hardware-write path (SOLAR_STORAGE period, SOC well above min, low shadow price): pre-fix wrote `discharge_rate=0` unconditionally; post-fix correctly wrote `discharge_rate=100`, covering the dip from the battery instead of forcing grid import

Closes #318


johanzander:
Added integration-level test parity: the original #187 \`SOLAR_EXPORT\` gate only ever had function-level + DP-scenario tests, never a direct test of \`_apply_period_schedule\` itself (the actual hardware-write path). Added \`TestSolarExportDischargeGate\` mirroring the new \`TestSolarStorageDischargeGate\` class, so both intents now have the same three layers of coverage: gate-boundary, DP-scenario, and BSM-integration.

Full suite re-verified after this addition: 1162 fast passed (+3), 360 slow passed, 0 failures.

PR #331, opened 2026-07-17
fix: debug export dropped TOU-segment and VPP entities from the replay snapshot

johanzander:
## Summary

Found while validating the mock `ha_statistics` fix (#327) against Frank's real debug log — replay logged `Unknown entity requested: select.pv_growatt_time_3_end` and similar for every TOU segment slot beyond what happened to already be captured.

- **`_serialize_entity_snapshot()` only walked `METHOD_SENSOR_MAP`** — a curated subset of sensor keys used for the Settings/Health UI display. Entities resolved through other paths (`_get_entity_for_service`, `_get_raw_state` — used by `read_tou_segments_from_entities()`, the Growatt VPP register reads, and native SolaX's VPP entities) were never captured. On replay, the mock silently returned `"unavailable"` for the initial hardware-detected TOU/VPP state.
- Every one of those resolution paths ultimately looks up the same active sensor map (`_resolve_entity_id` → `self.sensors[sensor_key]`), so rather than adding a platform-specific special case (per-controller override hooks — tried and reverted, see commit history), the snapshot now captures **every entity in the active sensor map directly**. This covers TOU segments and VPP registers (Growatt-via-solax_modbus and native SolaX) with one general fix, and automatically covers any future entity-reading code path too.
- While validating this against a real production export, found a second, related gap: `controller.sensors` is a one-time copy taken at process startup, only refreshed if a settings PATCH happens to hit the `sensors` branch specifically — so it can go stale relative to the persisted config (e.g. after a platform switch) without a restart. The exporter now accepts a freshly-computed `active_sensors` (via `SettingsStore.get_active_sensors()`), passed in by both call sites (`backend/api.py`, `backend/ai_chat.py`), falling back to `controller.sensors` if not provided.
- While generating a **fresh** real debug export to verify the above, found a third, unrelated bug: `debug_log_parser.py`'s `_SECTION_HEALTH_STATUS` matched `"## System Health Status"`, but the real header (`debug_report_formatter.py`) is `"## System Health Status (point-in-time snapshot at export)"`. The exact-string section scan never matched it, so it never advanced past `"## System Information"` — the Health Status JSON block that follows was silently captured as if it were System Information, overwriting the real `system_info` dict (and its `timezone` field) on **every single debug log**. This explains the persistent "No timezone in log" warning `from_debug_log.py` printed on every real export, and caused `mock-run.sh` replays to fall back to local server/UTC time instead of the installation's real timezone — observed directly as the optimizer starting from the wrong period (`period 92 (23:00)` instead of the scenario's actual `period 4 (01:00)`) and Nordpool price lookups failing for the wrong calendar day.

## Test plan

- [x] `black`/`ruff` — clean
- [x] `pytest -m "not slow"` — 1192 passed, 14 skipped
- [x] New unit tests: `test_debug_data_exporter.py::TestSerializeEntitySnapshot` (entity capture beyond `METHOD_SENSOR_MAP`, no-controller edge case, one entity's fetch failure not dropping others, `active_sensors` priority over stale `controller.sensors`) and `test_debug_log_parser.py::test_health_status_header_does_not_overwrite_system_info`
- [x] Verified against a **live production system** (real HA, not mock): switched the active inverter platform to `solax_modbus_growatt_min` (both `inverter.platform` and `sensors.platform`), triggered a fresh debug export, confirmed the exporter correctly attempted every entity in the newly-active sensor map — capturing real ones (200 OK) and gracefully skipping genuinely nonexistent ones (404, logged as a warning, not a crash). Restored original settings (`growatt_server_min`) immediately after.
- [x] Verified the timezone fix end-to-end: exported a fresh real debug log, confirmed `timezone` now resolves correctly (`Europe/Stockholm`), regenerated the mock scenario, and confirmed `mock-run.sh` now starts at the correct period (`period 4 (01:00)`, matching the scenario) with no price-lookup failure, and the `ha_statistics` consumption forecast now covers all 24 hours (previously only hours 0–15 in an older log that predated the exact-fidelity capture feature)


johanzander:
Updated after review (thanks for the second look): merged in `main`, which now includes #332 (`refresh_active_sensors()` — the correct, general fix for `controller.sensors` staleness across the setup wizard, settings PATCH, and platform-switch paths) and #333 (`SettingsStore` relocated to `core/bess/`).

With #332 landed, `controller.sensors` is now kept fresh at the source, so this PR's own `active_sensors` fallback plumbing (threaded through `DebugDataAggregator` as a workaround) was redundant — it was solving a problem that no longer exists upstream. Dropped that parameter, both call-site wirings in `backend/api.py`/`backend/ai_chat.py`, and the now-obsolete staleness regression test. `_serialize_entity_snapshot()` now reads `controller.sensors` directly.

Scope is now just the two originally-necessary fixes: capturing the full active sensor map (TOU/VPP entities) instead of the curated `METHOD_SENSOR_MAP` subset, and the `## System Health Status` header-parser bug. `black`/`ruff`/`pytest -m "not slow"` all green (1191 passed, 14 skipped).

Issue #118, opened 2026-06-10
Correction to Solax-Growatt integration: Growatt MIN-inverters now have functional VPP registers, and do not need TOU programming


nholmgaard:
... whereas the EMS registers and the total energy counters, which are automatically detected inserted, are by default disabled by the Solax-Growatt integration. I do not know if they can be programmatically enabled?

This would make it easier to install and take up the integration, which looks very promising indeed!

johanzander:
I was not aware that growatt-min could support VPP, but indeed promising. 
I just finished a long beta period with a few ppl that tried out the TOU method.

There is theoretical support for solax modbus, which has never been tested though, so should be fairly straightforward.

Would you be open to contribute or test it?

nholmgaard:
I made a (very small) contribution to get VPP functional, and can confirm that it works just fine on my MIN3600TL-XH via wired Modbus. 

Yes, I am trying BESS out right now, and will be happy to report back. I have gotten the EMS registers enabled and all green in the Settings, but no real usage - still showing "Initializing system". I will take a look tomorrow and let you know.

johanzander:
Great!

There is an export debug log function in BESS which is very handy for debugging!

nholmgaard:
I have some problems getting BESS to sell energy from the battery. The intent switch seems well intended, but does not cause the battery to discharge - Grid First is apparently not enough. Did the beta testers have batteries? I took over manually this evening using VPP to ensure that the battery was selling energy in the peak hours. I can provide scripts for controlling the battery via VPP, if you are interested.

johanzander:
I have only had two beta testers and they both have batteries but don't know if they have explicitly verified exporting works or was happy with the arbitrage.

So please export debug logs from Bess which will give me your configuration and the VPP commands and I can have a look at it

johanzander:
FYI: https://github.com/user-attachments/files/18301858/2.1.GROWATT.VPP.COMMUNICATION.PROTOCOL.OF.INVERTER_V2.01.pdf

johanzander:
@jdungen  - you use growatt modbus solax right? would you be interested in taking a stab at this one?

jdungen:
Yes for sure. But I use tuo with the 3 options load,grid and battery.

johanzander:
I guess this approach would make the tou usage for growatt solax redundant, and make the setup easier. Do you see a benefit with that?

jdungen:
[...]

PR #356, opened 2026-07-20
fix: Growatt VPP mode blocks passive solar charging during SOLAR_EXPORT

johanzander:
## Summary
- `SolaxModbusGrowattController` in `control_mode="vpp"` mapped `SOLAR_EXPORT` to the same command as `SOLAR_STORAGE`/`IDLE` (remote control disabled, falling back to native `load_first` self-use), because VPP hardware has no `charge_rate` register — the channel register-based platforms use to block passive solar charging for `SOLAR_EXPORT` specifically.
- Adds a `block_passive_charging` signal, computed once from strategic intent and threaded through `InverterController.apply_period`. `SOLAR_EXPORT` on VPP platforms now keeps `vpp_remote_control` enabled with `vpp_power=0` — the documented `grid first` state per the Growatt VPP protocol (V2.01, §3.5) — instead of disabling remote control into self-use.
- Register-based platforms (TOU/cloud/SPH) are unaffected. `SolaxController` (real SolaX hardware) accepts the new signal but is deliberately left unchanged — no equivalent SolaX vendor-protocol verification was done.

## Root cause
On the morning of 2026-07-20, the battery discharged fully (06:30–08:15, correctly, high sell prices), then a `SOLAR_EXPORT` period at 08:45 sent 100% of that period's solar into the battery instead of exporting it (remote control was disabled, handing control to the inverter's native self-use logic), and the optimizer correctly re-discharged the refilled energy at 09:00 — a legitimate decision given the SOE it inherited, but one that shouldn't have been necessary. Full diagnosis and design: [`docs/superpowers/specs/2026-07-20-vpp-passive-charge-block-design.md`](https://github.com/johanzander/bess-manager/blob/fix/issue-355-vpp-solar-export-hold/docs/superpowers/specs/2026-07-20-vpp-passive-charge-block-design.md).

## Fix
`compute_rates_for_period` now also returns `block_passive_charging` (derived from the existing `INTENT_TO_CONTROL["charge_rate"]==0`, no new table), threaded through `apply_period` → `_write_period_to_hardware`/`_apply_period_vpp`. Register-based controllers ignore it (unchanged). `SolaxModbusGrowattController._intent_to_vpp` uses it to select `grid_first` (`vpp_power=0`, remote control enabled) instead of self-use (`remote control disabled`) specifically for `SOLAR_EXPORT`.

As a side effect, this also fixes the same class of bug on the discharge-inhibit path: an inhibited `LOAD_SUPPORT`/`BATTERY_EXPORT` period (external sensor forcing `discharge_rate=0`) now correctly holds the battery instead of falling back to self-use mid-inhibit.

## Test plan
- [x] `./scripts/quality-check.sh` passes locally (fast suite: 1226 passed, 14 skipped; Black/Ruff/TS/ESLint clean)
- [x] `.venv/bin/pytest -m slow` passes (384 passed, 3 skipped, no flakes)
- [x] Code review: no CONFIRMED findings
- [x] Verified via mock-HA replay of the reporting user's actual debug bundle (`bess-debug-2026-07-20-090503.md`, downloaded from the issue attachment): confirmed via real service-call traces to mock-HA that a `SOLAR_EXPORT` period now sends `select.growatt_inverter_vpp_remote_control -> Enabled` + `number.vpp_bess_power_wrapper -> 0`, instead of the previous `Disabled` (self-use)

## Documentation
Updated `docs/INVERTER_PLATFORMS.md`, `docs/SOFTWARE_DESIGN.md`, and `docs/agents/bess-knowledge.md` (fixed a now-stale claim that predated this fix — "SOLAR_EXPORT writes `charge_rate=0` ... all configurations" — no longer true for VPP mode, which has no such register). Added a direct link to the official Growatt VPP protocol PDF (linked from #118) as the authoritative register reference for future VPP work.

## Not yet validated
Whether the inverter's `grid first` priority mode, entered via a forced `vpp_power=0` command, holds the battery exactly flat (vs. the only previously-exercised case, a nonzero forced `BATTERY_EXPORT` target) is a firmware behavior claim the register table's mode-selection logic doesn't fully prove on its own. Ships experimental pending confirmation from @ridax67 (already running the beta channel) via a normal debug export from a morning with solar surplus after a `SOLAR_EXPORT` period.

Relates to #355 — not auto-closing; closes once hardware-validated.


ridax67:
It has worked fine today  both morning and evening dumps.

Issue #381, opened 2026-07-24
feature: computed overnight reserve floor (Min SOC + historical night use) so one knob covers evening export + night load


Frank-Leysen:
## Context

Live Growatt MOD 5000TL3-XH / 15 kWh / Belgium Dynamic (Belpex H + Luminus), tester @Frank-Leysen. Related discussion lived on #126; opening a dedicated issue per your request.

Today BESS has a single **Min SOC** knob. In practice we need two different floors:

1. **Hard floor (Min SOC)** — never go below this (hardware / safety / end of overnight).
2. **Evening export / midnight planning floor** — how low evening `BATTERY_EXPORT` may go so enough energy remains for overnight house load until morning PV covers consumption.

With only one knob we keep hitting a bad trade-off (reported on #126, e.g. 17 Jul):

| Min SOC setting | What happens |
|-----------------|--------------|
| Low (e.g. 12–15%) | Strong evening export, but overnight can run too thin → avoidable grid import on Belgian buy≈0.35 vs sell≈0.12–0.16 |
| High (e.g. 47%) | Night reserve OK, but DP often **holds far above** Min SOC through the peak (we saw ~74% flat) → evening arbitrage largely wasted |

So: one Min SOC cannot express *"export hard into the evening peak, but keep X% reserved for overnight load support until solar returns."*

## Observed (today's numbers as example)

- Battery: **15 kWh**
- User Min SOC: **12%** → **1.8 kWh** hard floor
- Overnight battery use (last night → this morning, until PV starts covering load): **~2.8 kWh**
- Implied planning reserve after evening export: **1.8 + 2.8 = 4.6 kWh ≈ 30.7% SOC**

Separately, before tomorrow's day-ahead prices land (~14:00), the dashboard Net Grid Cost can look overly optimistic because the plan still assumes export down toward Min SOC; after prices, midnight SOC often rises (overnight reserve) and the headline shrinks. A visible computed floor would make that behaviour easier to understand even when the horizon is incomplete.

## Desired

Keep **one user setting: Min SOC** (hard floor).

Have BESS **compute** an overnight planning floor from history, e.g. last 7 nights:

```
planning_floor_kWh = min_soc_kWh + expected_overnight_battery_use_kWh (+ optional safety margin)
planning_floor_soc = planning_floor_kWh / capacity
```

Use `planning_floor_soc` in:

- evening `BATTERY_EXPORT` stop / midnight target
[...]

PR #385, opened 2026-07-24
fix: LOAD_SUPPORT discharge ceiling leaks real-time load spikes to grid

johanzander:
## Summary
- `LOAD_SUPPORT` and `BATTERY_EXPORT` shared one discharge-rate-ceiling branch, capping `LOAD_SUPPORT` at the DP's planned average power even though `load_first` is already self-limiting to the real house deficit
- Extends the existing SOLAR_EXPORT/SOLAR_STORAGE shadow-price gate (`intra_period_discharge_gate`) to `LOAD_SUPPORT`, opening the ceiling above the plan only when the DP's own marginal valuation says the reserve isn't needed elsewhere
- `BATTERY_EXPORT` is unchanged (no physical deficit backstop, so it still needs the hard cap)

## Root cause
`_map_intent_to_rates` (`core/bess/inverter_controller.py:479-486`) capped the discharge-rate register at the DP's planned *average* power for the period, for both `LOAD_SUPPORT` and `BATTERY_EXPORT`. That's correct for `BATTERY_EXPORT` (`grid_first` has no deficit backstop — an open ceiling would oversell beyond the arbitrage plan), but `LOAD_SUPPORT` runs through `load_first`, which is already self-limiting to the real house deficit — opening its ceiling can never cause it to export or overshoot beyond actual need, only to spend reserve covering a real spike.

Confirmed against live debug bundles reported on #381 (Frank-Leysen, Growatt MOD 5000TL3-XH via `solax_modbus_growatt_min`, TOU mode): small recurring overnight grid imports (0.1-0.2 kWh, a few tens of cents/night) occurred while SOE had 3+ kWh of headroom above `min_soe_kwh`, in the same 15-minute period the battery was actively discharging under `LOAD_SUPPORT` at exactly its written ceiling.

## Fix
`LOAD_SUPPORT` now reuses `intra_period_discharge_gate` (the same shadow-price economic gate already used for `SOLAR_EXPORT`/`SOLAR_STORAGE`), opening the discharge ceiling above the plan-scaled baseline when `buy_price × efficiency_discharge >= shadow_price` — i.e. when the DP's own marginal valuation says the reserve isn't earmarked for a later, more valuable use.

Important difference from `SOLAR_EXPORT`/`SOLAR_STORAGE`: their planned baseline is always 0, so the gate result fully determines the outcome. `LOAD_SUPPORT` already plans a nonzero rate (deliberate partial discharge, per #147), so the gate result is combined as `discharge_rate = max(planned_rate, gate_result)` — it can only **raise** the ceiling, never lower an already-committed plan. Lowering it would reopen the #147 regression (a hardcoded 100% `LOAD_SUPPORT` rate drained the battery early on high-consumption days, blowing past reserve planned for a later, pricier period).

**Known limitation, documented in `docs/agents/bess-knowledge.md`:** replaying real captured data from Frank's debug bundle through the actual DP showed that during a sustained overnight `LOAD_SUPPORT` drawdown, `shadow_price` sits within a cent or two of `buy_price` for the *entire* stretch — that near-tie is close to the defining condition for choosing `LOAD_SUPPORT` at all, so the strict gate inequality rarely opens in that specific regime. This fix does not eliminate every small overnight import Frank reported; that residual is tracked separately as its own issue, #393 (see PR discussion below — the earlier draft here cited #276, which review found to be closed and unrelated). Follow-up on #393 is the intended path for that remaining symptom, not a change to this gate.

## Test plan
- [x] `./scripts/quality-check.sh` passes locally (black/ruff clean; frontend 103/103)
- [x] `.venv/bin/pytest -m "not slow"` passes locally (1312 passed, 15 skipped)
- [x] `.venv/bin/pytest -m slow` passes locally (385 passed, 3 skipped)
- [x] New test file `core/bess/tests/unit/test_load_support_discharge_gate.py` — 4 tests: gate opens on favorable shadow price, baseline preserved (never zeroed) on unfavorable shadow price, no-stored-schedule keeps baseline, `BATTERY_EXPORT` unaffected
- [x] Plan-faithfulness (`R == P`) coverage via #388/#389, now merged into this branch: `test_run_scenario_realized_wires_shadow_price_gate_by_default` (`core/bess/tests/unit/test_simulator_shadow_price_gate.py`) drives a real DP-optimized LOAD_SUPPORT scenario through `run_scenario_realized` end-to-end and asserts the gate captures value the DP's plan couldn't — closing the gap review flagged as missing (see PR discussion)
- [x] Verified against real production data: replayed Frank's actual captured buy/sell/solar/consumption history (from his debug bundle's "Full Historical Data JSON") through the real `optimize_battery_schedule` + the fixed `_apply_period_schedule` path, confirming the gate fires correctly (and confirming, honestly, that it does *not* open for Frank's specific 3 leaked periods — see Known Limitation above)
- [x] Code review: no CONFIRMED blocking findings; one cosmetic docstring staleness fixed inline

Related: #381 (original report), #147 (must not regress), #393 (tracked follow-up for the residual near-tie symptom), #387 (sensor-collector measurement-lag fix, relevant to the disputed period-6 worked example — see PR discussion), #388 (plan-faithfulness gate-modeling fix, provides the R==P test above)

Closes #384


johanzander:
### Code review

Investigated the root-cause hypothesis (actuals vs. predictions) and the fix itself. Findings below — most importantly, the worked example this PR cites as evidence has a numeric inconsistency that should be resolved before merge.

**1. Worked example's "16% ceiling" was never actually applied to hardware — undermines the cited evidence**

The period-6 example (2026-07-22 01:30–01:45, from Frank's debug bundle) is built around a discharge-rate of 16% being the ceiling the battery hit. That 16% figure comes from a *later* optimization run's re-displayed schedule table for an already-elapsed period — it was never written to the inverter. The authoritative apply-time log for the run that actually executed period 6 shows:

```
2026-07-22 01:30:00 | battery_system_manager:2595 - Period 6 (01:30): Intent=LOAD_SUPPORT, Action=-0.11 kWh (-0.45 kW), DischargeRate=9%
2026-07-22 01:30:00 | battery_system_manager:2611 - HARDWARE: Setting discharge power rate to 9% for period 6
```

At this system's actual rating (5 kW, not the code's 15 kW default), 9% implies a ceiling of ~0.1125 kWh for the period — but the measured actual discharge was 0.2 kWh, ~1.8x that ceiling, and confirmed as genuine energy delivery (not a rounding/counter artifact). A hard power cap can't be exceeded by ~80%, so the "battery discharging at exactly its written ceiling" framing doesn't hold up numerically for this specific instance. Either the register-to-power mapping is wrong for this inverter, or `load_first` on this platform doesn't behave as the hard ceiling the code's docstring and this PR assume. The architectural diagnosis (LOAD_SUPPORT sharing a too-tight ceiling with BATTERY_EXPORT) may still be right, but this worked example shouldn't be cited as confirming it until the 9%-vs-0.2kWh gap is explained.

Also worth noting: the bundle's telemetry is 15-minute-period aggregate only, and the one instance with fine-enough timestamps to inspect coincides with two API-retry warnings in the log (`ha_api_controller:898`, 01:30:30 and 01:31:16 — the only two retries all day) — a control-layer confound that isn't ruled out by the evidence presented.

**2. Missing REQUIRED plan-faithfulness verification** (`docs/agents/simulator.md` says this is required for any change to control/rate mapping)

[docs/agents/simulator.md](https://github.com/johanzander/bess-manager/blob/f0e8fcf56c4fbb2d2d173388d7309bde4ad941fa/docs/agents/simulator.md#L20-L25) states changes to `inverter_controller.py`/rate mapping MUST verify `R == P` via `run_scenario_realized`/`verify_plan_faithfulness`. This PR's new tests only call `_apply_period_schedule` directly with hand-built stub arguments:

https://github.com/johanzander/bess-manager/blob/f0e8fcf56c4fbb2d2d173388d7309bde4ad941fa/core/bess/battery_system_manager.py#L2544-L2568

No existing test file was modified, and none of the ~30 existing files referencing LOAD_SUPPORT reach this new branch under any current fixture — confirmed by checking each candidate call chain (DP/optimizer tests never call `_apply_period_schedule`; TOU-scheduling tests go through a different layer; SolaX tests are structurally excluded via `discharge_rate_is_load_following=False`; the one close case never populates `schedule_store`). That means the new branch is currently only verified by synthetic inputs, not by any DP-produced schedule — exactly the gap `simulator.md` exists to catch.

**3. Stale issue citation for the documented residual limitation**

The CHANGELOG/`docs/agents/bess-knowledge.md` update attributes the residual near-tie shadow-price limitation to #276, but #276 is closed and about an unrelated topic (SOE-grid discretization/interpolation error from #275/#279, concluded "financially optimal," not a tracked bug). Worth citing a different/new issue instead.

johanzander:
### Follow-up: possible measurement-lag artifact in the period-6 worked example (unproven, flagging for awareness)

Digging into why the applied 9% ceiling (~0.1125 kWh) and the reported `battery_discharged=0.2 kWh` for period 6 don't reconcile (see prior comment), I checked the SOE-derived ground truth against the counter-based `battery_discharged` reading across the full night in the bundle.

**Pattern observed:** `battery_discharged` is a diff of live HA-polled cumulative counters in runtime mode (not pinned to the exact period boundary the way historical backfill is), and 3 times in this one night, a period reads `battery_discharged=0.000` immediately followed by a period reading roughly double the expected value — consistent with one period's counter tick landing on the wrong side of a poll boundary and bleeding into the next:

| periods | discharged(AC) | SOE-implied (2-period) |
|---|---|---|
| 5→6 | 0.00 + 0.20 = 0.20 | ~0.285 |
| 9→10 | 0.00 + 0.20 = 0.20 | ~0.1425 |
| 25→26 | 0.00 + 0.30 = 0.30 | ~0.285 |

Aggregate over the full 27-period night: SOE-implied discharge totals 2.708 kWh vs. reported `battery_discharged` totals 2.800 kWh (~3.4% diff) — within the noise band of two independent quantization grids (SOE steps at 1%/0.15 kWh, discharge counter at 0.1 kWh).

**This is not proven** — 5 of 8 zero-discharge periods in the same window do *not* show a doubled follow-up, and SOE's own 1% resolution isn't fine enough to fully confirm the mechanism rather than coincidence. I'm flagging it as a hypothesis worth checking, not a confirmed root cause: if it holds up, the period-6 worked example this PR cites (0.2 kWh actual vs. a 9% applied ceiling) may be partly or wholly a `sensor_collector.py` runtime live-poll diffing artifact rather than evidence of the discharge-ceiling gap. Would be worth checking whether this pattern holds across other users'/nights' debug bundles before leaning on this specific example as primary evidence.

johanzander:
### Tracking: two prerequisite issues opened

To resolve the open questions above (worked-example numeric inconsistency, missing plan-faithfulness coverage), two underlying gaps have been split into their own tracked issues so this PR isn't blocked on carrying the full investigation:

- #387 — `sensor_collector.py` runtime-mode measurement lag (the likely explanation for the period-6 worked example's discrepancy)
- #388 — plan-faithfulness simulator doesn't model the shadow-price discharge gate (blocks writing a real `R == P` test for this PR's change)

Plan: resolve #387 and #388 first (neither needs new field data), then re-derive this PR's worked example on clean data and add the now-possible plan-faithfulness test before merge. If the clean data shows the gap doesn't hold up, this PR's root-cause framing will need revisiting.

johanzander:
### Follow-up: #387 and #388 are merged; here's where each review point stands

**Point 2 (missing R==P plan-faithfulness test) — resolved.** #388/#389 merged (into this branch, not yet `main`) and fixed exactly this gap: the simulator's `derive_control_command`/`_map_rates` now calls the real `intra_period_discharge_gate`. Its own test suite (`core/bess/tests/unit/test_simulator_shadow_price_gate.py`, now on this branch) includes `test_run_scenario_realized_wires_shadow_price_gate_by_default`, which drives a **real DP-optimized LOAD_SUPPORT scenario** through `run_scenario_realized` end-to-end (not a hand-built stub) and asserts realized cost lands below planned cost by the gate-captured margin. This reaches the gate via the same code path `run_scenario_realized` drives in production, satisfying `docs/agents/testing.md`'s plan-faithfulness requirement for this PR's control-mapping change. No new test needed on top of what #389 already added.

**Point 3 (stale #276 citation) — fixed.** #276 is closed and unrelated (DP grid-discretization/interpolation error from #275, concluded financially optimal per #275/#276/#285 — see PR #286). Opened #393 to track the near-tie residual specifically and repointed the CHANGELOG, `docs/agents/bess-knowledge.md`, and this PR's description at it instead.

**Point 1 (disputed period-6 worked example) — cannot be conclusively re-derived, and I don't think it needs to be.** #387/#390's fix corrects sensor collection *going forward only* (a new live per-minute power-sample buffer feeding runtime gap-fill) — it has no mechanism to retroactively reprocess Frank's already-captured historical bundle. So the specific 9%-ceiling-vs-0.2kWh-actual discrepancy in that one period can't be re-run through the fixed code and confirmed clean or artifactual after the fact; it stays an open question tied to data that's now stale.

That said, this PR's actual evidence doesn't depend on that specific number. The DP-replay test in the Test Plan replays Frank's captured **price/consumption/solar history** (unaffected by the counter-quantization bug — that bug only touched `battery_discharged` energy readings, not price/load/solar inputs) through the real `optimize_battery_schedule` + `_apply_period_schedule` path, and independently confirms the gate mechanism behaves as designed, including correctly *not* opening for Frank's 3 leaked periods (the near-tie regime now tracked in #393). Combined with the architectural diagnosis (`LOAD_SUPPORT`'s `load_first` self-limiting backstop vs. `BATTERY_EXPORT`'s lack of one) and #389's new synthetic R==P test, I think the fix stands on evidence that doesn't route through the disputed historical number — the worked example was illustrative context for the original report, not load-bearing proof for the fix itself.

### Recommendation: merge

All three review points are addressed, quality-check/fast/slow suites are green on this branch with #387+#388 merged in. Flagging for your call rather than merging myself, per the release workflow rules — let me know if you want anything more before it goes in.

Issue #403, opened 2026-07-26
Quarterly schedule-update tick can be silently skipped when hardware retries run long, permanently dropping a period's actuals


johanzander:
## Summary

The quarterly schedule-update job (`app.py` `update_schedule_quarterly`, `CronTrigger(minute="0,15,30,45")`) can silently miss its scheduled run — and permanently drop a whole 15-minute period's actuals — when the previous run is still busy retrying a slow inverter hardware write when the next tick fires.

## Evidence (production, 2026-07-27, beta b25)

1. Add-on auto-updated b24→b25 at 00:26:58, restarting the container mid-period-1 (00:15-00:30).
2. On restart (00:27:06), a health-check safety net (`battery_system_manager.py:2946`, "No schedule exists yet... retrying the initial schedule build") fired an early schedule build for period 1, immediately followed by `start()`'s own normal call to the same period. Each hit the pre-existing `number/set_value` Supervisor timeout issue and retried with backoff.
3. Between the two overlapping calls, the "apply period 1 to hardware" step stayed busy from **00:27:06 to 00:31:56** (log: `Applied period settings for period 1 (00:15)` at both 00:29:33 and 00:31:56).
4. The job is registered with `misfire_grace_time=30` (`app.py:337-341`) and default `max_instances=1`. The 00:30:00 tick (which would run as "period 2" and record period 1 as its just-completed previous period) landed inside that busy window, missed its 30s grace, and was coalesced away by APScheduler. There is **no** `Updating battery schedule for period 2` log line anywhere in that day's log — direct proof it never ran.
5. The next tick to run was 00:45:00, as "period 3" — which only records period 2 as previous. Period 1 (00:15-00:30) was never anyone's "previous period" and was never recorded.
6. Confirmed live via `GET /api/export-debug-data?compact=false`: `historical_periods` has exactly one `null` entry for the day — period 1 — with all others (including period 0, correctly InfluxDB-backfilled at startup) present.
7. User-visible effect: the "Incomplete Historical Data" dashboard banner, persisting for the rest of the day since nothing ever retries a mid-day skipped tick.

## Root cause

- `misfire_grace_time=30` is too tight for a job whose critical section can legitimately take minutes (hardware-write retries with exponential backoff).
[...]

PR #408, opened 2026-07-27
fix: persist HistoricalDataStore to disk via DailyViewStore

johanzander:
## Summary
- `HistoricalDataStore` (today's live sensor/economic actuals) was the only one of four "today" stores with zero disk persistence — a mid-day restart (crash, add-on auto-update) wiped everything collected so far, recoverable only via an InfluxDB backfill limited to periods that had already elapsed before that process started.
- Folds it into the existing `DailyViewStore` as a write-through cache: every tick that collects new actuals also persists the merged `DailyView` to `/data/daily_views/{date}.json`; startup seeds `historical_store` from today's file first, then runs the existing InfluxDB backfill only for whatever's still missing (closing gaps from any cause, including issue #403).
- "Clear Savings History" now excludes today's file (it's a live cache, not completed history); the now-redundant explicit 23:55 `prepare_next_day` save was removed since incremental writes supersede it.
- Includes an exception-safety fix (`_persist_today_view()` is now genuinely best-effort per its docstring/spec) and a guard so `BESS_HISTORICAL_SEED_FILE` E2E replay mode never persists fixture data.
- Design spec and implementation plan are included under `docs/superpowers/`.

## Test plan
- [x] `.venv/bin/pytest -m "not slow"` — 1372 passed
- [x] `./scripts/quality-check.sh` — 0 errors
- [x] `.venv/bin/pytest -m slow` — all green
- [x] Mutation-verified: the two new integration call sites are proven load-bearing (deleting either fails a specific new test)
- [x] Full 5-task implementation plan executed with per-task review + a final whole-branch review, all findings resolved

Related: #403 (silently-skipped scheduler tick — this PR makes its data-loss symptom recoverable but doesn't fix the scheduler root cause, tracked separately).

### DIFF
--- a/backend/api.py
+++ b/backend/api.py
@@ -564,8 +564,9 @@ def _aggregate_quarterly_to_hourly(
 async def get_dashboard_available_dates():
     """List ISO dates that have dashboard data available (for date-picker greying).
 
-    Today is always included even though it isn't persisted to the
-    DailyViewStore until day rollover.
+    Today is always included even though it is deliberately excluded from
+    DailyViewStore.list_available_dates() by design — today's view is
+    persisted continuously, but only past days are listed as history.
     """
     from app import bess_controller
 
--- a/core/bess/battery_system_manager.py
+++ b/core/bess/battery_system_manager.py
@@ -1633,6 +1633,8 @@ class BatterySystemManager:
         else:
             logger.info("Historical store: no periods stored yet")
 
+        self._persist_today_view()
+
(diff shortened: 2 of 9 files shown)


johanzander:
Filed the deferred full-consolidation follow-up as #409 (folding `PredictionSnapshotStore` + `ScheduleStore` into the same `DailyView`-per-day format this PR uses for `HistoricalDataStore`).

PR #412, opened 2026-07-28
Add experimental Huawei EMMA + SUN2000 inverter support

valexi7:
Wire native EMMA TOU control, discovery, UI, mock-HA scenario, and tests for the Huawei EMMA + SUN2000 platform.
image


johanzander:
Thanks for putting this together — the EMMA/SUN2000 support is a real gap and the controller design here (subclassing `HuaweiController`, reusing the period-grouping logic) follows the pattern we already use for the other platforms.

Before merging, though, we'd like to reconsider the integration dependency. This PR wires BESS to `huawei_emma_management` (https://github.com/valexi7/Huawei-Modbus-TLS-Server), a new integration and reverse Modbus/TLS proxy. We'd prefer to build EMMA support on top of `huawei_solar` (https://github.com/wlcrs/huawei_solar) instead — the integration BESS's existing Huawei (LUNA2000) support already depends on, and the de-facto standard Huawei integration in the HA community (900+ stars, actively maintained, HACS-listed).

The reason: `huawei_solar` has picked up native EMMA support itself over the last few releases (currently v2.1.1) —

- A `huawei_solar.set_emma_tou_periods` service (`services.py`) that writes TOU periods to a battery behind an EMMA, using the *same* period-string format (`HUAWEI_LUNA2000_TOU_PATTERN`) `HuaweiController._periods_to_text()` already produces — this looks like it could be a small extension of the existing controller rather than a new one.
- EMMA-specific select entities (`EMMA_ESS_CONTROL_MODE`, `EMMA_TOU_PREFERRED_USE_OF_SURPLUS_PV_POWER`) and power-control registers (`EMMA_MAXIMUM_FEED_GRID_POWER_WATT`/`PERCENT`, `EMMA_POWER_CONTROL_MODE_AT_GRID_CONNECTION_POINT`).
- Several EMMA-targeted fixes across the 2.0.x line (capacity-control detection, timestamp computation, etc.), so this isn't a stub — it's had real iteration.
- Plain Modbus TCP straight to the EMMA (port 502) — no separate TLS proxy or config-entry plumbing needed.

Going this route would mean one Huawei integration for BESS to depend on instead of two, no new third-party project in the trust chain, and a smaller diff (extend `HuaweiController`/`HUAWEI_SUFFIX_MAP` rather than a new controller, suffix map, and config-entry-id threading through settings/discovery).

We haven't verified `set_emma_tou_periods` against a real SUN2000+EMMA installation — if you've already tested against real hardware and hit a wall with `huawei_solar` that pushed you toward the custom integration, we'd like to hear about it, since that'd change the calculus. Otherwise, would you be open to reworking this PR on top of `huawei_solar`?

PR #414, opened 2026-07-28
fix: Growatt VPP mode releases control for LOAD_SUPPORT instead of forcing a fixed rate

johanzander:
## Summary
- `SolaxModbusGrowattController`'s VPP control mode (`control_mode="vpp"`) mapped `LOAD_SUPPORT` identically to `BATTERY_EXPORT` — a fixed, forced discharge percentage via `grid_first` — instead of following actual household load.
- `LOAD_SUPPORT` now releases VPP control entirely (`vpp_remote_control` disabled), falling back to the inverter's own load-following self-consumption — the VPP-mode equivalent of TOU mode's existing `LOAD_SUPPORT → load_first` mapping. `BATTERY_EXPORT` is unaffected.

## Root cause
By the time execution reaches `_intent_to_vpp()`, the `LOAD_SUPPORT` vs `BATTERY_EXPORT` distinction was already gone — `_map_intent_to_rates()` collapses both intents into the same `(grid_charge=False, discharge_rate=X)` tuple, and `apply_period()`'s signature only ever received `grid_charge`/`discharge_rate`/`block_passive_charging`. This is the same shape of problem `block_passive_charging` was added to solve for `SOLAR_EXPORT` in #355.

Reported independently by two real-hardware testers (nholmgaard, ridax67) testing Growatt MIN inverters on vanilla b26 in #118:

> Load support intention commands the discharge rate to a fixed value instead of having it just follow the load. This causes unnecessary imports as well as exports... I think that Load support intention should just release VPP control, then it will work better. — @nholmgaard

## Fix
Added a new `strategic_intent` parameter to `apply_period()` (base `InverterController` + `SolaxModbusGrowattController`), following the same threading pattern as the existing `block_passive_charging` param — carried through `BatterySystemManager`'s three call sites (`_apply_period_schedule`, the period-write retry path including its recursive retry, and `apply_discharge_inhibit`). `_intent_to_vpp()` now returns `(0, False)` — release control — when `strategic_intent == "LOAD_SUPPORT"` and `discharge_rate > 0`, instead of `(-discharge_rate, True)`.

Two pre-existing tests in `test_discharge_inhibit_vpp_routing.py` (an unrelated #324 routing regression test) used `LOAD_SUPPORT` as their example intent to exercise a forced-rate scenario; since `LOAD_SUPPORT` no longer forces any rate, they were switched to `BATTERY_EXPORT` (which still forces a rate) so they keep testing what they were meant to test. A new test covers `LOAD_SUPPORT` staying released regardless of discharge-inhibit toggling.

Docs updated: `docs/SOFTWARE_DESIGN.md`'s VPP-mode paragraph and `docs/INVERTER_PLATFORMS.md`'s VPP intent-mapping table/semantics, both of which described the old forced-rate mapping.

## Test plan
- [x] `./scripts/quality-check.sh` passes locally (fast suite: 1369 passed, 15 skipped; black/ruff clean)
- [x] `.venv/bin/pytest -m slow` passes (391 passed, 3 skipped)
- [x] Code review (adapted, no open PR at review time): 1 CONFIRMED finding (stale "mirrors SolaxController" docstring claim) — fixed
- [x] Local mock-HA E2E verification: brought up the real backend + mock-HA stack against the `ci-growatt-vpp` scenario (faketime-pinned to the scenario's date so the real DP produced a real schedule). Observed the currently-active `LOAD_SUPPORT` period drive the mock inverter's `select.growatt_min_vpp_remote_control` entity to `Disabled` — VPP control released, exactly as intended (previously this would have stayed `Enabled` with a forced discharge percentage)

Closes #413


johanzander:
### Code review

Found 1 issue:

1. `LOAD_SUPPORT` fails to release VPP control whenever `discharge_rate == 0`, contradicting the PR's own stated goal — the `discharge_rate == 0` check in `_intent_to_vpp` (line 328) fires before the `strategic_intent == "LOAD_SUPPORT"` check (line 330), so any `LOAD_SUPPORT` period where the DP plan doesn't call for net discharge (routine — `_map_intent_to_rates` sets `discharge_rate = 0` whenever `battery_action_kw >= -0.01`), or where discharge-inhibit forces the rate to 0, falls into the `return 0, block_passive_charging` branch instead. Since `LOAD_SUPPORT`'s `INTENT_TO_CONTROL["charge_rate"]` is 0 (same as `BATTERY_EXPORT`/`SOLAR_EXPORT`), `block_passive_charging` evaluates `True`, so `remote_control_enabled=True` ("grid_first hold") is returned instead of the intended release. This directly contradicts CHANGELOG.md and `docs/INVERTER_PLATFORMS.md`, both of which state `LOAD_SUPPORT` now disables `vpp_remote_control` unconditionally, and contradicts the new test `test_load_support_stays_released_regardless_of_inhibit`, which never actually exercises a zero/inhibited discharge-rate case (it uses a fixed −3.75 kWh action throughout), so CI doesn't catch the gap.

https://github.com/johanzander/bess-manager/blob/c13859b010eddcb7e864f424203927eabff2864c/core/bess/solax_modbus_growatt_controller.py#L326-L332

johanzander:
Good catch — confirmed and fixed in 4d78650f. Reordered \`_intent_to_vpp\` so the LOAD_SUPPORT release check runs before the \`discharge_rate == 0\` branch, unconditional on rate. Added a regression test (\`test_load_support_releases_vpp_control_at_zero_discharge_rate\`) covering the exact production case (real \`block_passive_charging=True\`, \`discharge_rate=0\`). Fast suite (1370 passed), slow suite (391 passed) both green.

ridax67:
Here is a log file there vpp power is first sent as 0 and then -4 at 20.00

[logs-9 (2).txt](https://github.com/user-attachments/files/30475144/logs-9.2.txt)

johanzander:
Also cleaned up two stale doc/comment spots that still said "LOAD_SUPPORT (rate>0)" — leftover from before the discharge_rate==0 fix. Pushed in 60800751.

PR #437, opened 2026-07-30
feat: accept an external consumption forecast as a time-series entity

### LINKED ISSUE #428: Accept an external consumption forecast as a time-series entity
# Accept an external consumption forecast as a time-series entity

## Problem

BESS optimizes against forecasts. For solar it consumes one from Home Assistant
and does no modelling of its own. For consumption there is no equivalent input —
every available strategy is either a constant or an in-house statistical model.

`_get_consumption_forecast` (`core/bess/battery_system_manager.py:1117-1171`)
offers four strategies:

| Strategy | Source | Shape |
|---|---|---|
| `sensor` | `48h_avg_grid_import` | **scalar**, repeated 96× |
| `fixed` | `default_hourly` setting | constant |
| `influxdb_7d_avg` | InfluxDB history | in-house model |
| `ha_statistics` | HA long-term statistics | in-house model |

The `sensor` strategy looks like an external forecast seam but isn't:
`get_estimated_consumption()` (`core/bess/ha_api_controller.py:1235-1257`) reads
a single scalar state (`"unit": "W"`, `ha_api_controller.py:232-238`), divides
it by 4 and returns `[quarterly_consumption] * 96` — a flat line. It cannot
express a morning peak, an evening peak, or anything else.

Meanwhile `ha_statistics` builds a per-hour-of-day **trimmed mean** that
explicitly drops the maximum sample (`battery_system_manager.py:1370-1385`,
comment: *"trimmed mean for outlier robustness (e.g. EV charging spikes)"*).
That is the right call for a baseline — one EV session should not poison the
19:00 average forever — but nothing puts the load back afterwards. The forecast
is deliberately blind to the largest and most predictable loads in the house.

The result is a steady stream of "can BESS also model X?" requests, where X is a
different load phenomenon each time.

## Reported by

- #130 (@andys1802) — EV charging at 11 kW on a 20 A service. The optimizer
  exports stored energy because the forecast contains no sign of the coming EV
  session. Asked for a "prevent battery export" switch as a workaround.
- #126 (@Frank-Leysen) — midday EV charging on a hot day, plus a request for
  temperature-aware load forecasting from his own weather station
  (`weather.iborsb5`, currently wired only to the disabled temperature-derating
  feature).

Both are the same gap seen from different sides: a known, shaped load that BESS
has no way to be told about.

## Proposed behaviour

Add a consumption forecast strategy that reads a **time series** from a Home
Assistant entity, mirroring how the solar forecast is already consumed.

This is a deliberate scope boundary: **BESS optimizes against forecasts, it does
not build them.** Load modelling — EV sessions, weather-driven aircon,
occupancy, holidays, pool pumps — belongs in Home Assistant, where the user has
the local knowledge and the source entities. BESS never modelled solar
irradiance either. The existing `ha_statistics` / `influxdb_7d_avg` strategies
stay as the batteries-included default for users who won't build a forecaster;
they simply stop being the place every new load phenomenon has to land.

## Prior art in this repo

Three existing shapes, none of them a neutral contract:

- **Prices** (`core/bess/price_manager.py:200-270`) — one entity, both days as
  attributes. Primary path `raw_today` / `raw_tomorrow`: lists of timestamped
  records (`{start, value}`) with timestamp validation
  (`_parse_raw_data_for_date`) and DST handling; fallback to positional
  `today` / `tomorrow` arrays.
- **Solar** (`core/bess/ha_api_controller.py:2124-2180`) — two entities
  (today/tomorrow), each with a `detailedHourly` attribute of
  `{period_start, pv_estimate}` records. Hardcodes Solcast's field names and
  indexes by hour into a 24-slot array, then divides by 4.
- **Consumption** (`sensor`) — scalar, as described above.

Both prices and solar are *per-vendor adapters* rather than a general seam. This
issue is the opportunity to define a neutral one — and unlike Solcast or
Nordpool, the producer here is the user's own template, so we specify the format
rather than adapting to someone else's.

## Recommendation on shape (open for discussion)

Timestamped records, one entity, both days — the price pattern. It is
self-describing on granularity, horizon and day boundary, so it survives
15-minute vs 60-minute sources and a DP horizon longer than 24 h without an
assumed period length.

Note that the solar parser structurally cannot represent sub-hourly data
(`hourly_values[hour] = ...` is last-write-wins per hour). That is correct for
Solcast's `detailedHourly` and is not a bug today, but it is a concrete reason
not to copy the positional/hour-indexed approach for a user-authored series.

## Scope

In scope:
- A new strategy reading a list-valued entity, with normalization to the DP's
  period grid.
- Health-check coverage and a clear failure when the entity is missing,
  stale, malformed or short of the horizon.

Out of scope:
- Any load modelling inside BESS (EV sessions, weather conditioning,
  occupancy). That is the whole point of the seam.
- **Load control or shifting.** BESS accounting for a load is not the same as
  BESS deciding when it runs. Co-optimizing a shiftable load with the battery
  plan is a genuinely larger feature and a separate issue.
- Predicting *unannounced* load. A forecast series expresses "this will
  happen", never "this might happen" — see below.

## Explicit non-solution

This does **not** solve the unpredictable-arrival case in #130: if the user
cannot say in advance that the EV is coming, no forecast input helps. What
covers that is LOAD_SUPPORT following *actual* load rather than the planned
per-period budget (#352, #393) — an execution-layer fix that needs no prediction
at all.

## Open questions

- **Format**: timestamped records vs positional arrays; one entity with
  today/tomorrow attributes vs two entities. (Recommendation above.)
- **Granularity**: accept hourly, quarter-hourly or arbitrary, and normalize?
  Or require a fixed resolution?
- **Horizon**: 96 periods is not enough — the DP has been observed running 118
  periods. How far ahead must the series extend, and what happens when it is
  shorter?
- **New strategy vs upgrading `sensor`**: recommend a distinct strategy rather
  than making `sensor` accept both scalar and list. Type-sniffing the input
  would silently degrade to a flat line when a user's template breaks, which is
  exactly the implicit fallback `rules.md` prohibits.
- **Naming**: the UI currently labels the scalar strategy `consumption_forecast`
  (`frontend/src/components/settings/SensorConfigSection.tsx:95`,
  `frontend/src/lib/sensorDefinitions.ts:508`, "48h Avg Grid Import"). That
  naming is misleading today and will be more so once a real forecast strategy
  exists.
- Should BESS ship a documented example HA template (baseline statistics + an EV
  session) so the seam is usable without the user starting from nothing?

## Verification

Canonical scenario harness (`test_scenarios.py` + `expected_results`): the same
day with and without a shaped evening load in the supplied series, pinning the
SOE trajectory. Confirm the pin discriminates against the pre-fix input.

johanzander:
## Summary

- Adds `ha_consumption_series`: a new consumption-forecast strategy reading a user-authored HA time-series entity (`raw_today`/`raw_tomorrow` attributes of timestamped `{start, value}` records), mirroring the existing price-manager entity pattern rather than solar's hour-indexed parser (which can't represent sub-hourly data).
- BESS still does not model loads itself — no EV/weather/occupancy modelling was added. The user's own HA template or automation builds the shaped series; BESS only normalizes 15- or 60-minute record spacing onto the DP's quarter-hour grid.
- A missing, stale, malformed, or horizon-short series raises an explicit `ConsumptionForecastUnavailableError` — this strategy never silently falls back to a flat profile, unlike `ha_statistics`'s fixed-value fallback on insufficient data.
- Renames the scalar `sensor` strategy's misleading "Consumption Forecast" UI category to "Consumption Average (48h Rolling)" (it was never a real forecast) and adds a new "Consumption Forecast Series" category for the new entity.
- Adds an example HA template (baseline + a scheduled EV block) to `docs/USER_GUIDE.md` so the seam is usable without starting from a blank YAML file.

## Root cause

From the issue: every existing consumption strategy is either a constant or an in-house statistical model. `ha_statistics`'s trimmed mean deliberately discounts one-off spikes (the right call for outlier robustness) but as a side effect is permanently blind to a known, predictable load like an EV session — nothing puts that load back into the forecast. Solar has an external-forecast seam (Solcast); consumption had none.

## Fix

- `HomeAssistantAPIController.get_consumption_forecast_series()` / `get_consumption_forecast_series_tomorrow()` (`core/bess/ha_api_controller.py`) — fetch and normalize the entity, mirroring `price_manager.py`'s `raw_today`/`raw_tomorrow` parsing.
- `BatterySystemManager._get_consumption_forecast` dispatches to it for the new strategy; `_extend_consumption_predictions` (extracted from the existing tomorrow-horizon-extension logic) uses the entity's real `raw_tomorrow` data instead of repeating today's pattern when the DP horizon spans midnight — the same treatment solar's `_fetch_tomorrow_solar_forecast` already gets.
- Health-check coverage (`sensor_collector.check_prediction_health`) and the strategy-comparison endpoint (`get_consumption_forecast_comparison`) both cover the new strategy.
- Frontend: new radio option + description in `HomeFormSection.tsx`, sensor category rename + addition in `sensorDefinitions.ts`, updated `SensorConfigSection.tsx` detection logic, `ConsumptionForecastComparison.tsx` chart labels/colors.

## Test plan

- [x] `./scripts/quality-check.sh` passes locally (fast suite: 1399 passed, 15 skipped; frontend: 103 passed; Black/Ruff/TypeScript/ESLint clean)
- [x] `.venv/bin/pytest -m slow` passes (394 passed, 3 skipped)
- [x] New unit tests: entity parsing (happy path, hourly upsampling, stale/malformed/short-horizon/unsupported-interval failure modes), strategy dispatch + no-silent-fallback, health-check integration, and an end-to-end test that runs a shaped EV-load series through the real DP optimizer and confirms it produces a different (correct) plan than a flat series
- [x] Local E2E verification via `docker-compose.ci.yml` (real backend + mock-HA): configured the new strategy and a shaped entity via the real `PATCH /api/settings` API, confirmed `GET /api/consumption-forecast-comparison` returns the correct shaped hourly profile (EV hour at 16.0 kWh vs 1.2 kWh baseline) through the live system, confirmed the health check reports `OK` with the exact shaped series and then a clear `ERROR` (no silent fallback) when the entity's `raw_today` data was removed, and confirmed the new frontend strings are present in the actual served JS bundle

Closes #428


johanzander:
## Still conflicted — merge attempted twice today, both aborted

This PR is `CONFLICTING` against `main` and has been since ~2026-08-02. Two attempts today, neither landed:

1. A manual `sweep-prs` pass merged `origin/main` and hit **semantic conflicts in 5 files**, so it aborted per that skill's rule against resolving someone else's logic conflicts. Tree left clean.
2. A background session (`pr437-conflicts`) was dispatched to resolve them and produced **nothing** — `HEAD` still `0a6e91f4`, 3 commits ahead, no `MERGE_HEAD`, no tracked file written in 2.5 hours. Stopped. Nothing lost, nothing done.

### The conflicts

`git merge origin/main` conflicts in:

- `core/bess/ha_api_controller.py`
- `core/bess/sensor_collector.py`
- `core/bess/settings_store.py`
- `core/bess/tests/unit/test_prediction_health_check.py`
- `frontend/src/components/settings/HomeFormSection.tsx`

These are logic conflicts, not mechanical ones. Resolving them means preserving **both** this branch's external-consumption-forecast feature and ~2 weeks of `main` drift in the same files — per `docs/agents/rules.md`, with no silent fallbacks or defensive error handling added while reconciling.

### One trap worth knowing before you start

The worktree accumulated **150 untracked files byte-identical to `origin/main`**, which blocked the merge with `The following untracked working tree files would be overwritten by merge` before any conflict was even reported. Git truncates that list at ~4KB, so it cannot be enumerated from the error message — derive it instead:

```bash
comm -12 <(git ls-files --others --exclude-standard | sort) \
         <(git ls-tree -r origin/main --name-only | sort)
```

Then delete only the entries that are byte-identical to `origin/main` (verify with `git show origin/main:$f | cmp -s - "$f"`); the merge restores them. That has already been done once, so it may not recur — but check, because it presents as an unrelated failure.

bess-product-owner:
## Parked: not to be merged yet — awaiting user feedback

Maintainer decision, recorded here so it stops being re-derived every backlog pass. Board fields set to match:

- `Priority: P4`
- `Awaiting: reporter`

Together those suppress this PR from the rhythm pass's action list. It is still counted and listed as deferred, so it stays findable — it is parked, not dropped.

**Do not resolve the conflicts yet.** This PR has been `CONFLICTING` since ~2026-08-02 with semantic conflicts in five files:

- `core/bess/ha_api_controller.py`
- `core/bess/sensor_collector.py`
- `core/bess/settings_store.py`
- `core/bess/tests/unit/test_prediction_health_check.py`
- `frontend/src/components/settings/HomeFormSection.tsx`

Two attempts have already been spent on them — a `sweep-prs` pass that correctly aborted rather than resolve someone else's logic conflicts, and a background session that wrote nothing in 2.5 hours. A third attempt would cost the same and land in the same place, because the blocker is not the merge: the feature's shape needs user feedback first, and resolving conflicts against a design that may still change is wasted work.

Unpark by clearing `Awaiting` once that feedback is in.

PR #490, opened 2026-08-07
Consolidate PredictionSnapshotStore into the unified per-day DailyView format

johanzander:
## Summary
- Folds `PredictionSnapshotStore`'s persistence into `DailyViewStore`'s existing per-day file (`/data/daily_views/{date}.json`), under a new `"snapshots"` key alongside the existing `"view"` key, instead of its own bespoke `/data/bess_prediction_snapshots.json`.
- First of two PRs tracking #409 (follow-up to #408); `ScheduleStore` consolidation is deliberately deferred to a separate PR given its hot-path role in inverter writes (discharge-rate gating, export curtailment) vs. this store's purely display/diagnostic role. Issue #409 stays open until that second PR lands.
- Public API of `PredictionSnapshotStore` is unchanged — no call-site edits beyond the necessary import-path fix for two relocated deserialization helpers.

## Root cause
Per #409 (follow-up to #408's `HistoricalDataStore` persistence work): the "today" store landscape was still fragmented across three persistence formats/locations. `PredictionSnapshotStore` had its own file, discarded at midnight, separate from the `DailyView` schema `HistoricalDataStore`/`DailyViewStore` already share.

## Design
Full design doc: `docs/superpowers/specs/2026-08-06-prediction-snapshot-consolidation-design.md`. Key points:
- Shared `_load_container`/`_write_container` helpers in `daily_view_store.py` do an atomic (temp-file + `os.replace`) read-modify-write, so `DailyViewStore` (touches only `"view"`) and `PredictionSnapshotStore` (touches only `"snapshots"`) never clobber each other.
- Deserialization helpers (`_daily_view_from_dict`, `_period_data_from_dict`) relocated from `prediction_snapshot.py` to `daily_view_builder.py`, which already owns the dataclasses they construct.
- Legacy flat-file fallback: files written before this change (no `"view"`/`"snapshots"` wrapper) still load correctly.
- **Caught in final review, fixed before this PR:** the original design's premise — "a new calendar day is naturally a new file, so the explicit midnight `clear()` call is redundant" — only held for the on-disk side. `PredictionSnapshotStore` is a long-lived process singleton; without a rollover mechanism in the in-memory cache itself, snapshots would have accumulated forever (unbounded growth, stale cross-day data leaking into API responses). Fixed with date-aware lazy rollover (`_ensure_current_day()`, checked before every state-touching method) instead of relying on an external `clear()` call.
- **Also caught in final review:** two writers now sharing one file needed a real lock (not just atomic replace) plus unique temp filenames, to prevent a corruption/lost-update race between concurrent writers (quarterly scheduler tick vs. FastAPI request handler, different threads).

## Test plan
- [x] `./scripts/quality-check.sh` / `.venv/bin/pytest -m "not slow"` passes locally (1537 passed, 13 skipped)
- [x] `.venv/bin/pytest -m slow` passes locally (391 passed, 3 skipped)
- [x] New test coverage: `test_prediction_snapshot_store.py` (round-trip persistence, ordering, nearest-match, day-rollover with a single long-lived instance across a date change, concurrent-write locking with a real `threading.Barrier`-forced interleaving), extended `test_daily_view_store.py` (wrapped container shape, legacy-format fallback), regression test in `test_bsm_settings_and_lifecycle.py` proving the redundant `clear()` call is gone.
- [x] Local E2E verification (mock-HA + real backend via `docker-compose.ci.yml`, `ci-growatt-vpp` scenario, faketime-pinned to the scenario's date): ran a real optimization cycle, confirmed `/data/daily_views/{date}.json` contains both `"view"` and `"snapshots"` keys with real data (96 periods, 2 captured snapshots), confirmed `GET /api/prediction-analysis/snapshots` and `/timeline` return the expected shape, confirmed no legacy `/data/bess_prediction_snapshots.json` file is ever created.

Part of #409 — this PR covers `PredictionSnapshotStore` only. `ScheduleStore` consolidation is tracked as a follow-up PR against the same issue.


johanzander (commented):
Careful pass over the full diff. Eight findings below, posted inline. Severities are my own assessment; #1 is the one I'd treat as a blocker.

**Things I checked and found correct:**
- No lock-ordering cycle between `PredictionSnapshotStore._instance_lock` and the module-level `container_lock` RLock — nothing acquires them in the opposite order.
- `container_transaction` correctly skips the write when the caller's body raises (the exception propagates through the `yield` before `_write_container`).
- The legacy flat-file fallback in `load_day` handles both the pure-legacy shape and the hybrid one (legacy flat view + an injected `snapshots` key).
- The `_period_data_from_dict` / `_daily_view_from_dict` relocation is verbatim, `daily_view_builder.py` already imports all four model types, and the move does break the old `daily_view_store -> prediction_snapshot` import cycle as intended.

core/bess/prediction_snapshot.py
```diff
@@ -195,59 +197,36 @@ def get_snapshot_count(self) -> int:
...
+        path = self._today_path()
         try:
-            self._persist_path.parent.mkdir(parents=True, exist_ok=True)
-
-            # Serialize with default=str to handle datetime and date objects
-            with open(self._persist_path, "w") as f:
-                json.dump(data, f, default=str)
-
+            with container_transaction(path) as container:
+                container["snapshots"] = [asdict(s) for s in self._snapshots]
```
johanzander, line 263: **HIGH — the `BESS_HISTORICAL_SEED_FILE` replay guard is now bypassed.**

`BatterySystemManager._persist_today_view` (`battery_system_manager.py:3315`) returns early when `BESS_HISTORICAL_SEED_FILE` is set, with the comment *"persisting replayed fixture data to /data/daily_views would corrupt real disk state for a test/E2E run."*

`_capture_prediction_snapshot` (`battery_system_manager.py:761`) has no such guard, and after this PR it writes into that exact directory and that exact file.

**Scenario:** run `mock-run.sh` / a seeded E2E on a machine that has real history. Each optimization now does a read-modify-write of the real `/data/daily_views/{date}.json`, injecting fixture snapshots into the user's persisted day. Before this PR the same writes went to the throwaway `/data/bess_prediction_snapshots.json`, so the guard held.

**Fix:** add the same `os.environ.get("BESS_HISTORICAL_SEED_FILE")` check to `_capture_prediction_snapshot` / `_save_to_disk`, or route the snapshot store's persist dir through the same replay switch.

core/bess/prediction_snapshot.py
```diff
@@ -195,59 +197,36 @@ def get_snapshot_count(self) -> int:
...
 
     def _save_to_disk(self) -> None:
-        """Persist snapshots to disk to survive restarts."""
-        data = {
-            "date": time_utils.today().isoformat(),
-            "snapshots": [asdict(s) for s in self._snapshots],
-        }
-
+        """Persist snapshots to today's shared per-day file."""
+        path = self._today_path()
```
johanzander, line 260: **MEDIUM — snapshot retention changes from one day to forever (~100x disk growth).**

Previously snapshots lived in a single file that `_load_from_disk` discarded unless `stored_date == today`, plus an explicit midnight `clear()`. Now they live in `daily_views/{date}.json`, which the module docstring states has "the same forever-retention on disk".

Each `PredictionSnapshot` embeds a **full** `DailyView` (all 96 periods with energy/economic/decision data), and one is captured per optimization (~96/day). So a day's file goes from ~1 DailyView to ~1 + 96 DailyViews.

**Scenario:** on an HA add-on's `/data` (often SD-backed), a year of history goes from tens of MB to low GB. `get_disk_usage()` / an explicit `clear_all()` are the only relief; nothing in the PR bounds the snapshot list or prunes it at day end.

**Fix:** prune the persisted `snapshots` key at day rollover (keep only today), cap the list length, or thin what each snapshot stores rather than embedding a whole `DailyView` per capture.

core/bess/daily_view_store.py
```diff
@@ -36,27 +108,31 @@ def _is_today(self, path: Path) -> bool:
 
     def save_day(self, view: DailyView) -> None:
         """Persist the given day's full view, overwriting any existing file for that date."""
+        path = self._persist_dir / f"{view.date.isoformat()}.json"
         try:
-            self._persist_dir.mkdir(parents=True, exist_ok=True)
-            path = self._persist_dir / f"{view.date.isoformat()}.json"
-            with open(path, "w") as f:
-                json.dump(asdict(view), f, default=str)
+            with container_transaction(path) as container:
```
johanzander, line 195: **MEDIUM — every tick now read-modify-writes the whole snapshot-laden file.**

`save_day` used to `json.dump` the view straight out. It now loads the entire container — which by late afternoon holds dozens of embedded `DailyView`s (see the retention comment on `prediction_snapshot.py`) — and rewrites all of it, on every tick via `_persist_today_view`.

`load_day` has the mirror problem: `savings_aggregator._compute_buckets` (`savings_aggregator.py:173`) calls `store.load_day(day)` for every day in a year bucket, and each call `json.load`s the full file including all snapshots it then discards.

**Scenario:** the savings/history endpoints go from parsing ~365 x 60 KB to ~365 x 6 MB, and the hot tick path does an O(n^2) rewrite over the day.

**Fix:** store snapshots in a sibling file (`{date}.snapshots.json`) so the view path never touches them, or have `load_day` read only the `view` key.

core/bess/daily_view_store.py
```diff
@@ -36,27 +108,31 @@ def _is_today(self, path: Path) -> bool:
...
             return None
+
+        view_dict = container.get("view")
+        if view_dict is None:
+            # Legacy pre-consolidation flat file: the whole container IS the
+            # view dict (written before "view"/"snapshots" wrapper existed).
+            if "date" in container and "periods" in container:
+                view_dict = container
+            else:
+                return None
```
johanzander, line 214: **MEDIUM — a snapshot-only file with no `"view"` key is still reported as an available history date.**

*(This finding concerns `list_available_dates()` at line 140 and `savings_aggregator.py:162-176`, which are outside the diff hunks — anchored here because this `return None` is the branch that produces the symptom.)*

`list_available_dates()` returns every `*.json` stem; `load_day()` hits this `return None` when the container has no `"view"` and no legacy `date`/`periods` keys. The snapshot store can now create exactly such a file — most concretely via the unguarded replay path flagged on `prediction_snapshot.py`, but also on any day where a snapshot is captured while `_persist_today_view` no-ops (it returns early when `schedule_store.get_latest_schedule()` is `None`).

**Scenario:**
- `/api/dashboard/dates` offers the date, and `GET /api/dashboard?date=...` (`api.py:650`) then 404s on it.
- `savings_aggregator` sets `day_count = len(days_in_bucket)` (counting the file) while `totals` skips the `None` view, so the week/month/year average-per-day is silently diluted by a zero day.

**Fix:** have `list_available_dates()` filter to files that actually contain a `"view"` (or a legacy flat view), and/or count only days whose `load_day` returned non-`None` in the aggregator.

core/bess/prediction_snapshot.py
```diff
@@ -139,8 +131,10 @@ def store_snapshot(
             predicted_daily_savings=predicted_daily_savings,
         )
 
-        self._snapshots.append(snapshot)
-        self._save_to_disk()
+        with self._instance_lock:
+            self._ensure_current_day()
+            self._snapshots.append(snapshot)
```
johanzander, line 190: **LOW — a snapshot computed just before midnight can be filed under the next day.**

The snapshot object is built by the caller with `snapshot_timestamp=time_utils.now()` (`battery_system_manager.py:781`), but `_ensure_current_day()` re-reads the clock *after* `_instance_lock` is acquired.

**Scenario:** the optimization at 23:59:5x is slow, or the lock is contended. By the time this block runs, `_ensure_current_day` sees the new date, rolls over, and then appends the day-1 snapshot (day-1 `snapshot_timestamp`, day-1 `daily_view`) into `{day2}.json`. The new `TestConcurrentDayRollover` test only covers the inverse ordering.

**Fix:** anchor the write to `snapshot.snapshot_timestamp.date()` rather than `self._current_date`.

core/bess/prediction_snapshot.py
```diff
@@ -195,59 +197,36 @@ def get_snapshot_count(self) -> int:
...
-                return
-
-            raw_snapshots = data.get("snapshots", [])
-            self._snapshots = [_snapshot_from_dict(s) for s in raw_snapshots]
-            logger.info(
-                "Loaded %d persisted prediction snapshots from disk",
-                len(self._snapshots),
-            )
-        except Exception as e:
+        except (KeyError, ValueError) as e:
```
johanzander: **LOW — exception handling narrowed from `except Exception` to `except (KeyError, ValueError)`.**

`_daily_view_from_dict` does `d["energy"].items()` and iterates `d["periods"]`; a malformed or schema-shifted persisted container yields `AttributeError` / `TypeError`, which are no longer caught.

**Scenario:** `_load_from_disk` is called from `__init__` *and* from `_ensure_current_day()` — i.e. from inside every public accessor. So one bad entry under the `snapshots` key now either aborts `BatterySystemManager` construction or turns `/api/prediction-analysis/snapshots` into a 500, where the old code logged a warning and continued with an empty list.

The same narrowing applies to `_save_to_disk` (`except OSError`, line 215).

**Fix:** keep `except Exception` here (this is best-effort display/diagnostic data), or at minimum add `AttributeError` and `TypeError`.

core/bess/daily_view_store.py
```diff
@@ -2,28 +2,100 @@
...
+
+
+def _cleanup_stale_temp_files(path: Path) -> None:
+    """Best-effort removal of orphaned temp files left by a crashed write.
+
+    Temp files use a unique per-call name, so a failed write leaves a sibling
+    that no *.json glob (housekeeping, disk usage, clear_all) will ever see.
+    """
+    try:
+        for stale in path.parent.glob(f"{path.stem}.*.tmp"):
```
johanzander, line 75: **LOW — unlinks temp files it does not own.**

The glob `{stem}.*.tmp` matches *any* writer's temp file, and `container_lock` is process-local.

**Scenario:** two processes ever touch the same `/data/daily_views` — the add-on plus a mock-run container, or an add-on restart whose shutdown write overlaps the new process. One writer deletes the other's in-flight temp between its `open()` and `os.replace()`, and that `os.replace` then fails with `FileNotFoundError`, which `save_day`'s `except OSError` swallows — silently dropping the write.

**Fix:** restrict cleanup to files older than some age (e.g. mtime > 60s), or to this pid's own prefix.

core/bess/prediction_snapshot.py
```diff
@@ -2,28 +2,28 @@
...
+from core.bess.daily_view_builder import (
+    DailyView,
+    _daily_view_from_dict,
 )
+from core.bess.daily_view_store import container_transaction, read_container
 
 logger = logging.getLogger(__name__)
 
-PERSIST_PATH = Path("/data/bess_prediction_snapshots.json")
+PERSIST_DIR = Path("/data/daily_views")
```
johanzander, line 37: **LOW — no handling of the old persistence file on upgrade.**

`/data/bess_prediction_snapshots.json` (the removed `PERSIST_PATH`) is neither migrated nor deleted.

**Scenario:** on the upgrade tick the user loses the current day's already-captured snapshots — the prediction-analysis endpoints go empty until the next optimization — and the orphaned file stays on `/data` forever, since no code path will ever reference or remove it again.

**Fix:** a one-line best-effort `unlink(missing_ok=True)` of the legacy path at startup (optionally reading its `snapshots` into today's container first if its `date` is today).

bess-agent:
@claude-bot review

johanzander:
Closing unmerged — the design premise is wrong at the root, so this isn't fixable by review rounds.

Consolidating the storage *format* was implemented as consolidating into a single *file*, which forced a shared read-modify-write container, a lock, temp-file cleanup, a legacy fallback, cross-owner pruning, and ultimately ~6 behavioural changes in what was meant to be a maintainability refactor. Two that matter most: snapshot capture is skipped entirely during `BESS_HISTORICAL_SEED_FILE` replay, which empties the prediction-analysis endpoints *and* the debug bundle's snapshot sections (`debug_data_exporter.py:1055`, `:1126`) — a regression in existing functionality; and every view persist now rewrites the whole snapshot blob, roughly doubling daily write volume to ~800 MB/day.

Revised approach recorded in #409. The genuinely useful parts here — the relocated deserialization helpers and much of the new test coverage — will be salvaged into smaller PRs.

Issue #513, opened 2026-08-08
PWL window solver mis-ranks plans: picks 0.584 SEK worse schedule than grid DP under its own objective


johanzander:
## Summary

The PWL window solver — trusted as "exact" and spliced over the grid DP's output in production (#450) — selected a plan **0.584 SEK worse than the grid DP's plan under the DP's own reward objective**, on a plan that was fully inside its feasible action set. Found during the discretization benchmark (#512); this is the one adverse outlier in that run, isolated to a solver defect.

## Reproduction

Fixture: `core/bess/tests/unit/data/historical_2025_01_13_night_low_no_solar.json` (24 periods, no solar, night-low prices, initial SOE 3.0).

Patch `tie_detection.detect_tie_windows` to return `[Window(0, horizon)]` (forces the PWL solver to re-decide every period, terminal SOE pinned to the grid DP's terminal SOE = 3.000), run `optimize_battery_schedule` via `core.bess.tests.helpers._scenario_inputs`, and compare against a run with windows forced empty:

```
grid cost 202.6524   pwl cost 203.2367   delta +0.5843
```

Both plans start at SOE 3.0, end at exactly 3.000, and discharge exactly 12.0 kWh across the morning peak. They differ only in *which* hours (battery_action kWh, buy price SEK/kWh, home load 5.2 kWh/period):

| t | buy | grid action | pwl action |
|---|---|---|---|
| 6 | 1.673 | 0.000 (idle) | −1.920 |
| 7 | 2.136 | −5.160 | −4.920 |
| 8 | 2.605 | −5.220 | −5.160 |
| 9 | 1.948 | −1.620 | 0.000 (idle) |

The PWL plan sells its marginal ~1.9 kWh at the **cheapest** hour (1.673) and idles through the 1.948 hour.

## Evidence this is a solver defect, not accounting or candidate-set

1. **Not reward-vs-flows divergence (#497 class)**: recomputing both trajectories through `_compute_reward` gives total reward = −202.6524 (grid) vs −203.23665 (pwl) — identical to the reported costs to full precision. PWL's plan is genuinely worse under the objective the solver optimizes.
2. **Not a missing candidate**: every grid action is on the 1%-of-max discharge lattice (5.220 = 87% of 6.0 kW, 5.160 = 86%, 1.620 = 27%), so the grid plan is expressible in the PWL action set, from the same start SOE, to the same pinned terminal SOE.
[...]

Issue #487, opened 2026-08-07
consumption statistics feedback loop: inverter self-consumption unmeasured while battery sleeps, night estimate degrades progressively


bess-agent:
## Reported mechanism (unverified)

Surfaced by @ridax67 in #466 ([comment](https://github.com/johanzander/bess-manager/issues/466#issuecomment-5210211953)). Filing separately because it concerns the consumption-forecast *input*, not the optimizer.

When the HA-statistics-based home consumption estimate is used, a plausible self-reinforcing underestimation loop exists for overnight periods:

1. Night consumption is underestimated → the plan under-provisions battery energy for the night.
2. The battery hits the reserve floor early and the inverter goes to sleep.
3. While the inverter sleeps, its own self-consumption (reported ~50% of total night load at low overnight power levels) is no longer measured by the sensors feeding the statistics.
4. The statistics therefore learn an even *lower* night consumption for those hours.
5. Next night the plan under-provisions further → battery drains earlier → more unmeasured hours.

Each iteration worsens the estimate: a systematic downward bias, not random noise.

## Suggested first verification steps

- Confirm how the consumption estimate sources its data and whether inverter self-consumption is inside or outside the measured load during battery-asleep periods.
- Check a real bundle (ridax67's in #466) for whether measured overnight load drops when the inverter sleeps.
- If confirmed: exclude battery-asleep periods from the learning window, or impute self-consumption during them.

## Relations

- #466 — origin thread (IDLE tie-break, separate fix)
- #418 — asymmetric-loss overnight buffer (parked); this bias would compound that problem

bess-product-owner:
Closing as **cannot reproduce**.

**No data supports the reported mechanism, and the one bundle with enough history contradicts it.**

The premise check (above) measured 7 days of overnight consumption from the only bundle carrying sufficient HA Recorder history (`2026-08-04-203125`, serial `rkm0d7n04x`):

| Night | kWh |
|---|---|
| 07-28 | 11.50 |
| 07-29 | 9.00 |
| 07-30 | 11.10 |
| 07-31 | 12.40 |
| 08-01 | 9.60 |
| 08-02 | 10.80 |
| 08-03 | 10.70 |

[...]

Issue #485, opened 2026-08-06
Hysteresis: don't rewrite inverter schedule when the new plan is a near-tie with the applied one


johanzander:
## Problem

The optimizer re-solves every 15 minutes. When two plans are economically near-tied, a small forecast update can flip the chosen intent for future periods back and forth between runs (observed in #466: 19:00 and 22:15 oscillated LOAD_SUPPORT↔IDLE across the morning's runs, ~1.7 öre apart across the whole day).

Consequences by control model:
- **VPP**: mostly cosmetic — confusing schedule churn in the UI, but each re-plan just updates the same remote-control target.
- **TOU**: real hardware concern — every flip rewrites TOU segments to the inverter's flash memory, so near-tie oscillation generates pointless write cycles for zero economic gain. (#320 explored debouncing this from the TOU side.)

## Proposed fix

Hysteresis at apply-time in `battery_system_manager`: before writing a new schedule to hardware, compare the new plan's total value against the currently-applied plan's value under the same (new) forecast. If the improvement is below a small threshold, keep the applied schedule and skip the hardware write.

Notes:
- By definition of a near-tie, skipping the write costs ~nothing economically; the win is stability and flash wear.
- The "how close is close" machinery already exists: the tie-detection work from #461/#467 (`tie_detection.py` on the `design/450-hybrid-dp-pwl-tie-resolution` branch) defines a principled per-period noise threshold (`TIE_NOISE_FACTOR × SOE_STEP_KWH × |dV/dSoE|`) that could be reused or adapted rather than inventing a new magic constant.
- Threshold must be on **plan value**, not on intent labels — two plans can share labels but differ in rates, and vice versa.
- Care needed around genuinely-changed inputs: hysteresis must never hold back a plan that is meaningfully better (e.g. after a price update); it only suppresses churn inside the indifference band.

## Origin

#466 investigation — full numbers in that thread.

Frank-Leysen:
﻿**Cross-link from live Belpex site (20 Aug)** — possible near-tie / horizon-driven plan flip

Same evening Belpex peak (~€0.196) was planned as **BATTERY_EXPORT** at 09:00 (`horizon` 60), then cancelled at 14:45 (`horizon` 133) in favour of D+1 evening export at ~€0.199 (+~0.3 c/kWh).

Full write-up + both debug gists: https://github.com/johanzander/bess-manager/issues/602#issuecomment-5356454017

[...]

Issue #466, opened 2026-08-05
Question: How is IDLE used?


ridax67:
Yesterday evening Bess dumped battery until almost empty and during the night ran in IDLE mode (keep battery soc) until morning when it used what was left for powering during morning.

At least, I think that was the plan and it was a good plan,high price evening, low price night that justified selling first and buying from grid during night.

Unfortunately with Growatt Vpp that wasn't what happened.  It seems IDLE is mapped to load first so the battery was quickly empty.  Isvit correct IDLE should pick a running mode keeping soc level and house power from grid?  I believe the best Vpp setting foe that is Vpp Power +1, remote enabled (Battery first 1).

I will test a little when it is darker.

bess-product-owner:
Closing this one — it started as a question and ended as three shipped fixes, so there's nothing left open on it:

- IDLE was mapped to load-first, which emptied the battery instead of holding it. Fixed.
- Near-ties between IDLE and LOAD_SUPPORT resolved arbitrarily. They now always go to LOAD_SUPPORT; genuine arbitrage holds with a decisive margin are untouched. Shipped in b5.
- The 06:00-06:59 sunrise IDLE was the 100 W discharge step being discarded as too small to be real, plus oversized discharges being filtered out. Fixed.

Your 2026-08-12 follow-up confirmed it: two IDLE periods, both motivated by later pricing. Thanks for the persistence on this one — the near-tie framing in particular became the design principle for the fix.

Reopen if it drifts back.

PR #514, opened 2026-08-09
docs: normative optimizer target architecture + migration plan

johanzander:
## Summary
- Adds `docs/agents/optimizer-architecture.md` — **normative**: every PR touching the optimizer core must uphold principles P1–P7 or amend the document in the same PR. Codifies the 2026-08-08 architecture assessment (ten correction layers, three defect classes), the PWL-everywhere benchmark results (#512, #513), and ridax67's #466 principle ("IDLE must express that the optimizer WANTS to hold, never that it EXPECTS balance") as the intent-semantics rule.
- Adds the phased migration plan: Phase 0 drain WIP (#510, #511, #507, #508, fold-scoping) → Phase 1 single selector (bit-parity, mechanical) → Phase 2 lexicographic tie-policy table (closes the #466 sunrise-crossover residual, subsumes the #466/#510 tie-breaks, unblocks #485) → Phase 3 single flow record (closes #497) → Phase 4 executable-command candidates (closes #320/#352, supersedes #354). Parallel tracks: #487, terminal-value extraction, #513, #512 (sweep-gated).
- Indexes the architecture doc in CLAUDE.md's agent documentation table.

## Notes
- Docs only — no code changes.
- Phases 1–2 are specified to implementation level; Phases 3–4 to acceptance level and start with their own detailed plan/design doc per repo convention.

### DIFF
--- a/docs/agents/optimizer-architecture.md
+++ b/docs/agents/optimizer-architecture.md
@@ -1,0 +1,162 @@
+# Optimizer Target Architecture — Normative
+
+**Status: normative.** Every change to `core/bess/dp_battery_algorithm.py`,
+`core/bess/pwl_window_dp.py`, `core/bess/tie_detection.py`,
+`core/bess/models.py` (flow derivation), `core/bess/strategic_intent.py`,
+or `core/bess/simulation/` must either uphold the principles below or amend
+this document in the same PR, with the reason. "It was the fastest fix for
+the symptom" is not an amendment reason — that is the failure mode this
+document exists to end.
+
+## Why this document exists
+
+Between 2026-06 and 2026-08 the optimizer accumulated **ten** hand-written
+correction layers between the raw Bellman argmax and the emitted action
+(#240, #269, #313, #282, #429, #450, #459, #466, #467, #510). Each was
+individually correct, reviewed, and regression-pinned — and each manufactured
+the conditions for the next. An independent architecture assessment
+(2026-08-08) found the recurring-issue treadmill is structural, driven by
+three defect classes, not by implementation sloppiness. The principles below
+close those classes; the migration order is in
+`docs/superpowers/plans/2026-08-09-optimizer-target-architecture.md`.
+
+The three defect classes, with their issue lineage:
+
+| Class | Mechanism | Issues it produced |
+|---|---|---|
+| **1. Plan/policy mismatch** | DP optimizes point-forecast energy *quantities*; the inverter executes reactive *policies* (load-tracking modes absorb forecast error, IDLE locks it out). The asymmetry is invisible to an expected-value argmax. | #466, #418, #393, #485, part of #352 |
+| **2. Plan/actuation mismatch** | DP plans flows → intent is classified from flows → command derived from intent. Every seam in that chain (percent lattice, grid_first rate semantics, the #350 noise fold running on planned flows) leaks money between plan and execution. | #320, #352, #497/#511, #354, the PLAN_EXECUTION_GAP pins |
+| **3. Solver numerics** | Grid-snapped V + argmax-on-float-noise over a plateau-riddled objective; reward shaping (#240, #269) manufactures new plateaus; each pathological pick ships a bespoke tie-break. | #450, #459, #466, #510, #512, #513 |
+
+The user-facing formulation of Class 1, owed to ridax67 (#466), is the
+design principle for intent semantics:
+
+> **IDLE must express that the optimizer WANTS to hold energy (a decisive
+> margin), never that it merely EXPECTS balance (a coin flip).**
+
+## Principles
+
+### P1. One selector
+
+There is exactly **one** implementation of candidate enumeration +
+evaluation + selection, parameterized by a continuation-value evaluator
+`eval_V(next_soe) -> float`. The grid backward induction, the grid forward
+replay, the PWL backward induction, and the PWL forward replay all call it.
+
+Forbidden: adding a candidate type, guard, cap, or preference to one call
+site and mirroring it by hand into the others. The four ~170-line
+hand-mirrored near-clones (grid replay `_best_action_at_continuous_state`
+vs `_pwl_best_action_at_continuous_state`) are the #236-class /
+`DISCHARGE_LATTICE_PCT_EPS`-class bug factory; the mirror comments
+("branch for branch") are the symptom.
+
+### P2. Ties resolve through one lexicographic preference table
+
+All near-tie resolution (within the single epsilon definition, P5) happens
+in **one ordered preference table**, applied once, inside the P1 selector.
+The baseline order, subsuming the #466 and #510 tie-breaks:
+
+1. Never prefer a candidate that imports more grid energy than the argmax
+   winner.
+2. Never override a decisive winner (margin > epsilon).
+3. Within epsilon: **load-tracking discharge (up to net load) >
+   store-surplus > idle/hold.**
+
+Consequences, by construction:
+
+- IDLE is only ever emitted on a decisive margin — it always means a
+  deliberate hold (arbitrage), never a balance coin flip. This covers the
+  evening near-ties *and* the sunrise/sunset crossover cases from #466.
+- A new tie symptom is fixed by **adding or reordering a table row** — with
+  its rationale and economic bound in the row's docstring — never by adding
+  a new `_prefer_*` function or a new call site.
+- Epsilon-compounding is impossible: the table is applied once against the
+  argmax value, not chained.
+
+Reward shaping that flattens the objective (price floors, export-credit
+zeroing) MUST land together with the table row that resolves the
+indifference region it creates (generalizes the #269/#510 invariant in
+`bess-knowledge.md`).
+
+### P3. Candidates are executable commands
+
+Every candidate the selector considers corresponds to a command the target
+inverter can actually execute — mode (load_first / grid_first /
+battery_first / grid-charge), rate on the hardware's actual lattice, and
+the mode's **reactive semantics** (a load-first discharge cap tracks real
+load; it is not a fixed energy quantity). Candidate value is computed by
+simulating that command's response to the forecast.
+
+Consequences:
+
+- R == P (realized equals planned) becomes a structural property, not a
+  test-suite achievement. The plan cannot propose what the hardware cannot
+  execute (#320, #352, #497/#511, #354).
+- Strategic intent becomes an **input** (the chosen command) rather than a
+  label re-derived downstream from flows. `classify_strategic_intent` on
+  planned flows is a legacy seam to be removed, not extended.
+
+### P4. One flow record per candidate; noise models live at ingestion only
+
+Each candidate's physical flows are computed **once**, and both the reward
+and the reported `EnergyData` derive from that same record. Reward-side
+edits that do not propagate to flows (the #497/#459 divergence class) are
+forbidden.
+
+Sensor-noise heuristics (e.g. the #350 small-export fold) apply **only** in
+the sensor-ingestion path (`sensor_collector.py`). Planned and simulated
+flows are exact by definition and must never pass through a noise model.
+
+### P5. One epsilon, one owner
+
+The tie/noise threshold is defined in exactly one place
+(`tie_detection.epsilon_for_period`) and consumed everywhere ties are
+compared — the preference table, tie-window detection, hysteresis (#485).
+No caller re-derives or hand-picks a SEK margin.
+
+### P6. Exactness claims are certified or bounded
+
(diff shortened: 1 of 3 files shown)


docs/superpowers/plans/2026-08-09-optimizer-target-architecture.md
```diff
@@ -0,0 +1,365 @@
...
+Baseline table (each row = one guard/preference, applied in order, all
+measured against `candidates[argmax_index].value` — never chained):
+
+1. **Guard:** candidates importing more grid than the argmax winner are
+   ineligible (+1e-9 tolerance).
+2. **Guard:** if the argmax winner is a discharge, return it (decisive or
+   not — no preference reorders a discharge winner; #466/#510 ordering
+   made structural).
+3. **Prefer** the largest load-tracking discharge ≤ net load (+half a
+   rate step, capped at `BATTERY_EXPORT_THRESHOLD_KWH/dt`) within epsilon
```
johanzander, line 242: **Should-fix: row 2 silently narrows the #466 guard from "non-idle winner" to "discharge winner".**

Today *both* `_prefer_*` functions bail on any non-idle argmax winner — including a **charge** winner (`_prefer_load_covering_discharge` bails at `dp_battery_algorithm.py:1430` on `abs(best_power) > POWER_TOLERANCE_KW`; `_prefer_curtailed_charge_absorb` at `:1494`). Row 2 as written protects only discharge winners, so under the new table a within-epsilon **charge** winner during positive net load falls through to row 3 and can be flipped to a load-covering discharge — a charge→discharge action flip that passes the row-1 guard (discharge imports less grid) and is not one of the plan's "two deliberate semantic changes".

Either extend row 2 to "non-idle winner" (true parity with today's ordering), or declare this as the third deliberate change with its own acceptance test.

docs/agents/optimizer-architecture.md
```diff
@@ -0,0 +1,162 @@
...
+No caller re-derives or hand-picks a SEK margin.
+
+### P6. Exactness claims are certified or bounded
+
+A solver output spliced over another solver's output (the #450 PWL path)
+must either carry a certification the machinery actually earns, or an
+explicit error bound. #513 (PWL mis-ranks a plan by 0.584 SEK while
+"exact") is the standing counterexample: until fixed, PWL splices are
+treated as *heuristic* improvements, and any new reliance on their
+exactness is forbidden.
```
johanzander: **Should-fix: P6 is aspirational but reads as descriptive.**

`splice_schedule` (`dp_battery_algorithm.py:2290`) splices **unconditionally** — there is no compare-against-grid-segment, accept-only-if-better gate. The only guard is the exactness certification raise (`PWLWindowUnderRefinedError`), and #513 is the proof that certification doesn't hold. So the *existing* reliance on PWL exactness is exactly what P6 forbids, and nothing polices it — "treated as heuristic" is currently a doc claim, not a property of the code.

Cheap fix that makes P6 true: accept a spliced window only if its replayed cost is ≤ the grid segment's (a `_replay_accounting_pass` comparison). Suggest adding that as a P6 compliance item or a small Phase 1/2 rider.

docs/agents/optimizer-architecture.md
```diff
@@ -0,0 +1,162 @@
...
+> margin), never that it merely EXPECTS balance (a coin flip).**
+
+## Principles
+
+### P1. One selector
+
+There is exactly **one** implementation of candidate enumeration +
+evaluation + selection, parameterized by a continuation-value evaluator
+`eval_V(next_soe) -> float`. The grid backward induction, the grid forward
+replay, the PWL backward induction, and the PWL forward replay all call it.
```
johanzander: **Should-fix: P1's "exactly one implementation" collides with Phase 1's own escape hatch.**

The backward induction (`_run_dynamic_programming:949`) is fully numpy-vectorized — on the order of tens of thousands of candidates per period — and the migration plan sensibly allows keeping the vectorized fast path if bit-parity holds. But that outcome (vectorized path + `select_action` sharing enumeration functions, pinned by a parity test) violates P1 *as literally worded*, forcing a doc amendment during Phase 1 itself.

Suggest rewording P1 to what's actually meant: one candidate-**enumeration** + one tie policy, with any vectorized evaluator parity-pinned against the selector.

docs/agents/optimizer-architecture.md
```diff
@@ -0,0 +1,162 @@
...
+
+Forbidden: adding a candidate type, guard, cap, or preference to one call
+site and mirroring it by hand into the others. The four ~170-line
+hand-mirrored near-clones (grid replay `_best_action_at_continuous_state`
+vs `_pwl_best_action_at_continuous_state`) are the #236-class /
+`DISCHARGE_LATTICE_PCT_EPS`-class bug factory; the mirror comments
+("branch for branch") are the symptom.
```
johanzander: **Should-fix: wrong citation.** The "branch for branch" mirror comment is at `dp_battery_algorithm.py:447`, on the `_compute_reward`/`_compute_reward_grid` pair — which this doc's own "What this document does NOT change" section explicitly blesses. The duplication P1 actually targets (`_best_action_at_continuous_state` 213 lines vs the PWL clone ~221 lines) has "mirroring" comments (`pwl_window_dp.py:164,442,458,912`) but not that phrase — a reviewer grepping the quote lands on the protected pair.

Also worth tightening: the `_prefer_*`/`_discharge_candidates`/`_charge_candidate`/`_tie_margin` helpers are **not** hand-cloned in PWL — `pwl_window_dp.py:21-28` imports and calls them directly. Only the two enumerate/select functions are true clones. That makes P1's claim more precise, not weaker.

docs/superpowers/plans/2026-08-09-optimizer-target-architecture.md
```diff
@@ -0,0 +1,365 @@
...
+      existing flow math, not duplicating it); reward derived from that
+      record; `_replay_accounting_pass` consuming the stored records
+      instead of recomputing; sensor-noise heuristics verified to exist
+      only in `sensor_collector.py` (grep every `EnergyData` construction
+      site: `sensor_collector`, `influxdb_helper`, `debug_data_exporter`,
+      `simulation/inverter_simulator`, backend API paths — each gets an
+      explicit exact-vs-ingested decision in the plan).
+- [ ] Acceptance (already written, this is the point): the #497 pin in
+      `test_flow_coherence.py` — 182/1875 incoherent periods — goes to
+      **0**, and the plan-execution gap pins that the fold caused
```
johanzander, line 320: **Should-fix: cited artifact doesn't exist on main.** `test_flow_coherence.py` (and its 182/1875 pin) is not on main — it lives in an unmerged worktree. Since it's Phase 3's acceptance anchor, either add landing it to Phase 0's drain list or mark it "pending WIP" here. Same for `scripts/bench_pwl_everywhere.py` (the #512 gate, referenced at line ~345) — also not on main yet.

docs/superpowers/plans/2026-08-09-optimizer-target-architecture.md
```diff
@@ -0,0 +1,365 @@
...
+    ctx: TieContext,
+) -> int:                          # chosen index
+```
+
+Baseline table (each row = one guard/preference, applied in order, all
+measured against `candidates[argmax_index].value` — never chained):
+
+1. **Guard:** candidates importing more grid than the argmax winner are
+   ineligible (+1e-9 tolerance).
+2. **Guard:** if the argmax winner is a discharge, return it (decisive or
```
johanzander: **Docstring note (no change needed): row 1 is consistent with today** — `_prefer_curtailed_charge_absorb` contains the identical grid-import guard internally (`dp_battery_algorithm.py:1504`), so row 1 subsumes rather than contradicts the #510 rule; the #510 winner holds SOE with ~zero import, and the only candidate blocked is the grid top-up, blocked today too.

The one foreclosure worth recording in row 1's rationale: in **negative buy-price** windows (exactly when curtailment floors bite), a within-epsilon grid top-up is arguably the safer pick against a solar shortfall — and row 1 permanently bans it. Codifies status quo, so fine, but a future amendment will want this context.

docs/superpowers/plans/2026-08-09-optimizer-target-architecture.md
```diff
@@ -0,0 +1,365 @@
...
+
+**The two deliberate semantic changes vs today, and their acceptance:**
+
+- Row 3 drops the `epsilon <= 0.0` early-return (flagged in the #510
+  review: today the tie-break is disabled exactly where the value function
+  is flat, the most degenerate tie). Acceptance: on ridax67's
+  2026-08-07-232503 bundle fixture, the 06:00–06:59 crossover period must
+  resolve to a tracking mode, not IDLE, whenever a lattice-feasible
+  discharge ≤ net load exists; where none exists (86 W load vs 200 W
+  minimum gear) IDLE legitimately stands and the test pins that too.
```
johanzander: **Note: the acceptance hinges on whether the 06:00 crossover is a *bit-exact* tie.** Dropping the `epsilon <= 0` early-return is economically sound — cycle cost is already inside `_compute_reward`, so on an exact tie tracking is weakly dominant under forecast error (IDLE imports the whole consumption miss at buy price; tracking absorbs it at ~breakeven). But at epsilon 0, "within epsilon" catches only bit-exact ties. If the ridax67 crossover period actually has a tiny nonzero gap, the test can only pass by introducing an epsilon *floor* — and a floor **would** trade real model value and needs its own stated bound in the row docstring. Worth checking the bundle before writing the RED test.

docs/superpowers/plans/2026-08-09-optimizer-target-architecture.md
```diff
@@ -0,0 +1,365 @@
...
+
+**Steps:**
+
+- [ ] Write `test_action_selector_parity.py`: for every fixture in
+      `core/bess/tests/unit/data/`, run `optimize_battery_schedule` on
+      current `main` (captured golden outputs: actions + soe_trajectory +
+      battery_solar_cost, stored as a generated JSON under
+      `tests/unit/data/golden/` in the first commit) and assert the
+      refactored path reproduces them bit-identically.
+- [ ] Run it against unrefactored code to prove the golden capture is
```
johanzander, line 168: **Note: state the golden-fixture lifecycle across phases.** Phase 2 deliberately changes behavior, which breaks Phase 1's bit-parity goldens by design. Presumably the answer is "regenerate goldens as part of Phase 2's measured-delta step" — say so explicitly, so nobody "fixes" the parity test by deleting it.

docs/superpowers/plans/2026-08-09-optimizer-target-architecture.md
```diff
@@ -0,0 +1,365 @@
...
+---
+
+## Phase 0: Drain the WIP queue (entry criteria, not new work)
+
+No new architecture work starts while overlapping WIP is open — parallel
+sessions rediscovering each other's findings is how this program got
+confusing.
+
+- [ ] Merge PR #510 (approved 2026-08-08; two optional test nits may ride
+      along or follow up).
```
johanzander: Already done — #510 merged 2026-08-08 (`742d5906`, current main HEAD). Can be checked off.

johanzander (commented):
Thorough review of the proposed architecture (claims fact-checked against main, tie-policy economics stress-tested via bess-analyst).

**Verdict:** the architecture is sound and worth adopting — the three defect classes are real, the issue lineage checks out against the code, and the compliance checklist is enforceable rather than aspirational. But two findings should be fixed in this PR since the document is normative from the moment it lands:

1. **Row 2 silently narrows the #466 guard** from "non-idle winner" to "discharge winner" — an undeclared third semantic change (inline comment).
2. **P6 is aspirational but reads as descriptive** — today's `splice_schedule` relies on PWL exactness unconditionally, which is exactly what P6 forbids, and nothing polices it (inline comment).

Plus three smaller fixes (P1 wording vs the Phase 1 vectorized escape hatch, two cited artifacts that don't exist on main yet, and a wrong "branch for branch" citation) and a few docstring-level notes — all inline.

**What checked out clean:** all named symbols exist where claimed; the four tie-break call sites and the `epsilon <= 0` early-return; the terminal-value method split (`battery_system_manager.py:1920`, #422 day-scoping in the caller at `:2124`); the `PLAN_EXECUTION_GAP_SEK` +0.73/+0.56/+0.45 pins; the #350 fold still in `EnergyData._calculate_detailed_flows` (`models.py:157`); no stray hardcoded epsilons contradicting P5; row 1 subsumes (not contradicts) the #510 tie-break — `_prefer_curtailed_charge_absorb` contains the identical grid-import guard internally; and P7's structural-over-stochastic stance is economically bounded (worst observed forfeiture +0.032 SEK vs the 0.05 SEK/fixture budget).

Issue #352, opened 2026-07-19
bug: low-rate BATTERY_EXPORT writes grid_first at a forced sub-load discharge rate, causing avoidable grid import on any load spike


pookey:
Jeezzzz.... ask a Robot to write you a bug report and you get War and Peace.  Anyway, the issue is that in a recent BESS there seem to have been a change where discharge rates are managed (or managed more aggressively).  The discharge rate is (I think!) proportional to the size the the battery and/or inverter capacity (for me, they are both 5kw).

What this means is that this evening, when I'm typically using not much power it's fine, but I put the oven on to cook dinner, and you get this:

<img width="429" height="414" alt="Image" src="https://github.com/user-attachments/assets/bf5e2afe-b0a0-4e5b-b708-3af8b4216c55" />

It's causing me to draw from the grid for high power demands in the evenings.  I'm almost certain this didn't use to happen.

The below might be nonsense, or might help ;)

--

## What happened

A `BATTERY_EXPORT` period whose planned export is tiny (a few Wh above the classification threshold) is written to the inverter as `grid_first` with a discharge rate sized to the *forecast* battery power — e.g. 16% of a 5 kW inverter = 0.8 kW. `grid_first` does not load-follow: that rate is a hard commitment for the whole 15-minute period.

The moment real house load exceeds that rate, the shortfall is imported at the buy price. On my system this evening (Growatt MIN, 5 kW, UK/Octopus) the export window 20:45–21:59 is planned at rates of 12–19%, i.e. 0.6–0.95 kW, against a buy price of £0.253–0.260 and a sell price of £0.120. Anything ordinary — kettle, oven, immersion — puts the house 2–4 kW over the written rate for several minutes, importing at £0.25 to protect a planned export worth pennies at £0.12.

The upside being defended is genuinely tiny. Period 84 (21:00–21:15) plans a 0.25 kWh draw from storage → 0.2375 kWh AC after discharge efficiency, against 0.22 kWh forecast home consumption. Planned export: **0.0175 kWh** — £0.0021 of revenue. One 15-minute kettle event in that period costs ~0.5 kWh × £0.253 = **£0.13**, roughly 60× the entire upside of the period being in `grid_first` at all.

## Why it happens

Three pieces compose into this:

[...]

PR #516, opened 2026-08-09
feat: halve the DP grid resolution to 0.1 kW / 0.025 kWh (#512)

johanzander:
## Summary
- Halves the DP's state/action grid (`POWER_STEP_KW` 0.2 → 0.1, `SOE_STEP_KWH` 0.05 → 0.025, keeping the reachable-state invariant `SOE_STEP_KWH == POWER_STEP_KW × 0.25h`), recovering **2.43 SEK/day of realized cost across the 35-fixture corpus**.
- Fixes an enumeration-order hole the finer grid exposed in the #466 load-covering tie-break: its guard assumed exact ties always surface with IDLE as the argmax winner, but at the new resolution ties started surfacing at *partial* covers, silently bypassing the swap. The guard now also fires on partial load-covers; charges and genuine exports remain untouched.
- Re-pins 31 fixture `expected_results` from the optimizer, makes grid-coupled unit tests resolution-parametric, retargets the #450 mechanism tests to a fixture that still near-ties (and pins that #450's original reproduction fixture no longer does), and re-measures the tie-coverage harness and #502 gap pin.

## Root cause
From the #512 benchmark: the 0.2 kW / 0.05 kWh grid's value-function discretization left 0.01–0.36 SEK/day unrealized on 19/33 fixtures versus a horizon-spanning exact-PWL bound — error in V itself, which Bellman-optimality checks are structurally blind to (#282 postmortem), mostly in periods the #450 tie detector cannot flag. PWL-everywhere was benchmarked and rejected in the issue (certification failures on 11/33 fixtures, 90–1850× latency).

## Measurements

**Provenance (all figures below, and in `dp_constants.py` and `CHANGELOG.md`, come from this one run):** branch head vs the #511 merge base `33f62129`, realized cost through the inverter simulator (not planned cost), both arms measured in the same worktree with only `dp_constants.py` and `dp_battery_algorithm.py` swapped between them, 35 fixtures. An earlier draft of the `dp_constants.py` docstring quoted 2.08 SEK/day and worst +0.079; those came from a pre-guard-fix sweep on the pre-#511 base and are superseded — the three sites now agree.

| | this PR (2×) | rejected 4× variant |
|---|---|---|
| Realized recovery | **−2.43 SEK/day** | −3.03 SEK/day (25% more) |
| Fixtures better / worse / unchanged | 26 / 8 / 1 | 29 / 6 / 0 |
| Worst per-fixture regression | **+0.0498** | +0.0671, +0.0651 (two over budget) |
| Corpus latency | 15.2 s → **11.2 s** | 26.2 s (2.3×) |
| Worst single-fixture solve | 4.79 s → **1.81 s** | 4.48 s |

**Budget question (raised in review): closed.** Worst per-fixture regression is **+0.0498**, inside the 0.05 SEK/fixture budget in `docs/superpowers/plans/2026-08-09-optimizer-target-architecture.md` — but only just, and `dp_constants.py` now records it that way rather than implying comfortable margin. Grid refinement is not per-fixture monotone, which is exactly why the 4× variant is rejected: it recovers more in aggregate but **breaches** the budget on two fixtures. That is now the recorded rejection rationale (the old "~14% more at 4.4× latency" was a stale pre-#511 figure).

**Latency, per review note:** the aggregate improvement is not hiding a worse tail — the PR's worst single-fixture solve (1.81 s) is lower than main's worst (4.79 s), because fewer near-ties reach the #450 PWL re-solve. ~4× the backward-induction work is more than offset.

**R == P:** exact on every fixture in both arms, except the one known #502 exemption (simulator cannot model PV export-limit curtailment), re-measured +0.0203 → +0.0214 and re-pinned.

## Scope assessment
Local: the diff changes the values of two existing purpose-built constants and widens one existing tie-break's trigger to match its own documented preference ("among eligible candidates the largest coverage wins"). No new parameters, flags, fallbacks, construction sites, or triggers.

Sequenced deliberately after #511 (candidate-set materiality): the finer grid *alone* on pre-#511 code loses most of its planned gain at execution (measured: planned −2.17 but realized only −1.24, with realized regressions up to +0.41/day) because it plans many more small above-load exports that the #350 sensor-noise fold reclassifies to non-exporting commands — the class #511 removed at the root, per optimizer-architecture P3/P4. Refs #352: the small-export `grid_first` exposure class stays out of plans under this grid too.

## Documentation
`docs/SOFTWARE_DESIGN.md` discretization step updated; `dp_constants.py` carries the resolution history, the measurement provenance, and the 4× rejection rationale. `docs/agents/bess-knowledge.md` grep came back clean — its only 0.05-kWh mention is a hand-built test magnitude, not the grid.

**Also fixes a now-false normative claim** (review finding 2): Phase 2 row 2 of `docs/superpowers/plans/2026-08-09-optimizer-target-architecture.md` stated that both `_prefer_*` functions bail on any non-idle winner. That is false in both directions after this PR — `_prefer_load_covering_discharge` deliberately fires on partial-cover winners (a literal implementation of the old row 2 would re-open the hole this PR closes), and `_prefer_curtailed_charge_absorb` bails on discharge winners only, so it can already improve a charge winner today. Row 2 is rewritten as a per-preference guard citing #512, since a single shared non-idle guard cannot express the deliberate asymmetry.

## Test plan
- [x] `./scripts/quality-check.sh` passes (0 errors, 0 warnings; frontend 124/124)
- [x] `pytest -m "not slow"`: 1648 passed — `pytest -m slow`: 417 passed
- [x] Fixture pins verified discriminating: running the re-pinned suite against the coarse grid fails 24 scenario tests
- [x] `/code-review` run; single finding (stale docstring) fixed
- [x] Observed end-to-end via mock-HA E2E (`docker-compose.ci.yml`, `ci-normal-day`, libfaketime at the scenario date): real backend built the schedule (savings 86.14 SEK), all 24 nonzero planned actions on the executable lattice with load-exact evening discharges (−0.650/−1.150/−1.500 kW — inexpressible on the old 0.2 kW action grid), intent→TOU conversion ran through the Growatt controller
- [ ] Beta-first: real-hardware validation via the beta channel before stable release

Closes #512


johanzander (commented):
Reviewed the diff, ran the fast suite on the branch (1648 passed, 14 skipped) and the changed-logic tests specifically (tie-break + #450 + plan-faithfulness: 27 passed). The measurement discipline here is strong — R == P on every fixture, discriminating-pin verification, the rejected 0.05 kW halving recorded with its cost, and the deliberate sequencing after #511 with the pre-#511 counterfactual measured rather than asserted. Two findings, one of which is a blocker for a *different* file than this diff touches.

**1. Blocker (cheap): the PR contradicts itself on both headline numbers.**

`dp_constants.py` says the change "recovers **2.08** SEK/day across the corpus ... net of small non-monotone per-fixture regressions (worst **+0.079**)". The PR body and CHANGELOG say **2.43** SEK/day (~2.4) and worst realized regression **+0.0498**.

These can't both be right, and the difference is material: **+0.079 breaches the 0.05 SEK/fixture regression budget** that `docs/superpowers/plans/2026-08-09-optimizer-target-architecture.md` requires every phase to stay inside, while +0.0498 sits just under it. The code comment is the durable artifact — whichever measurement is current, please reconcile the three sites (constant docstring, PR body, CHANGELOG) and state which run they came from. If +0.079 is the live number, the budget question needs an explicit answer before merge, not a silent pass.

**2. Should-fix, in `docs/superpowers/plans/...` not in this diff: this PR invalidates a normative claim, and Phase 2 would regress it.**

The #466 guard change is correct and well-argued — which candidate an argmax returns among exact ties *is* an enumeration-order accident, and the fix is faithful to the tie-break's own documented preference. I verified the logic: charges bail (`-best_power > TOL` false), genuine exports bail (`-best_power > max_cover_p`), full covers pass but no-op, partial covers swap up. Correct as intended.

But the merged plan doc's Phase 2 row 2 (line 235) now states something false:

> **Guard:** if the argmax winner is any non-idle action (`abs(power) > POWER_TOLERANCE_KW` — charge OR discharge), return it. This is exact parity with today: both `_prefer_*` functions bail on any non-idle winner (dp:1430, dp:1494).

After this PR, `_prefer_load_covering_discharge` deliberately does *not* bail on a partial-cover discharge winner. Implementing row 2 as written in Phase 2 would silently re-introduce exactly the hole this PR closes. Please update row 2 in the same PR to: bail on charge winners and on discharges already beyond the load cover; partial covers remain eligible for improvement. (Cite #512 so the Phase 2 implementer sees why.)

**Notes, no action needed:**
- Retargeting the #450 mechanism tests is handled well — pinning that the original reproduction fixture *stops* tying is better than silently losing the assertion. Worth knowing that #450's original user-reported regression is now protected indirectly.
- The latency claim (17.4 s → 13.3 s) is a corpus aggregate that depends on fewer PWL windows firing. Since ~4x the backward-induction work is now offset by fewer re-solves, a day that still ties heavily could be slower than before. Not a merge concern on an hourly scheduler, but if a worst-case *per-fixture* latency was measured, it's worth putting next to the aggregate.

Issue #393, opened 2026-07-25
LOAD_SUPPORT shadow-price gate: near-tie residual leaves some overnight grid imports uncovered


johanzander:
## Context

Split out from PR #385 (LOAD_SUPPORT shadow-price discharge gate, closing #384) during code review. The CHANGELOG and `docs/agents/bess-knowledge.md` entries for #385 originally cited #276 as the tracked follow-up for a documented residual limitation, but #276 is closed and is about an unrelated topic (DP grid-discretization/interpolation error from #275, concluded "financially optimal" per the #275/#276/#285 investigation — see PR #286). This issue is the correct tracking home for the #385-specific residual instead.

## The residual limitation

`docs/agents/bess-knowledge.md`'s "Known limitation, not yet fixed" note (added by #385): during a sustained overnight `LOAD_SUPPORT` drawdown, `shadow_price` tends to sit within a cent or two of `buy_price` for the entire stretch — close to the defining condition for choosing `LOAD_SUPPORT` at all — so `intra_period_discharge_gate`'s strict inequality (`buy_price × efficiency_discharge >= shadow_price`) rarely opens in that regime. Small, real overnight grid imports (observed: 0.1-0.2 kWh/night, a few tens of cents) can still occur even with ample SOE headroom above the floor.

#385 explicitly rejected widening the gate's tolerance to paper over this — the gap between "genuinely arbitrary near-tie" (safe to open) and "real reserve correctly protected for a meaningfully pricier future peak" (must stay closed) is not currently distinguishable from the gate alone.

## What this issue proposes investigating

Whether the near-tie is itself a symptom of DP value-function granularity/precision near the LOAD_SUPPORT decision boundary, or a fundamentally separate question requiring a different signal than shadow price alone to disambiguate "arbitrary tie" from "real future value." No prototype yet — filed so the residual isn't lost, not because a specific fix direction is proposed.

## Note on the originating worked example

[...]

PR #525, opened 2026-08-09
refactor: one ordered preference table for near-tie resolution (Phase 2, P2)

johanzander:
**This is Phase 2 of the optimizer migration, implementing P2** ("Ties
resolve through one lexicographic preference table") from
`docs/agents/optimizer-architecture.md`. Plan:
`docs/superpowers/plans/2026-08-09-optimizer-target-architecture.md`,
Phase 2. Follows Phase 1 (#521, P1).

## What changed

`core/bess/tie_policy.py` (new) holds one ordered preference table,
applied once from `action_selector.select_action`. The two bespoke
tie-break helpers are deleted; their docstring rationale moved into the
rows. Repo-wide `grep -rn "_prefer_"` returns nothing — P2's exit gate.

### The table as implemented

| # | Kind | Rule | Lineage |
|---|---|---|---|
| 1 | Eligibility | A candidate importing more grid energy than the argmax winner (+1e-9) is never preferred. | lifted out of the curtailment helper; now table-wide |
| 2 | Eligibility | Within `epsilon` of the argmax winner's value, **inclusive**, with `epsilon` from `tie_detection.epsilon_for_period` and nowhere else (P5). Measured against the argmax for every row — never chained. | #450 / P5 |
| 3 | Preference | Prefer the largest load-tracking discharge ≤ this period's net load. Stands down on a **charge** winner and on a discharge **already beyond the cover**; an IDLE **or partial-cover** winner stays eligible. | #466, widened by #512 |
| 4 | Preference | When `sell_price_floored`, prefer the highest-`next_soe` eligible candidate. Stands down on a **discharge** winner (including a row 3 swap) but may still improve a **charge** winner. | #510 / #269 |
| 5 | — | Otherwise the argmax winner stands. | |

Rows 3 and 4 keep their **own** guards, not a shared one. The asymmetry is
deliberate in both directions: a shared "bail on any non-idle winner"
guard would re-open the partial-cover hole #512 closed, and would also
remove row 4's existing ability to improve a charge winner.
`test_row_4_may_still_improve_a_charge_winner` and
`test_partial_cover_winner_is_improved_to_a_fuller_cover` pin the two
sides.

### The two deliberate semantic changes

1. **Epsilon band anchored at the argmax, not chained.** Previously row 4
   received row 3's output and measured its band against it, so the two
   could spend up to `2 * epsilon` between them. Pinned by
   `test_epsilon_band_is_measured_against_the_argmax_not_the_previous_row`.
2. **`epsilon <= 0` early return dropped.** Both helpers disabled
   themselves exactly where the value function is flattest. The band is
   now inclusive, so at `epsilon == 0` it admits only bit-exact ties.

## Epsilon-floor decision: **no floor**, measured

The plan required measuring a real period's margin before writing the RED
test for the `epsilon <= 0` change, and forbade silently introducing a
floor to make a test pass.

Measured across all 36 fixtures (2194 selector calls):

- 234 periods have `epsilon == 0`; 114 of those carry a bit-exact tie at
  the argmax value.
- The table fires on **0** of them. Those exact ties are the known
  IDLE / SOLAR_EXPORT-bypass duplicate pair (identical power, identical
  `next_soe`), never an IDLE-vs-load-cover pair — with a flat V,
  discharging is strictly better or strictly worse, so it does not tie.
- Of the 32 `epsilon == 0` periods where a strictly larger load cover was
  actually available, **the smallest gap to it is 0.0080 SEK**
  (next: 0.0124, 0.0139, 0.0209, 0.0225 …).

0.0080 SEK is a real ranking, not grid-snap noise: a floor large enough to
cross it would trade genuine model value on 32 periods, and per-period
forfeitures of that size compound past the 0.05 SEK/fixture budget. So no
floor is introduced, and the epsilon-0 semantics are pinned at unit level
(`test_zero_epsilon_still_ranks_a_nonzero_gap`,
`test_zero_epsilon_resolves_a_bit_exact_tie_to_the_load_cover`) rather
than against a fixture that cannot exhibit the case.

## Measured deltas: zero

Per-fixture planned cost (`economic_summary.battery_solar_cost`) and
emitted actions were compared against `origin/main` via the Phase 1
goldens:

**0 of 36 fixtures changed. Worst |Δ planned cost| = 0.0000 SEK. 0 actions
differ on any fixture.** Well inside the 0.05 SEK/fixture budget.

Consequently **no `PLAN_EXECUTION_GAP_SEK` pin moves** — the sole pin
(`regression_2026_08_08_143843`: +0.0214) is unchanged and re-verified by
the slow suite.

### Golden regeneration

`scripts/capture_selector_goldens.py` was re-run **from this branch** (as
required for a behavior-changing phase) and all 36 goldens regenerate
**byte-identically** — `git status` on `core/bess/tests/unit/data/golden/`
is empty. `test_action_selector_parity.py` is neither deleted nor skipped
and passes as-is.

This is the plan's first surprise (see below): Phase 2 was expected to
break the goldens by design. It does not.

### #466 evening near-ties (acceptance)

Replaying `regression_2026_08_06_466` (bundle 2026-08-06-110152,
dt=0.25 h), the evening periods resolve to load tracking exactly as
before: periods 32–45 (19:00–22:15) emit `LOAD_SUPPORT` with the winner
already at the exact cover, and periods 46–50 swap a partial cover up to
the fuller one (row 3 firing at epsilon 0.001564). Period 51 sits at
`epsilon == 0` with no larger cover candidate available at all, so nothing
to resolve. #393's overnight-residual regression
(`test_load_support_gate_regression_393.py`) passes unchanged.

## Tests

- New: `core/bess/tests/unit/test_tie_policy.py` — 22 cases. Every case the
  two retired helpers were pinned against, ported with the same candidates
  and expected indices, plus the row 3/row 4 asymmetry, the
  no-epsilon-compounding property, and the epsilon-0 semantics.
- `test_idle_tie_break.py` / `test_curtailment_charge_early_tiebreak.py`:
  candidate-level cases moved to `test_tie_policy.py` with the rule they
  test; the end-to-end replay and fixture tests stay and are unmodified.

```
.venv/bin/pytest -m "not slow"  ->  1677 passed, 14 skipped
.venv/bin/pytest -m slow        ->   457 passed,  3 skipped
.venv/bin/black --check .       ->  200 files unchanged
.venv/bin/ruff check .          ->  All checks passed
```

(`./scripts/quality-check.sh` reports "not installed" warnings because it
looks outside `.venv`; the tools were run directly.)

## Doc amendment

`docs/agents/optimizer-architecture.md`: `tie_policy.py` added to the
scope gate, and P2 marked satisfied with the three things the baseline
order stated loosely (per-preference row 2, inclusive band / no floor with
the 0.0080 SEK measurement, and that the crossover is Phase 4's).

No CHANGELOG entry: measured user-visible effect is exactly zero (repo
policy on pure refactors).

## Explicitly NOT in this PR

- The **P6 splice cost-gate rider** listed as a Phase 2 step. It is a
  separate mechanism in `splice_schedule`, not tie policy, and is not
  needed for P2's exit gate. Needs its own PR.
- The mock-HA crossover replay step — #517 already covers the crossover
  via the candidate space, and the plan itself retracted it as this
  phase's acceptance.
- Answering/closing #466 or #393. Beta graduation first (repo policy); no
  reporter issue is closed here.

### DIFF
--- a/core/bess/tests/unit/test_tie_policy.py
+++ b/core/bess/tests/unit/test_tie_policy.py
@@ -10,7 +10,7 @@ used to refuse.
 import pytest
 
 from core.bess.action_selector import Candidate
-from core.bess.tie_policy import TieContext, apply_tie_policy
+from core.bess.tie_policy import TieContext, _eligible_indices, apply_tie_policy
 
 
 def _candidate(value, power, next_soe, new_cost_basis, reward, grid_imported):
@@ -190,15 +190,40 @@ def test_row_3_swap_survives_row_4():
 
 
 def test_epsilon_band_is_measured_against_the_argmax_not_the_previous_row():
-    # Epsilon-compounding: a candidate 1.5x epsilon behind the argmax is
-    # ineligible, even though it sits within epsilon of the candidate an
-    # earlier row would have chosen. Chaining the bands -- what the two
-    # retired helpers did -- would let the table spend 2x epsilon.
+    """The band is anchored at the argmax: a candidate 1.5x epsilon behind
+    it is ineligible even though it sits within epsilon of `near`.
+
+    HONEST SCOPE (review finding): this pins the anchor, but it does NOT
+    discriminate against the retired chained helpers -- they return the
+    same index here, because with home == solar == 0 row 3 stands down and
+    the chained anchor *is* the argmax.
+
+    No end-to-end case can discriminate, and that is a property of the
+    table rather than a gap in the fixture: row 3 returns either the argmax
+    or a discharge, and row 4 returns immediately on a discharge winner, so
+    the two anchors provably coincide whenever row 4 actually runs. (The
+    Phase 2 measurement confirmed it empirically: 0 divergences across 2194
+    selector calls.) Anchoring is therefore a structural guarantee that
+    survives a future row being inserted, widened, or reordered -- none of
+    which the chained arrangement would have survived -- not a behavior
+    difference observable today.
+
+    What IS discriminating is the direct assertion below: eligibility is
+    computed from `argmax_index`, so re-pointing it at any other index
+    fails here.
+    """
     epsilon = 0.01
     winner = _candidate(10.0, 0.0, 12.0, 0.035, 0.0, 0.0)
     near = _candidate(10.0 - 0.9 * epsilon, 1.0, 12.4, 0.035, 0.0, 0.0)
     far = _candidate(10.0 - 1.5 * epsilon, 2.0, 12.9, 0.035, 0.0, 0.0)
     assert _pick_floored([winner, near, far], 0, epsilon=epsilon) == 1
+
+    # Anchor pinned directly: measured from candidate 0 (the argmax), only
+    # `winner` and `near` are within the band. Measured from `near` instead
+    # -- the chained anchor -- `far` would also qualify.
+    candidates = [winner, near, far]
+    ctx = _ctx(epsilon=epsilon, home=0.0, solar=0.0, floored=True)
+    assert _eligible_indices(candidates, 0, ctx) == [0, 1]
 
 
 # ------------------------------------------------------- epsilon == 0 (P2)
--- a/core/bess/tests/unit/test_tie_policy.py
+++ b/core/bess/tests/unit/test_tie_policy.py
@@ -1,0 +1,240 @@
+"""The one ordered preference table (P2, `core/bess/tie_policy.py`).
+
+Every case the two retired tie-break helpers were pinned against, ported
+onto `apply_tie_policy` with the same candidates and the same expected
+indices, plus the cases that only exist once the rows share one table:
+the row 3 / row 4 asymmetry, and the epsilon == 0 semantics both helpers
+used to refuse.
+"""
+
+import pytest
+
+from core.bess.action_selector import Candidate
+from core.bess.tie_policy import TieContext, apply_tie_policy
+
+
+def _candidate(value, power, next_soe, new_cost_basis, reward, grid_imported):
+    """Build a `Candidate` from the positional layout these fixtures were
+    written against, so the cases below stay literally what they were when
+    the tie-breaks lived on 6-tuples."""
+    return Candidate(
+        power=power,
+        next_soe=next_soe,
+        reward=reward,
+        new_cost_basis=new_cost_basis,
+        grid_imported=grid_imported,
+        value=value,
+    )
+
+
+def _ctx(epsilon=0.01, home=0.25, solar=0.0, dt=0.25, floored=False):
+    return TieContext(
+        epsilon=epsilon,
+        home_consumption=home,
+        solar_production=solar,
+        dt=dt,
+        sell_price_floored=floored,
+    )
+
+
+# ---------------------------------------------------------------- row 3 (#466)
+
+# Candidate fields: (value, power, next_soe, new_cost_basis, reward, grid_imported)
+IDLE = _candidate(10.00, 0.0, 6.0, 0.0, 0.0, 0.25)
+COVER = _candidate(9.995, -1.0, 5.74, 0.0, 0.25, 0.0)  # covers 1 kW net load
+OVER = _candidate(9.999, -3.0, 5.21, 0.0, 0.30, 0.0)  # past load -> would export
+CHARGE = _candidate(9.99, 2.0, 6.5, 0.0, -0.5, 0.75)
+
+
+def _pick(candidates, argmax_index, **ctx_kwargs):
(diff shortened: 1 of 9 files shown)


core/bess/tests/unit/test_tie_policy.py
```diff
@@ -0,0 +1,240 @@
...
+    candidates = [IDLE, COVER, ABSORB]
+    chosen = apply_tie_policy(
+        candidates,
+        0,
+        _ctx(epsilon=0.01, home=0.25, solar=0.0, floored=True),
+    )
+    assert chosen == 1
+
+
+def test_epsilon_band_is_measured_against_the_argmax_not_the_previous_row():
```
johanzander, line 192: **This test does not discriminate old vs. new behavior.** Running its exact candidates through `origin/main`'s chained `_prefer_load_covering_discharge` → `_prefer_curtailed_charge_absorb` returns index `1` — the same as `apply_tie_policy`. The case uses `_pick_floored` (home == solar == 0), so row 3 stands down and the chained anchor *is* the argmax.

So deliberate semantic change #1 (epsilon band anchored at the argmax rather than chained) currently has no test that would fail if it were reverted. Either build a case where row 3 actually moves the index and row 4 still runs, or state in the test that the property is structurally guaranteed rather than empirically pinned.

core/bess/tie_policy.py
```diff
@@ -0,0 +1,255 @@
...
+    as inert for a future row: it is a live bound.
+
+    Recorded foreclosure: in a negative-buy-price window a within-epsilon
+    grid top-up is arguably the safer pick against a solar shortfall, and
+    this row permanently bans it. That codifies today's behavior; an
+    amendment relaxing it must cite this note and bring field evidence.
+
+    **Row 2 -- within epsilon of the argmax value.** Measured once, against
+    `candidates[argmax_index].value`, never against a previous row's output:
+    chaining let two rows spend up to `2 * epsilon` between them, which is
```
johanzander: The stated motivation isn't true of the code being replaced: chaining could never actually spend `2 * epsilon`. `_prefer_load_covering_discharge` returns either `argmax_index` or an index whose `power < -POWER_TOLERANCE_KW`, and `_prefer_curtailed_charge_absorb` returned immediately on a discharge `best` — the two bands could not compound in any reachable state.

Suggest softening to "made unrepresentable rather than merely unreachable", so a future reader doesn't treat this as evidence of a past miscomputation.

core/bess/tie_policy.py
```diff
@@ -0,0 +1,255 @@
...
+    bit-exact-tie case this row's own reproduction fixture
+    (`regression_2026_08_08_143843`) exhibits -- the tie there is exact by
+    construction, since a floored sell price makes both options earn
+    literally the same reward.
+    """
+    incoming = candidates[incoming_index]
+    if incoming.power < -POWER_TOLERANCE_KW:
+        return incoming_index
+
+    chosen_index = incoming_index
```
johanzander: `_row_stored_energy` seeds both its guard and `chosen_next_soe` from `incoming_index` (row 3's output) rather than `argmax_index` — the one surviving piece of row-to-row chaining in a module whose premise is that no row reads another's output.

It's inert today only because row 3 can return nothing but the argmax or a discharge (and a discharge trips the early return). Insert a row between 3 and 4, or widen row 3 to return a non-discharge index, and row 4's baseline SOE silently becomes that row's output instead of the argmax — reintroducing exactly the coupling row 2's docstring calls unrepresentable, with no test to catch it.

Consider seeding `chosen_next_soe` from `candidates[argmax_index]` and keeping only the guard on `incoming_index`: behavior-identical today, and the invariant then holds by construction.

johanzander (commented):
Code review: **no correctness regressions found in the production path.**

Verification run: ruff + black clean on the changed files; `test_tie_policy.py` 22/22 pass; and `origin/main`'s two retired helpers were executed side-by-side against every candidate scenario in the new test file — only the two `epsilon == 0` cases differ (`eps0_bitexact_cover`, `eps0_bitexact_curtail`), which independently corroborates the "0 of 36 fixtures changed" claim.

Checked and cleared: the row-3 winner guard rewrite (`winner_power > tol` / `-winner_power > max_cover_p`) is equivalent to the old `is_idle or is_partial_cover`; row 1's promotion to a table-wide guard cannot downgrade a partial cover, since the winner is always in `eligible` and row 3 takes the max discharge over that set; `tie_policy` → `dp_battery_algorithm` is not a runtime import cycle; `epsilon` from `epsilon_for_period` is non-negative by construction; both forward replays pass `sell_price_floored` and route through the single selector.

Three non-blocking findings inline.

PR #534, opened 2026-08-10
Phase 3: one flow record per candidate (P4)

bess-agent:
Executes Phase 3 of the optimizer target-architecture migration
(`docs/superpowers/plans/2026-08-10-phase3-one-flow-record.md`).

A consolidation refactor plus **one live economic fix**. Acceptance is golden
bit-parity, and it held on all 36 fixtures at every step.

## Task 0 — mandatory re-measurement

The plan's survey ran on a stale checkout, so every figure was re-derived on
`a81ac613` (36 fixtures / 2168 periods) before anything was relied on.

| Measurement | Stale claim | Re-measured | Verdict |
|---|---|---|---|
| Fixtures / planned periods | 32 fixtures | **36 / 2168** | as expected |
| `_compute_reward` vs `_build_period_data`, `grid_imported` | 0.0 | **0.0** (0 of 2461 rows) | confirmed |
| Same, wear cost | ≤6.7e-16 | **≤6.661e-16** (42 of 2461 rows) | confirmed |
| Knife-edge selections | 46 / 2194 | **56 / 2194** | higher — see below |
| Fixtures firing the idle guardrail | 0 of 32 | **0 of 36** | confirmed |
| `EnergyData` heuristic firings on planned data | 0 | **0** | confirmed |

No re-measurement contradicted the plan materially, so the scope stood. Two notes:

- **`grid_imported` agrees exactly**, so Tasks 1–3 are hygiene, not a bugfix — stated
  plainly as Task 0 requires. The only deviation is the wear term's two
  algebraically-identical formulations, at float-noise magnitude.
- **Knife-edge count is higher, not lower.** The stale definition was not reproducible
  (Phase 2 replaced tie resolution), so this was measured against the question that
  matters for bit-parity: how often the strict-`>` argmax is decided by a gap under
  1e-12 against a candidate the goldens would record *differently*. 56 argues *harder*
  for the plan's "do not re-associate the reward arithmetic" instruction. Ties by the
  codebase's own ambiguity criterion (`TIE_DEDUP_SOE_KWH`) number only 4.

## What changed

**Task 1** — `PeriodFlows` carries a candidate's complete flows, derived in one place
(`_period_flows`). `_compute_reward` prices it and returns it, `select_action` stores it
on `Candidate`, `_build_period_data` consumes it. `flows` is required with no recomputing
default, and `import_cap_kwh` is gone from `_build_period_data` — a reporting-side copy of
that throttle could only ever drift from the priced one.

**Task 2** — pricing is split out as `_price_flows(flows, ...)`, so the objective is
literally a function of the record. The record travels with its action through
`resolve_pwl_window` → `splice_schedule` → `_replay_accounting_pass`. This closes a real
divergence path: the replay took its own `import_cap_kwh`, and a mismatch with the solve
would have silently priced a period the plan never contained.

**Task 3** — `_create_idle_schedule` loses its fourth copy of the IDLE physics. That copy
had drifted, and collapsing it adopts the correct basis: it charged the full
`battery_charged * sell_price` as solar opportunity cost where `_price_flows` discounts
the AC-clipped share. Measured: 4 of 36 fixtures have an AC cap, the basis moves on
exactly one (`synthetic_clear_sky_ac_clipping`) by **0.049977 SEK/kWh**. No *comparison*
moves — the guardrail reads `battery_solar_cost`, which `cost_basis` does not feed, and it
fires on 0 of 36 fixtures. But when it does fire this schedule is returned, so a reported
field (`decision.cost_basis`) shifts on AC-capped systems.

**Task 4 — the live defect.** `_calculate_initial_cost_basis` derived
`min(battery_charged, solar_production)`, ignoring the house load, so solar the home had
already consumed was counted again as charging the battery and the grid energy that really
did it was booked as free. Both this and the energy-balance log table (labelled "Estimate
solar to battery (simplified)") now read the split `EnergyData` already derived.

> Measured on **real recorded data** (30 debug bundles): **118 charging periods across 15
> bundles, 48.9 kWh of grid charging booked as free solar**, and on the pinned day the cost
> basis is understated by **66%** (0.5936 vs 0.9844 SEK/kWh). Two known limitations are
> bounded with measurements at the site: the split is inverted during deliberate
> `GRID_CHARGING` (26 periods / +3.85 SEK, against +55.15 SEK correctly attributed in
> load_first), and the unattributed-remainder estimator is biased upward only.

**Task 5 — attempted, then backed out. The P4 violation it targeted still stands.**
Its stated precondition was that it be a *zero-behaviour-change* compliance move, and
review showed it is not one, twice over:

- Two of the three "heuristics" are **physical bounds**, not noise models — a flow cannot
  exceed its source. Gating them let an exact record attribute more export to the battery
  than the battery discharged, and the withdrawn revision's own test asserted that
  fabrication as expected. Both are unconditional again, with the invariant now pinned in
  the model.
- Gating the #350 fold **is** a behaviour change, and it reaches hardware. My claim that
  `_discharge_is_unexecutable` (#497) keeps the optimizer out of the band was wrong: that
  predicate computes its deficit as `home - solar`, from raw solar, blind to AC clipping.
  Whenever `ac_cap < home < solar` it stands down while the AC stage still leaves the
  battery overshooting by <`GRID_FLOW_RESOLUTION_KWH`. Constructed and confirmed: intent
  flips LOAD_SUPPORT → BATTERY_EXPORT, which maps to `grid_first` — the export feedback
  loop `battery_system_manager` documents.

Task 0's "0 firings on 2168 planned periods" was weak cover: only 4 of 36 fixtures have an
AC cap, and a 2170-run sweep I ran to defend the change found 0 hits **but never entered
the `ac_cap < home` regime**, so it could not have found one.
`test_planned_flows_still_pass_through_the_ingest_fold` now pins the current
non-compliant behaviour with the reproduction, and says it should flip when the gate lands
rather than be deleted. Closing it needs an AC-aware candidate filter — Phase 4.

**Task 6** — corrects three unmeasured claims in the parent plan (details below).

## On bit-parity and Task 4

The phase brief expected Task 4 to move corpus numbers and break parity. **It did not**,
and the reason is structural rather than reassuring: corpus fixtures pass
`initial_cost_basis` in as an input, so the goldens never exercise
`_calculate_initial_cost_basis`. The corpus cannot measure this fix in either direction —
the 36.16 kWh figure is the evidence, not the gate. The regression test was verified to
FAIL against the pre-fix split rather than pass vacuously.

## Code review

Two passes. **The maintainer's review found the two most serious defects** — both in
Task 5, which is why that task is now withdrawn (see above). It also asked for an
allocation number for `PeriodFlows`: measured, no regression. Median
`optimize_battery_schedule` over 7 runs, main vs branch — 57.4/54.8 ms, 172.9/181.4 ms,
53.4/54.7 ms, within run-to-run noise.

`/code-review high` (22 agents) returned 10 findings. Two were real regressions this
branch introduced, both fixed in `18f5d58e`:

- **Unattributed charge booked at zero cost.** `EnergyData.grid_to_battery` is capped by
  the grid counter, so on measured data the attributed flows can sum to less than
  `battery_charged` — while the full amount still entered `running_energy`, diluting the
  basis downward. The comment called the remainder "sub-resolution"; nothing guaranteed
  that. Now priced at cycle cost, as this method already prices energy of unknown
  provenance.
- **Stale flow record at the spliced window boundary.** A window owns `[start, end)` but
  writes the SOE at `end`, so that period reported a post-splice `battery_soe_start`
  alongside pre-splice charge figures. Re-derived per window now.

Also fixed: a stale `window_resolutions` annotation, and a `_period_flows` docstring that
overclaimed being "the only place the charge split is computed" while `_state_transition`
and the numpy mirrors still carry copies.

Deliberately not fixed (recorded with reasons): the simulator's missing `import_cap_kwh`
(pre-existing, and its `_state_transition` call omits it too, so a flows-only fix would be
half a fix); old prediction snapshots deserializing `measured=True` (nil impact — 0
firings on planned data, intra-day transient).

## Judgement calls worth review

- **`_create_idle_schedule` gained a `currency` argument** only because
  `_build_period_data` declares one it never reads. Removing the dead parameter would pull
  `inverter_simulator.simulate`'s public signature into this diff. Left alone and flagged
  — a clean follow-up.

## Verification

- Golden bit-parity: **36/36 byte-identical** on `actions`, `soe_trajectory`,
  `battery_solar_cost`, at every commit
- `assert_flow_coherence` green throughout (floor, not goal)
- 1688 fast + 457 slow passed; `./scripts/quality-check.sh` 0 errors

Task 6's plan corrections: "six flow derivations" was never measured (a census found ten
sites, five removed here); the claim that noise heuristics live only in
`sensor_collector.py` was false (three are in `models.py`, plus `energy_flow_calculator`'s
fallbacks); and the `EnergyData` construction-site list named two modules that construct
none.


core/bess/battery_system_manager.py
```diff
@@ -2980,21 +2980,52 @@ def _calculate_initial_cost_basis(self, current_period: int) -> float:
...
+                # exactly as the pre-existing charge at `initial_soe` above.
+                unattributed = max(
+                    0.0,
+                    event.energy.battery_charged - solar_to_battery - grid_to_battery,
+                )
+                unattributed_cost = (
+                    unattributed * self.battery_settings.cycle_cost_per_kwh
+                )
+
+                new_energy_cost = solar_cost + grid_cost + unattributed_cost
```
bess-agent, line 3067: **medium — the `unattributed` bucket prices leftover charge *cheaper* than the retired formula, in exactly the case this PR set out to fix.**

When `solar_production > 0` but `grid_imported` under-reads, the old split assigned the remainder to grid at `buy_price + cycle_cost`; this assigns it to `cycle_cost` only.

Not hypothetical — running the split over the bundle this PR commits as its own evidence, `historical_2026_07_18_charge_attribution.json` period 39 (`charged=0.5, solar=0.4, home=0.0, grid_imported=0.0`) has `resid=0.1 kWh` that the retired formula booked at `buy+cycle` and this books at `cycle`.

Degenerate case: a period with grid charging and a zero/reset grid counter now books the *entire* charge as effectively free, understating the basis and making stored energy look cheaper to discharge — the precise failure mode Task 4 targets. `test_real_day_has_charge_neither_source_explains` only asserts `actual > uncosted` (vs. costing it at zero), so it cannot catch this direction.

Suggest pricing the remainder conservatively at `buy_price + cycle_cost` rather than at the cheapest available basis.

core/bess/dp_battery_algorithm.py
```diff
@@ -568,47 +711,85 @@ def _compute_reward(
...
+    the selection loop used would otherwise silently price a period the plan
+    never contained.
+
+    Takes no `import_cap_kwh`: the cap is a constraint on *flows*, already
+    applied inside `_period_flows`. Re-applying it here could only
+    disagree.
     """
     current_buy_price = buy_price[period]
     current_sell_price = sell_price[period]
     ac_cap_kwh = _effective_ac_cap_kwh(battery_settings, dt)
```
bess-agent, line 803: **low — `_price_flows` re-derives `ac_cap_kwh` instead of taking it from the record's producer.**

`_period_flows` accepts `ac_cap_kwh` as an explicit parameter; this recomputes it. The two can silently disagree — the exact reward-vs-flows divergence class P4 exists to remove, just relocated.

Concretely: a caller that derives flows with `ac_cap_kwh=None` (unclipped) and then prices them gets `solar_opportunity_cost` computed from a clipped `export_without_storing`, mispricing the period. Latent today only because every production caller happens to pass `_effective_ac_cap_kwh(...)`.

Either carry the cap on `PeriodFlows`, or drop the parameter from `_period_flows` so there is one source.

core/bess/dp_battery_algorithm.py
```diff
@@ -1939,13 +2109,44 @@ def optimize_battery_schedule(
...
+        #
+        # Re-derive exactly that one period per window, which is what the
+        # pre-Phase-3 replay did for every period. Its action and its
+        # `next_soe` are untouched by the splice; only the state it starts
+        # from moved.
+        boundary_ac_cap_kwh = _effective_ac_cap_kwh(battery_settings, dt)
+        for window in windows:
+            if window.end >= horizon:
+                continue
+            flows_trajectory[window.end] = _period_flows(
```
bess-agent, line 2188: **low — the splice-boundary re-derivation only fully fixes the IDLE case.**

For a room-limited STORE at `window.end`, `_period_flows` recomputes `battery_charged` from the *post*-splice `soe`, but `soe_trajectory[window.end + 1]` still reflects the *pre*-splice start. So the re-derived `battery_charged` no longer matches `battery_soe_end - battery_soe_start`, and `_price_flows` prices `energy_stored = (s2b + g2b) * eff` while dividing the cost basis by the stale `next_soe`.

Bounded by `_end_soe_pin_tolerance` and not caught by `assert_flow_coherence` (no SOE-delta check), but it means the comment's claim of restoring "what the pre-Phase-3 replay did" holds only for IDLE boundaries. Worth either re-deriving `next_soe` too, or narrowing the comment.

bess-agent (commented):
Code review of the Phase 3 consolidation. Both suites pass on this branch (1690 fast + 458 slow). I traced every removed derivation against its `_period_flows` replacement (STORE throttle ordering, `grid_imported += grid_to_battery` placement, IDLE `_idle_battery_flows` returning 0.0 discharge, `battery_action_kwh` branching, `create_decision_data` defaults vs. the hand-built `DecisionData` in `_create_idle_schedule`) and found no divergence. Window merging (`w.start <= last.end`) guarantees windows are never adjacent, so the new boundary loop cannot clobber a re-solved record.

Three findings inline: one medium (cost-basis direction), two low (latent divergence + a comment that overclaims).

core/bess/dp_battery_algorithm.py
```diff
@@ -1939,13 +2115,54 @@ def optimize_battery_schedule(
...
+        # `battery_soe_start` its own `battery_charged` cannot produce.
+        #
+        # Re-derive exactly that one period per window, which is what the
+        # pre-Phase-3 replay did for every period. Its action and its
+        # `next_soe` are untouched by the splice; only the state it starts
+        # from moved.
+        #
+        # This restores the flow record's agreement with the period's own
+        # *start* state, which is what the splice moved. It does not make a
+        # room-limited STORE boundary's `battery_charged` agree with
```
bess-agent: **medium — this comment has the case backwards, which defeats its purpose.**

It says the re-derivation does not make a **room-limited** STORE boundary's `battery_charged` agree with `battery_soe_end - battery_soe_start`. Room-limited STORE is precisely the case that self-corrects:

- `solar_to_battery + grid_to_battery = room_throughput = (max_soe - soe) / eff`, so `energy_stored = max_soe - soe`
- `_state_transition` clamps `next_soe = min(max_soe, soe + charge) = max_soe`

so `next_soe - soe == energy_stored` for *any* start SOE, including the spliced one.

The residual gap actually occurs when STORE is **rate-limited** (or throttled by `import_cap_kwh`), where `energy_stored` is SOE-independent while the stale `next_soe` still encodes the pre-splice start — the mismatch is then exactly the pin drift delta.

A maintainer auditing "is room-limited STORE still incoherent?" will find it is not, conclude the note is stale, and never learn that the rate-limited boundary is the one carrying the gap. Since this comment exists specifically to answer a review finding, the inverted case is worth fixing.

core/bess/tests/unit/test_cost_basis_real_bundle.py
```diff
@@ -0,0 +1,225 @@
...
+    )
+
+    # And it must not be priced BELOW what the retired formula charged for the
+    # same energy. The retired split assigned everything above solar to the
+    # grid at buy price, so pricing the remainder at cycle cost would have been
+    # cheaper than the bug -- understating the basis, the very failure this
+    # method was fixed for. This is the direction the first version of this
+    # test could not see, because it only compared against costing at zero.
+    retired = _replay_cost_basis(rows, cycle_cost, _retired)
+    assert actual >= retired, (
```
bess-agent: **medium — this guard is vacuous on this fixture.**

I reverted the production code to `unattributed * cycle_cost` and `test_real_day_has_charge_neither_source_explains` still **passed** — only the other two tests caught the revert.

The comparison is against the whole-day basis, which is dominated by the attribution change (20 differing periods) rather than by the 14 x 0.1 kWh remainder, so the remainder's pricing is invisible in it. The commit message claims "Both now assert the result also lands at or above what the retired split charged for the same kWh" — that holds only for `test_cost_basis_calculation.py`.

To actually pin it here, hold the attribution fixed and vary only the remainder's price: compare `_replay_cost_basis(_correct)` with the remainder at `cycle_cost` vs at `buy_price + cycle_cost`.

core/bess/tests/unit/test_cost_basis_real_bundle.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def test_real_day_has_charge_neither_source_explains(base_system):
+    """The unattributed-charge case is real, not hypothetical.
+
+    Independent lifetime counters do not reconcile exactly, so
+    `EnergyData.grid_to_battery` -- capped by the grid counter's own reading --
+    can leave part of a recorded charge unexplained. Every such kWh still
+    divides into the running total, so leaving it uncosted drags the basis
+    down. It is priced at cycle cost, as this method already prices the
```
bess-agent: **low — stale docstring, stating the exact position this commit retracts.**

"It is priced at cycle cost, as this method already prices the pre-existing charge of unknown provenance." The code it documents now prices it at `buy_price + cycle_cost`, and the body of this same test asserts against that.

core/bess/tests/unit/test_cost_basis_real_bundle.py
```diff
@@ -0,0 +1,225 @@
...
+                energy=energy,
+                economic=EconomicData(buy_price=row["buy_price"]),
+                timestamp=datetime.now(tz=TIMEZONE),
+                data_source="actual",
+                decision=DecisionData(),
+            ),
+        )
+
+
+def _replay_cost_basis(rows, cycle_cost, attribute):
```
bess-agent: **low — `_replay_cost_basis` is a mirror, so the equality assertion pins less than the commit message claims.**

It reproduces the loop in `_calculate_initial_cost_basis` line-for-line (same seeding, same `min(discharged, energy)`, same `<= 0.1` reset, same pricing terms). `assert actual == approx(expected, rel=1e-9)` therefore cannot detect a formula error — only an attribution regression, which the `retired != expected` guard already covers.

"Pinned against 73 actual recorded periods" overstates it; a literal expected value (e.g. the stated 0.9875 SEK/kWh) would pin the arithmetic, the mirror does not.

core/bess/dp_battery_algorithm.py
```diff
@@ -274,6 +273,159 @@ def _ac_flows_grid(
...
+    solar_production: float,
+    battery_settings: BatterySettings,
+    dt: float,
+    import_cap_kwh: float | None = None,
+) -> PeriodFlows:
+    """Derive one candidate action's complete flow set -- the only place a
+    planned period's *reported and priced* flows are computed.
+
+    Not the only place the charge split appears, and the difference matters
+    if you are about to edit it: `_state_transition` carries its own copy to
```
bess-agent, line 321: **low — the invariant this PR just enforced is not extended to `_state_transition`.**

"The AC cap must be derived one way so the producer and pricer of a flow record cannot disagree" now holds for `_period_flows`/`_price_flows`, but `_state_transition` still takes `ac_cap_kwh: float | None = None` and computes the `next_soe` that `_period_flows` consumes. Both production call sites (`dp_battery_algorithm.py:1561`, `simulation/inverter_simulator.py:237`) omit the argument.

Harmless today only because `_state_transition` uses the cap solely inside its `if import_cap_kwh is not None` branch and neither site passes an import cap. The moment either does — `_create_idle_schedule` gaining fuse-cap awareness, or the simulator being handed `import_cap_kwh` — `next_soe` would be computed with `cap=None` while `_period_flows` derives the real cap, reintroducing the exact divergence class P4 removed, with nothing in the signature to make it visible.

core/bess/battery_system_manager.py
```diff
@@ -2980,21 +2980,59 @@ def _calculate_initial_cost_basis(self, current_period: int) -> float:
...
+                # fix) is strictly worse than the formula being replaced here:
+                # the retired split assigned every kWh above solar to the grid
+                # at buy price, so on a period with an under-reading counter
+                # that "fix" priced the energy CHEAPER than the bug did --
+                # understating the basis, which is the exact failure Task 4
+                # exists to close. Degenerately, a reset grid counter during
+                # grid charging would have booked the whole charge as nearly
+                # free. Caught in review on this repo's own evidence bundle,
+                # `historical_2026_07_18_charge_attribution.json` period 39.
+                unattributed = max(
```
bess-agent, line 3059: **low — the unattributed estimator is asymmetric, and the asymmetry now costs money.**

`EnergyData.grid_to_battery = min(charged - solar_to_battery, grid_imported)`, so a grid counter that **under**-reads produces a positive residual priced at `buy_price + cycle_cost`, while a counter that **over**-reads is absorbed into `grid_to_home` and produces nothing. Combined with the `max(0.0, ...)` clamp, every counter-resolution rounding event can only push the basis up, never down.

On the pinned day this is +0.003 SEK/kWh (0.9844 -> 0.9875), but it scales with the number of charging periods and with `buy_price`, and it applies even to a system charging purely from solar.

Not a regression against `main` (the retired formula priced the same energy at buy price too, uncapped), and the direction is the conservative one — flagging so the bias is a known, bounded property rather than a surprise if a user's counters drift more than this bundle's.

bess-agent (commented):
Re-review after the three prior findings were addressed. Each was checked empirically (revert the fix, confirm a test fails), not just read.

**Prior findings**

1. **Unattributed charge pricing — fixed and genuinely pinned.** Reverting to `unattributed * cycle_cost` makes `test_real_day_cost_basis_uses_the_derived_charge_split` and `test_charge_the_split_cannot_explain_is_still_costed` fail. The new split is per-period monotonically >= the retired one, so the basis can only move up.
2. **`ac_cap_kwh` divergence — addressed differently, and correctly.** `_period_flows` now derives the cap internally too, so producer and pricer cannot disagree by construction. Every former caller passed exactly `_effective_ac_cap_kwh(battery_settings, dt)`; fast suite green (1690 passed).
3. **Splice-boundary stale `next_soe` — not fixed, documented instead — and the comment names the wrong case.** See inline. The re-derivation itself does discriminate: neutering the `for window in windows` loop makes `test_splice_boundary_flows.py` fail.

Six new findings inline: two medium, four low.

core/bess/battery_system_manager.py
```diff
@@ -2980,21 +2980,71 @@ def _calculate_initial_cost_basis(self, current_period: int) -> float:
...
+                # having charged the battery -- booking grid energy as free.
+                # Measured over the fixture corpus before the fix: 31 of 419
+                # charging periods disagreed, 36.16 kWh of grid charging
+                # booked as solar across 10 of 36 fixtures, worst single
+                # period 5.2 kWh (solar 5.8, home 5.2, charged 6.0 -- the
+                # whole 5.8 kWh of solar counted to the battery while the home
+                # was consuming 5.2 of it). The understated basis then fed
+                # `optimize_battery_schedule(initial_cost_basis=...)`, making
+                # stored energy look cheaper to discharge than it was.
+                solar_to_battery = event.energy.solar_to_battery
```
bess-agent, line 2996: **medium — solar/grid split inverts during deliberate grid charging**

This replaces `min(battery_charged, solar_production)` with `EnergyData.solar_to_battery`/`grid_to_battery`. That split allocates solar to the **home first**, which is right for surplus charging but inverted during deliberate grid charging (`battery_first`), where PV is DC-coupled straight to the battery and the house runs off the grid.

Verified against the live model:

```
EnergyData(solar=2.0, home=2.0, battery_charged=3.0, grid_imported=3.0)
  -> solar_to_battery=0.0, grid_to_battery=3.0
```

All 3 kWh get booked at `buy + cycle` (basis 1.40 at buy=1.0 / cycle=0.4) when 2 kWh were free PV (true basis 0.73). The retired formula gave 2.0/1.0 and was the accurate one in this regime.

The safety argument in the comment ("can only push the basis up") holds directionally, but the overstatement here is large, not a rounding-band effect, and midday grid-charging with concurrent PV is a normal DP outcome. Worth either bounding it or extending the comment — it currently only reasons about the opposite case.

core/bess/dp_battery_algorithm.py
```diff
@@ -1400,70 +1583,58 @@ def _create_idle_schedule(
...
+            dt=dt,
         )
-
-        decision_data = DecisionData(
-            strategic_intent=classify_strategic_intent(0.0, energy_data),
-            battery_action=0.0,
+        # Cost basis only. The reward is discarded: this schedule's cost is
+        # summed from the reported PeriodData below, exactly as before, and
+        # the guardrail comparison reads that sum -- not this term.
+        _reward, current_cost_basis = _price_flows(
```
bess-agent, line 1607: **low — reported `cost_basis` does move on AC-capped systems**

`_create_idle_schedule` now routes its cost basis through `_price_flows`, adopting the AC-clip-discounted solar opportunity cost (measured at 0.049977 SEK/kWh on one fixture).

The guardrail *decision* is unaffected — it compares `battery_solar_cost`, which does not read `cost_basis`. But when the all-IDLE net wins, this is the returned schedule, so every period's reported `decision.cost_basis` shifts on AC-capped systems. The PR description says "nothing observable moves"; the reported field does move, just not any comparison. Documentation-level, not a defect.

Issue #538, opened 2026-08-10
E2E never exercises a Growatt VPP register write — control scenario has no VPP entities


bess-agent:
## Problem

No E2E test has ever observed the backend writing a Growatt VPP register. Found while verifying #537 (the VPP half of #520) — Step 8's local run logged `Sensor growatt_vpp_remote_control (entity_id: select.growatt_min_vpp_remote_control) is unavailable`, so no VPP write happened at all.

The VPP entities and the control scenario both exist. They have just never been in the same place:

| Fixture | VPP control entities | Used for |
|---|---|---|
| `scripts/mock_ha/scenarios/ci-growatt-vpp.json` (16 sensors) | **none** | the control E2E — nothing to write to |
| `ci-wizard-growatt-vpp.json`, `ci-wizard-growatt-vpp-ridax-118.json` (75 entities, incl. `select.growatt_inverter_vpp_remote_control`, `input_number.vpp_bess_power`) | yes | the **wizard** loop in `e2e/run-e2e.sh` only — setup/discovery |
| `e2e/tests/inverter-schedule-control-model.spec.ts` | mocked in-browser via `mockInverterApis(page, 'vpp_power', …)` | frontend column rendering |

So the wizard proves discovery works, the Playwright spec proves the UI renders, and nothing proves a period apply reaches the hardware.

## What this does and does not invalidate

**Not vacuous.** The unit tests (`test_vpp_discharge_gate_capability.py`, `test_solax_modbus_growatt_vpp.py`, `test_vpp_load_support_gate_520.py`) drive real `_apply_period_schedule` and assert the actual command values through `MockHomeAssistantController`. The intent→command *mapping* is genuinely covered — #537's RED test fails against the pre-fix mapping, so it discriminates.

**What is uncovered** is everything between the mapping and the wire:

- entity resolution for VPP control entities on a control scenario
- the real HA service-call shape for `select`/`number` writes
- the 15-minute refresh that stops the inverter's dead-man timer lapsing (#404) — a live behaviour a real tester reported, exercised by no test outside unit level
- the "skip the write when remote control is already disabled" path in `_apply_period_vpp`

The logic is verified; the delivery is not.

## Suggested fix

Add the VPP control entities to `ci-growatt-vpp.json` so the control E2E can observe register writes, then assert on them.

[...]

Issue #520, opened 2026-08-09
Apply the intra-period discharge gate to LOAD_SUPPORT on both platforms (settles #384/#393)


johanzander:
> [!IMPORTANT]
> **Status header — maintained by `bess-agent`, last updated 2026-08-13. The original issue text below is unedited and partly superseded; read this first.**
>
> | half | status |
> |---|---|
> | **TOU** | **Done** — #524 merged. |
> | **VPP** | **Open, and the design below is withdrawn.** |
>
> **Two corrections to the text below:**
>
> 1. The **"What to change → VPP"** section (gate closed → do not release control) **is withdrawn.** On TOU a closed gate still delivers the planned discharge and only declines to raise the ceiling; on VPP that mapping delivers nothing. All 172 gate-closed LOAD_SUPPORT periods in the 36-fixture corpus carry a real planned discharge — 118.11 kWh the design would abandon. Caught by @ridax67. Draft PR #537 implemented it and has been closed.
> 2. The **"9 periods — 1.5%"** figure does not reproduce (see the first comment below).
>
> **Current direction for the VPP half:** track `vpp_power` against measured house load, so the commanded power is the net intent rather than a fixed figure real load eats into. Unblocked 2026-08-11 when @ridax67 supplied the Growatt VPP protocol V2.01 document confirming `vpp_power` / `vpp_time` / `vpp_remote_control` are RAM registers, making a frequent tracking loop free of flash-wear cost. Also fixes #352. Sits in Phase 4 of the optimizer migration.
>
> **Not part of this issue:** @Frank-Leysen's 2026-08-13 hardware report is a separate value-function defect, now tracked as **#571** — the discharge gate reads a raw snapped `V` grid difference while the policy uses `_interpolate_value`. The one part of that report which does belong here is the load forecast gap (1.96 kW forecast vs 4.8 kW actual), which the load-tracking work above addresses.

---

## The question

House load exceeding forecast within a period is not an edge case — it is the normal condition. The open question is what should happen when it does, and today the answer depends on which inverter you own.

There is one extra kWh to source. From the battery it costs the stored energy's value (`shadow_price`); from the grid it costs `buy_price`. Battery wins exactly when:

```
buy_price × efficiency_discharge >= shadow_price
```

That expression contains only prices and the plan. **Nothing in it is platform-specific.** Same house, same prices, same plan should give the same behavior on VPP and TOU. It does not.

## Current state — each platform is missing the opposite half

[...]

PR #541, opened 2026-08-10
feat: simulate Growatt VPP and pin it against v10.0.2 (#539)

bess-agent:
## Summary

Growatt VPP had no simulation, so "no regression from beta" was unenforceable for it — a requirement with no instrument behind it. This adds one, pinned against the released tag.

- `simulation/vpp_simulator.py` — the VPP execution model
- `tests/unit/vpp_capture.py` + `scripts/capture_vpp_baseline.py` — one shared definition of "run this fixture and record what VPP would do", same arrangement as `golden_capture.py`
- `test_vpp_regression_baseline.py` — the pin

Closes the gap that blocks #537.

## The command mapping is not re-implemented

`derive_vpp_commands` drives the production path end to end — `compute_rates_for_period` → `_intent_to_vpp`, the same two calls `_apply_period_schedule` makes. Nothing about the intent → rates mapping is restated.

*Corrected during review:* the first revision rebuilt those rates from `INTENT_TO_CONTROL` by hand — a second copy of a mapping production owns, inside the harness whose whole premise is that it cannot describe a command production would not send. It agreed on every intent (baseline byte-unchanged after the refactor), but it also carried a `.get(intent, IDLE)` fallback where production raises, so a typo'd intent would have silently simulated IDLE.

## Why the baseline is "v10.0.2 plans, today's execution model"

Not today's plans. The split holds the execution model fixed so a failure isolates a *planner* change — and it's the only way to reach the tag at all: `vpp_simulator` needs `_period_flows`, which arrived with Phase 3, so the tagged code cannot run today's simulator.

Regenerating therefore takes a tag checkout and two steps, documented at the top of `scripts/capture_vpp_baseline.py`.

## Measured while building it

Plans have **already moved on 35 of 36 fixtures** since v10.0.2 — Phase 2's preference table, #512's finer grid, #524's TOU gate, #526's authorization. `historical_2025_01_05_no_spread_no_solar` is the single fixture still planning exactly as the release does (re-verified after the capture fix).

So the pin cannot assert equality with the release. It pins two things instead: v10.0.2's plan executes to the recorded commands and cost (the execution model), and **today's plan matches what was recorded** (the planner, on every fixture). The drift between them is reported as a single number rather than left to be rediscovered.

## Verified it detects real changes, in both directions

- **Execution-model change** — applying #537's mapping (`LOAD_SUPPORT` → `1, True`) fails **68** tests.
- **Planner-side change** — perturbing `SOE_STEP_KWH` fails **35** tests via the current-plan pin. The first revision of this PR could not detect this at all (see below).

That is the instrument #537 was blocked for: it can now quote a measured delta instead of "effect unknown".

## Three defects found in review, all measured before fixing

The first revision's baseline was invalid on two independent grounds:

1. **The `+1%` hold was simulated as a full-rate grid charge.** STORE physics are binary — any power above `POWER_TOLERANCE_KW` charges at the full rate and the magnitude is never read. Measured: `_state_transition(soe=3.0, power=0.06)` and `power=6.0` both return 9.0. Every IDLE period encoded fictional grid import; on the **no-solar** fixture `historical_2024_08_16` the first two IDLE periods each added 6 kWh (soe 3.0 → 9.0 → 15.0), and realized cost was 243.88 instead of **230.47**. A hold now returns 0.0 so the IDLE branch runs; rates the binary physics cannot express raise rather than mis-simulate.

2. **The capture reverted the scenario fixtures along with the code.** `core/bess/tests/unit/data` sits under `core/bess`, and 70 fixture files changed since v10.0.2 — including `price_data`, from which `_scenario_inputs` derives buy/sell prices. Plans were optimized on the tag's prices and executed on today's. Measured: **15 of 36** recorded v10.0.2 plans were wrong. The checkout now excludes that directory.

3. **The plan pin was vacuous on 35 of 36 fixtures.** It asserted only *which* fixtures had already diverged from v10.0.2 — and since 35 had, any further regression kept them in the same set and passed silently. The baseline now records today's plan per fixture and pins it.

## Read a delta as "changed", never "worse"

At 15-minute point forecasts there is no within-period load spike, so this models the intra-period gate's **cost** but never its **benefit**. It will score gate-closed changes as a loss whether or not they are one.

Not speculative — it's the same structural blindness that makes the TOU simulator deliberately *not* mirror the gate (mirroring moves 27/36 fixtures by +25.67 SEK of pure artefact). Against a fixed baseline the bias cancels and a delta means *changed*, which is the question. Stated in both module docstrings, because quoting such a delta as an economic verdict is a misuse this repo has already retracted figures for.

## One trap worth recording

The baseline lives in `data/baselines/`, not `data/`. `golden_capture.fixture_names()` globs `data/*.json`, so a file dropped in the parent silently becomes a 37th scenario and changes what the bit-parity gate means. Caught during development — the corpus did briefly read 37.

## Test plan

- [x] `./scripts/quality-check.sh` — 0 errors
- [x] 1690 fast + 532 slow passed
- [x] Harness verified to fail on both an execution-model change (68 tests) and a planner change (35 tests), then restored
- [x] Golden corpus confirmed still 36 after adding the baseline

## Not covered

Real hardware. A simulation proves the model is self-consistent, not that the inverter obeys it. Confirming `vpp_power=+1` actually holds the battery still needs a tester on `control_mode=vpp`. #538 (E2E never writes a VPP register) remains the complementary gap at the delivery layer.

The `grid_first` semantics are modelled per #118/#466 (holds against charging, still serves load) rather than #355's reading, because that one is backed by a real-hardware report. Stated in the code: if it's wrong, baseline and comparison inherit it equally, so regression detection is unaffected and only an absolute cost reading would be.

---

## Added after review: three guards from the pre-release audit

The branch picked up four commits from Pass 2 of the pre-release validation (`docs/pre-release-validation`), because each one is a test that belongs with this harness rather than in a docs branch.

### `test_platform_mapping_fidelity.py` — how much of a plan each platform can execute

Written after **#537 reached review with a design defect**. It mapped #520's closed discharge gate onto VPP as a `battery_first` hold. On TOU, gate-closed still delivers the planned discharge and merely declines to raise the ceiling; on VPP a hold delivers nothing.

| | |
|---|---|
| LOAD_SUPPORT periods | 603 |
| gate closed | 172 (28.5%) |
| ...carrying a planned discharge | **172 — every one** |
| planned energy #537 would abandon | **118.11 kWh** |

The inference behind it: VPP carries BATTERY_EXPORT's planned magnitude faithfully (`power_pct` is the plan-scaled rate, negated), so it looked as though VPP could express "discharge, but only this much" generally.

| intent | planned rates | VPP commands | |
|---|---|---|---|
| IDLE / SOLAR_STORAGE / SOLAR_EXPORT / GRID_CHARGING | 1 | 1 | not rate-bearing |
| BATTERY_EXPORT | 101 | 101 | faithful |
| **LOAD_SUPPORT** | **101** | **1** | **lossy** |

LOAD_SUPPORT is the only lossy cell, and it sits directly beside the one intent that generalises the wrong way. The test sweeps the full planned-action range per intent and fails if any undeclared intent becomes lossy. Verified discriminating: clearing `VPP_LOSSY_INTENTS` fails 2 of 8.

**Honest limit:** it does not itself fail on #537 — that parameter does not exist on this branch, so the sweep cannot reach the gate-closed path. It pins the asymmetry that caused the wrong inference, and needs extending to sweep gate states when #537 is redesigned.

Nothing in the suite could have seen this: the golden corpus, `test_scenarios` and the R==P checks all pin what the DP *plans*. Executable fidelity was never measured. The VPP simulator in this PR cannot adjudicate it either — by its own docstring it scores gate-closed changes as a loss whether or not they are one. What caught it was @ridax67's comment on #520.

### `#399` — the guard was a proxy

`test_seeds_state_from_hardware` asserted `_vpp_status_confirmed` is seeded from the read-back. That is the mechanism, not the write count ridax reported, and it stays green if the flag is seeded and then ignored. #329's write-count test never reaches the read-back path, because that instance already set the flag from its own first write — so the restart case, which is what #399 is about, had no outcome coverage.

Added: a fresh controller reads back an already-Enabled inverter, commands a period, and both flash registers must see zero writes — plus a positive assertion that the period command still goes out, so it cannot pass by the controller doing nothing.

### `#302` — no guard at all

Deleting the DST end-time cap in `_groups_to_tou_intervals` left **all 1714 fast tests green**. With it removed the interval is emitted as `24:59`, which is what HA's `select_option` rejected out from under @Frank-Leysen. The fall-back day comes once a year, so a refactor could drop the cap in September and the first signal would be a user's inverter failing on the changeover night.

Added: asserts the emitted interval's `start_time`/`end_time` are valid wall-clock values — what reaches hardware, not the capping branch.

### Also: `fix: make the scenario corpus able to see what it pins`

`test_scenarios` rebuilt the optimizer call by hand and dropped `export_curtailment_active`, `initial_cost_basis` and `home_settings`, so the one fixture captured to pin #510's tie-break *under* the export-curtailment price floor was asserted with curtailment off. Now passes `_scenario_inputs` through whole, as `golden_capture.py` already did, and tightens the cost tolerance from 0.1 to 0.001 SEK. Re-pinned: −2.74447 → −2.765873 (the correct-inputs solve was always the cheaper one).

Corpus discriminating power, measured: removing ridax's `_row_load_covering_discharge` previously failed **0** tests in `test_scenarios`; it now fails **18**.

## Updated test plan

- [x] `./scripts/quality-check.sh` — 0 errors
- [x] **1715** fast + 532 slow passed
- [x] Each added guard observed to fail with its fix reverted, then restored


core/bess/simulation/vpp_simulator.py
```diff
@@ -0,0 +1,272 @@
...
+
+    if command.power_pct < 0:
+        # Forced discharge/export at the commanded rate, regardless of load
+        # (#324) -- this is the behaviour that makes the gate's TOU ceiling
+        # semantics wrong here.
+        rate_kw = settings.max_discharge_power_kw * abs(command.power_pct) / 100.0
+        delivered = min(
+            rate_kw * dt, available * settings.efficiency_discharge, ac_headroom_kwh
+        )
+        return -delivered / dt
```
bess-agent, line 230: **Zero-delivery forced discharge silently becomes a passive solar _charge_.** (medium)

When `available == 0` (SoE at min) or `ac_headroom_kwh == 0`, `delivered` is 0 and this branch returns `-0.0`, which is not `< -POWER_TOLERANCE_KW`. So `_state_transition`/`_period_flows` take the **IDLE** branch and absorb solar surplus into the battery — under a `power_pct < 0` grid_first forced-export command, and in direct contradiction of this same file's `power_pct == 0` branch (which models grid_first as holding the battery against charging) and of `BATTERY_EXPORT`'s `charge_rate = 0`.

Reachable and now pinned into the baseline: on `realworld_2026_04_29_220919` period 37 the command is `(-1, True)`, SoE 2.5 (== min), solar surplus 1.1 kWh, and the simulator charges to 3.567 kWh. Realized cost is -92.12 with this behaviour vs -93.75 if the period held (1.63 SEK). Also occurs on `regression_2026_08_08_143843` and (harmlessly, no surplus) `realworld_2026_04_22_202249`.

Same defect class as the `+1%`-hold bug this PR fixes. Returning `None` when `delivered <= 0` matches the file's own hold convention.

core/bess/simulation/vpp_simulator.py
```diff
@@ -0,0 +1,272 @@
...
+
+    # power == 0, remote control enabled: grid_first. Battery is held against
+    # charging from solar surplus (#355), but still serves load (#118).
+    if deficit > 0:
```
bess-agent, line 243: **The grid_first "still serves load" branch is dead across the entire corpus.** (low)

Replacing `if deficit > 0:` with `if False:` leaves all 74 tests passing — no SOLAR_EXPORT period in any of the 36 fixtures has `home > solar`. The #118/#466-vs-#355 modelling choice the PR description argues at length is therefore entirely unpinned: a future change to it (or #537's revisiting of grid_first) produces zero test signal, contrary to the docstring's implication that the baseline captures the choice.

Worth either a targeted unit test or an explicit note that this branch is uncovered.

core/bess/tests/unit/test_vpp_regression_baseline.py
```diff
@@ -0,0 +1,132 @@
...
+        "and state the measured VPP delta in the PR."
+    )
+    replayed = simulate_plan(name, entry["current_plan"])
+    assert replayed["commands"] == entry["current"]["commands"]
+    assert replayed["realized_cost"] == pytest.approx(
+        entry["current"]["realized_cost"], abs=1e-9
+    )
+
+
+def test_drift_from_the_released_version_is_recorded():
```
bess-agent, line 126: **This test executes no production code.** (low)

It compares `baseline[name]["current_plan"]` against `baseline[name]["plan"]`, both read from the same JSON file, so `moved` is a constant of the checked-in artefact and `len(moved) == 35` can only fail if someone hand-edits the baseline.

The docstring's claim that "a new divergence shows up as a new name" is false — `current_plan` is stored, never recomputed here (the recompute lives in `test_current_plan_is_pinned`). It is documentation shaped like a test.

core/bess/tests/unit/test_vpp_regression_baseline.py
```diff
@@ -0,0 +1,132 @@
...
+    """How far today's plans have moved from v10.0.2, as a single visible
+    number rather than something to rediscover.
+
+    Deliberate changes since the tag (Phase 2's preference table, #512's
+    finer grid, #524's TOU gate, #526's authorization) already moved most of
+    the corpus, so this is a recorded fact, not a pass/fail on equality.
+    """
+    baseline = _baseline()
+    moved = [
```
bess-agent, line 158: This indexes `baseline[name]` for every name in `fixture_names()`, so a fixture added without a baseline entry raises a bare `KeyError` here instead of the actionable "regenerate per `scripts/capture_vpp_baseline.py`" message the other three tests take care to emit. (low)

core/bess/tests/unit/test_vpp_regression_baseline.py
```diff
@@ -0,0 +1,132 @@
...
+        "`scripts/capture_vpp_baseline.py`."
+    )
+    entry = baseline[name]
+
+    assert capture_plan(name) == entry["current_plan"], (
+        "the DP now plans this fixture differently. If deliberate, re-pin "
+        "and state the measured VPP delta in the PR."
+    )
+    replayed = simulate_plan(name, entry["current_plan"])
+    assert replayed["commands"] == entry["current"]["commands"]
```
bess-agent, line 114: **`test_current_plan_is_pinned` omits the SoE trajectory** that `test_vpp_execution_of_the_baseline_plan_is_unchanged` (L78) asserts. (low)

Since this is the half that covers today's plans on all 36 fixtures, an execution-model change that shifts the SoE path but nets out in commands and total cost (e.g. a charge/discharge efficiency-application change, or the zero-delivery finding above) escapes the stricter half of the pin on current behaviour. Adding the `soe_trajectory` assertion is free — `simulate_plan` already returns it.

core/bess/tests/unit/vpp_capture.py
```diff
@@ -0,0 +1,84 @@
...
+    )
+    return {
+        "commands": [[c.power_pct, c.remote_control_enabled] for c in commands],
+        "realized_cost": sim.realized_cost,
+        "soe_trajectory": [inputs["initial_soe"]]
+        + [p.energy.battery_soe_end for p in sim.period_data],
+    }
+
+
+def capture_fixture(name: str) -> dict:
```
bess-agent: `capture_fixture()` is dead code: nothing imports it (the script uses `capture_plan` + `simulate_plan`, the test module imports neither), and the dict it builds (`plan` plus flattened command fields, no `current`/`current_plan`) no longer matches the baseline schema — so anyone who reaches for it as "the" capture entry point produces a baseline the new tests reject. (low)

bess-agent (commented):
## Code review

**Verification performed:** ran the new suite (74 passed, 11s), reproduced the regeneration recipe in a throwaway worktree (the `:(exclude)` pathspec works on git 2.50; recaptured v10.0.2 plans match the committed baseline on **all 36 fixtures**), and mutation-probed the new simulator branches.

The core claims hold up: `derive_vpp_commands` does drive production, the intra-period gate correctly does not apply to VPP (`discharge_rate_is_load_following` is False there), units match `run_scenario_realized`, the baseline dir sits outside `golden_capture`'s glob, and every command category (charge100 / +1 hold / grid_first hold / discharge / disabled) is present in the corpus.

One correctness finding worth fixing before merge, plus five test-quality notes — all inline below.

PR #561, opened 2026-08-13
chore: make the worktree the permission boundary so agent runs stop stalling

johanzander:
Long autonomous runs — `implement-issue` especially — kept halting on permission
prompts *inside a worktree*, for commands the setup was supposed to have already
cleared. Two independent causes, both silent.

## 1. The auto-allow hook was an enumeration, and enumerations lose

`auto-allow-worktree-destructive.sh` listed destructive command shapes (`rm -rf`,
`git reset --hard`, `git rebase`, `git merge`, `git push --force`) and auto-allowed
only those inside a linked worktree. Everything else fell through to
`ask: Bash(rm *)` in `settings.json`.

So `rm -f scripts/mock_ha/scenarios/bess-debug-huawei.json && .venv/bin/pytest …`
prompted — and **no allowlist entry could ever have fixed it**, because an `ask`
rule beats an `allow` rule. Each new command shape needed another hook entry, which
left the list permanently one command behind.

**Inverted.** cwd in a linked worktree now auto-allows Bash. A worktree is
disposable by construction; a prompt there protects nothing. Two classes still
prompt:

- **Globally scoped** — no worktree contains these: `sudo`, `podman machine
  rm`/`system reset` (the shared VM every worktree's E2E stack uses), `gh pr
  merge`/`release`/`repo delete`, and any push that moves a shared ref
  (`main`/`beta`, `--all`, `--mirror`, `--tags`, version tags).
- **Reaches outside the worktree** — an absolute path beyond its root, or a target
  hidden behind `..`, `~`, a variable, or a command substitution.

Both emit an explicit `ask`. Falling through would have been wrong in both
directions: `settings.json` *allow*-lists `Bash(git push *)`, so a shared-ref push
would sail through unprompted, while `podman machine rm` would hit a deny it can no
longer reach once any allow is in play.

The main checkout is untouched — the hook returns before any of this when cwd isn't
a linked worktree, so it keeps every `ask`/`deny` rule exactly as configured.

## 2. Local settings never travel into a worktree

Only *tracked* files follow a worktree. `.claude/settings.json` and
`.claude/hooks/*` do; `.claude/settings.local.json` is gitignored and so exists
**only in the main checkout**.

A session entering a worktree therefore drops every rule and mode set there —
including `defaultMode` — while the tracked `ask:` list keeps applying. A run that
was fully autonomous in the main checkout starts prompting the moment it moves.
**24 of 25 linked worktrees on this machine were missing the file.**

`link-worktree-local-settings.sh` symlinks it in, sweeping *all* linked worktrees so
the pass is idempotent, payload-agnostic, and self-healing for worktrees predating
it. Registered on `WorktreeCreate` (native creation; fires before settings load) and
`SessionStart` (covers `git worktree add`, the `using-git-worktrees` fallback, which
fires no `WorktreeCreate` — settings are already loaded by then, so such a worktree
picks its link up on its *next* session; the fleet-wide sweep is what makes that
self-correcting rather than permanent).

## What the first revision got wrong

The original version of this PR claimed the escape list excluded "commands whose
blast radius leaves the worktree." **That claim was false**, and `/code-review`
demonstrated it. The guard inspected only tokens beginning with `/`, so all of these
were auto-allowed from inside a worktree — each reproduced against the real hook:

| Auto-allowed (should not have been) | Why it slipped |
|---|---|
| `rm -rf ../../../../bess-manager-sibling` | relative traversal, never checked |
| `rm -rf "/Users/…/bess-manager/.venv"` | leading quote meant the token didn't start with `/` |
| `rm -rf $HOME/GitHub/bess-manager` | unexpanded variable |
| `rm -rf /tmp/../Users/…/.venv` | matched the `/tmp/*` allowance, then traversed out |
| `git -C /Users/…/bess-manager reset --hard` | not an `rm`/`mv` verb, so never scrutinized |
| `cp secrets.txt /Users/…/leak.txt`, `find /Users/… -delete` | same |

Also fixed from that review: the escape list matched bare substrings, so `grep
"sudo " docs/` and `git push origin fix/mainpage-bug` were wrongly pushed back to a
prompt (now anchored at command-word position); and `awk '/^worktree /{print $2}'`
truncates a checkout path at its first space — in `auto-allow` that **inverted the
boundary**, making the "am I the main checkout?" test fail *in the main checkout*
and auto-allowing everything there.

## Honest limits

The containment check guards **accidents** — a stale absolute path from before a
worktree switch, a `../..` from the wrong cwd — as the Bash-side counterpart of
`check-worktree-path.sh`. It is **not a sandbox**: a hook sees the command string,
not the syscalls, so indirection can defeat it. The real containment is that a
worktree is disposable. CLAUDE.md says so, so this isn't re-litigated as either a
security boundary or a pointless one.

## Verification

Expectation-checked matrix against the real hook:

- **Inside a worktree — 39/39.** 18 ordinary commands allowed (`pytest`, `npm ci`,
  `git merge`, `rm -f`, `git push origin fix/mainpage-bug`, `grep "sudo " docs/`),
  9 globally-scoped → `ask`, 12 escape attempts → `ask`.
- **From the main checkout — 9/9 fall through.** Strictly worktree-scoped.

Link sweep produced 24 symlinks and correctly skipped the one worktree carrying a
real (non-symlink) local settings file. `./scripts/quality-check.sh`: **0 errors**
(3 warnings are the script probing `PATH` for pytest/black/ruff instead of
`.venv/bin` — pre-existing; this diff contains no Python).

## Scope

`.venv` and `frontend/node_modules` have the identical doesn't-travel problem — this
worktree started with neither, which is what made `quality-check.sh` fail on first
run. Left to `scripts/worktree-setup.sh` (#556) rather than duplicated here.

Two review findings are knowingly **not** addressed, as judgement calls rather than
defects: the symlink is bidirectional (a `/permissions` edit from inside a worktree
writes through to the main checkout — that is the intended single source of truth),
and `WorktreeCreate` ordering is assumed rather than proven, since it can't be
exercised from inside an existing worktree session. The `SessionStart` registration
and fleet-wide sweep are what bound the cost of that assumption being wrong.

https://claude.ai/code/session_01W29UVDwQPwVRRqBn2sMgHa

### DIFF
--- a/.claude/hooks/auto-allow-worktree-destructive.sh
+++ b/.claude/hooks/auto-allow-worktree-destructive.sh
@@ -21,12 +21,20 @@
 # prompted, and every new command shape needed another entry. Inverting it
 # removes that whole class of stall.
 #
-# Two exceptions never get auto-allowed:
+# Every check below therefore fails CLOSED: where a pattern list is
+# unavoidable it enumerates what is SAFE, so a shape nobody thought of
+# costs one prompt rather than a silent auto-allow.
 #
-#   1. Globally-scoped actions (sudo, podman machine rm, gh release, ...) --
-#      matched at command-word position, not as a bare substring, so that
+# Three exceptions never get auto-allowed:
+#
+#   1. Globally-scoped actions (sudo, the shared podman VM, anything
+#      touching GitHub, a push that can move a shared ref) -- matched at
+#      command-word position, not as a bare substring, so that
 #      `grep "sudo " docs/` is not mistaken for running sudo.
-#   2. Commands that can reach OUTSIDE the worktree. Those are put through
+#   2. Mutations of state SHARED with the main checkout. A linked worktree
+#      has its own working tree but ONE object database and ONE ref
+#      namespace with every other checkout (see mutates_shared_state).
+#   3. Commands that can reach OUTSIDE the worktree. Those are put through
 #      a stricter check (see needs_scrutiny / escapes_worktree below).
 #
 # THREAT MODEL: accidents, not an adversary. The damage this prevents is a
@@ -53,25 +61,67 @@ fi
 # `grep "sudo " docs/` as running sudo and reintroduce the prompt.
 CMD_START='(^|[;&|(]|&&|\|\||\n)[[:space:]]*'
 
+# --- Hard denials -------------------------------------------------------
+#
+# `podman machine rm` / `podman system reset` are `deny` entries in
+# settings.json -- a hard block on destroying the shared podman VM every
+# worktree's E2E stack runs on. A hook decision SUPERSEDES a settings rule,
+# so answering "ask" here would quietly downgrade that block to a single
+# keystroke. Re-emit the deny instead of weakening it.
+is_hard_denied() {
+  printf '%s' "$1" | grep -qE "${CMD_START}podman[[:space:]]+(machine[[:space:]]+(rm|reset)|system[[:space:]]+reset)"
+}
+
 # --- Globally-scoped actions -------------------------------------------
 #
-# Effects that no worktree can contain: elevated privileges, the shared
-# podman VM every worktree's E2E stack runs on, a GitHub mutation, or a
-# push that moves a shared ref.
+# Effects that no worktree can contain: elevated privileges, anything that
+# reaches GitHub, or a push that can move a shared ref.
 #
-# These get an explicit "ask" rather than a fall-through. Falling through
-# would be wrong in both directions: settings.json ALLOWS `Bash(git push *)`,
-# so a shared-ref push would sail through unprompted, while `podman machine
-# rm` would hit a deny it can no longer reach once any allow is in play.
+# These get an explicit "ask" rather than a fall-through, because
+# settings.json ALLOWS `Bash(git push *)` and `Bash(gh *)` -- falling
+# through would let them run unprompted rather than prompt.
 is_globally_scoped() {
   printf '%s' "$1" | grep -qE "${CMD_START}sudo[[:space:]]" && return 0
-  printf '%s' "$1" | grep -qE "${CMD_START}podman[[:space:]]+(machine[[:space:]]+(rm|reset)|system[[:space:]]+reset)" && return 0
-  printf '%s' "$1" | grep -qE "${CMD_START}gh[[:space:]]+(pr[[:space:]]+merge|release|repo[[:space:]]+(delete|archive))" && return 0
-  # Any push that can move a shared ref: main/beta branches, tags, --all,
-  # --mirror. NOTE these must be an explicit deny, not a fall-through --
-  # settings.json allow-lists `Bash(git push *)`, so falling through would
-  # allow them outright rather than prompt.
-  printf '%s' "$1" | grep -qE "${CMD_START}git[[:space:]]+push([[:space:]]|$).*([[:space:]](main|beta)([[:space:]]|$)|--all|--mirror|--tags|[[:space:]]v[0-9])" && return 0
+
+  # `gh` reaches GitHub, which no worktree contains. Enumerating destructive
+  # subcommands fails the wrong way: `gh api -X DELETE repos/...` is the
+  # general form of every entry such a list could hold, and `gh workflow
+  # run` / `gh secret set` / `gh repo edit` are not obviously "destructive"
+  # names. So invert -- every `gh` asks unless it is a read-only query.
+  # Counting both forms catches a compound `gh pr view && gh pr merge`,
+  # where matching only the first invocation would clear the whole command.
+  local gh_all gh_readonly
+  # `grep -c` counts LINES, and a command is normally one line, so both
+  # counts would be 1 for that compound. Count occurrences with -o | wc -l.
+  gh_all=$(printf '%s' "$1" | grep -oE "${CMD_START}gh[[:space:]]" | wc -l | tr -d ' ' || true)
+  gh_readonly=$(printf '%s' "$1" | grep -oE "${CMD_START}gh[[:space:]]+(pr[[:space:]]+(view|list|diff|checks|status)|issue[[:space:]]+(view|list)|run[[:space:]]+(view|list|watch)|release[[:space:]]+(view|list)|repo[[:space:]]+view|workflow[[:space:]]+(view|list)|label[[:space:]]+list|search|auth[[:space:]]+status)([[:space:]]|$)" | wc -l | tr -d ' ' || true)
+  [ "$gh_all" -ne "$gh_readonly" ] && return 0
+
+  # Any push that can move a shared ref. The ref has to be matched in every
+  # refspec form -- `main`, `HEAD:main`, `br:refs/heads/main`, `+main` --
+  # not just as a bare word, since the bare word is the spelling least
+  # likely to appear when something force-updates a ref. Force pushes ask
+  # unconditionally: settings.json already asks for them, and this hook's
+  # explicit allow would otherwise override that.
+  if printf '%s' "$1" | grep -qE "${CMD_START}git[[:space:]]+push([[:space:]]|$)"; then
+    printf '%s' "$1" | grep -qE "(--force|--force-with-lease|[[:space:]]-f([[:space:]]|$)|--all|--mirror|--tags)" && return 0
+    printf '%s' "$1" | grep -qE "(^|[[:space:]:+])(refs/heads/)?(main|beta)([[:space:]]|:|$)" && return 0
+    printf '%s' "$1" | grep -qE "(^|[[:space:]:+])(refs/tags/)?v[0-9]" && return 0
+  fi
+  return 1
+}
+
+# --- Shared repository state --------------------------------------------
+#
+# A linked worktree has its own working tree, but shares ONE object
+# database and ONE ref namespace with the main checkout and every other
+# worktree. So these reach far outside the worktree even without `-C`:
+# deleting a branch another agent is on, expiring the reflog that would
+# have recovered it, or removing another agent's checkout outright.
+mutates_shared_state() {
+  printf '%s' "$1" | grep -qE "${CMD_START}git[[:space:]]+(branch|tag)[[:space:]]+(.*[[:space:]])?(-D|-d|--delete)([[:space:]]|$)" && return 0
+  printf '%s' "$1" | grep -qE "${CMD_START}git[[:space:]]+(gc|prune|reflog[[:space:]]+expire|update-ref[[:space:]]+-d|filter-branch)([[:space:]]|$)" && return 0
+  printf '%s' "$1" | grep -qE "${CMD_START}git[[:space:]]+worktree[[:space:]]+(remove|prune)([[:space:]]|$)" && return 0
   return 1
 }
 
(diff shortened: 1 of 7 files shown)


.claude/hooks/auto-allow-worktree-destructive.sh
```diff
@@ -25,15 +56,110 @@ if [ -z "$command" ]; then
...
-  *"git reset --hard"*) ;;
-  *"git rebase"*) ;;
-  *"git merge"*) ;;
-  *"git push --force"*|*"git push -f "*|*"git push -f")  ;;
-  *) echo '{"continue": true}'; exit 0 ;;
-esac
+# Anchored at command-word position: start of string, or after a shell
+# separator (; & | && || newline). A bare substring match would treat
+# `grep "sudo " docs/` as running sudo and reintroduce the prompt.
+CMD_START='(^|[;&|(]|&&|\|\||\n)[[:space:]]*'
```
bess-agent: **High — anchored guards are all bypassable by a command prefix.**

`CMD_START` anchors on start-of-string or a shell separator, so an env-var assignment prefix (or `nohup`/`time`/`command`/`env`) shifts the command word and defeats *every* anchored guard in this file. Verified from inside a worktree against the real hook:

- `PODMAN=1 podman machine rm` → **allow** — a settings.json `deny` silently upgraded to an unprompted allow, destroying the shared podman VM every worktree's E2E stack uses
- `GIT_SSH_COMMAND=ssh git push origin main` → **allow** — shared-ref push guard bypassed
- `nohup gh pr merge 561` → **allow**

Env prefixes are ordinary generated-command shape (`PYTHONPATH=. …`, `TZ=UTC …`), so this is an accident-class miss, not only an adversarial one.

Fix: after `CMD_START`, consume an optional `([A-Za-z_][A-Za-z0-9_]*=[^[:space:]]*[[:space:]]+)*` prefix, plus `nohup|time|command|env`.

.claude/hooks/auto-allow-worktree-destructive.sh
```diff
@@ -25,15 +56,110 @@ if [ -z "$command" ]; then
...
+  # not just as a bare word, since the bare word is the spelling least
+  # likely to appear when something force-updates a ref.
+  #
+  # `--force-with-lease` is exempt: it is the one force form that REFUSES to
+  # clobber an update it has not seen, which is exactly the accident a
+  # prompt here would be guarding against. It is also routine after the
+  # Step 4 merge, so asking costs a stall on every run. Bare `--force`/`-f`
+  # keeps asking -- that one overwrites unconditionally. Order matters
+  # below: test the lease form first, or `--force` matches its prefix.
+  if printf '%s' "$1" | grep -qE "${CMD_START}git[[:space:]]+push([[:space:]]|$)"; then
```
bess-agent: **Medium — global git options bypass the shared-ref push guard.**

This only matches a literal `git`+space+`push`, so anything between the two words sails through. Verified:

- `git -C sub push origin main` → **allow**
- `git --git-dir=sub/.git push origin main` → **allow**
- `git -c push.default=matching push` → **allow**

`needs_scrutiny` catches `-C` only when the path is absolute or contains `..`, so a relative `-C` isn't caught either.

.claude/hooks/auto-allow-worktree-destructive.sh
```diff
@@ -25,15 +56,110 @@ if [ -z "$command" ]; then
...
+    if ! printf '%s' "$1" | grep -qE '\-\-force-with-lease'; then
+      printf '%s' "$1" | grep -qE "(--force|[[:space:]]-f([[:space:]]|$))" && return 0
+    fi
+    printf '%s' "$1" | grep -qE "(--all|--mirror|--tags)" && return 0
+    printf '%s' "$1" | grep -qE "(^|[[:space:]:+])(refs/heads/)?(main|beta)([[:space:]]|:|$)" && return 0
```
bess-agent: **Medium — the refspec match scans the whole command string, not the push arguments.**

Any `main`/`beta` word anywhere in a compound turns a safe push into a prompt. Verified:

- `git push -u origin feat/x && gh pr create --draft --base main --title x --body-file /tmp/b.md` → **ask**
- `git commit -m "fix main crash" && git push origin feat/x` → **ask**

The first one is literally `implement-issue`'s closing step — i.e. the exact stall this PR exists to remove. Scope the refspec scan to the `git push` segment rather than `$1` as a whole.

.claude/hooks/auto-allow-worktree-destructive.sh
```diff
@@ -25,15 +56,110 @@ if [ -z "$command" ]; then
...
+  # merged worktrees in a loop, so asking turns one cleanup step into dozens
+  # of prompts -- and the danger it would be guarding is already handled
+  # upstream: git itself REFUSES to delete a branch checked out in another
+  # worktree ("error: cannot delete branch 'x' used by worktree at ...",
+  # exit 1), so it cannot cut another agent out from under itself. What is
+  # left is an unchecked-out branch, whose SHA git prints on delete and
+  # which stays reachable -- protected because `reflog expire`, `gc` and
+  # `filter-branch` below keep asking. `git tag -d` DOES stay: a tag is how
+  # a release is addressed, and nothing refuses to delete one.
+  printf '%s' "$1" | grep -qE "${CMD_START}git[[:space:]]+tag[[:space:]]+(.*[[:space:]])?(-d|--delete)([[:space:]]|$)" && return 0
```
bess-agent, line 186: **Medium — `mutates_shared_state` misses the stash, and two more shared-config writes.**

The stash is a single shared `refs/stash` across all worktrees, but it isn't guarded. Verified from inside a worktree:

- `git stash clear` → **allow**
- `git stash drop` → **allow**

Both wipe stashes created in the main checkout. Same class as the `reflog expire`/`gc` cases this function does guard, and the project convention leans on `git stash` for branch isolation.

Also missed (lower severity, same function):

- `git config --global user.email …` → **allow** (writes `~/.gitconfig`, outside every worktree)
- `git remote set-url origin …` / `git remote remove origin` → **allow** — a linked worktree shares `.git/config` with the main checkout, so this repoints or removes the shared remote for every checkout at once

.claude/hooks/auto-allow-worktree-destructive.sh
```diff
@@ -45,19 +171,157 @@ fi
...
+# same enumeration bug as before, only inverted: on this side a miss is a
+# silent auto-allow, not an extra prompt, and the misses were the commonest
+# in-place writers there are -- `sed -i` into the main checkout, `touch`,
+# `mkdir -p`, `python3 -c "shutil.rmtree(...)"`. Trigger instead on the
+# thing that actually makes a command dangerous here: naming a path that
+# could sit outside the worktree. escapes_worktree then decides.
+#
+# The verb list survives only for the indirection case -- `rm -rf $HOME/x`
+# names no literal path but still resolves to one.
+needs_scrutiny() {
```
bess-agent, line 280: **Medium — read-only inspection outside the worktree now prompts.**

`needs_scrutiny` fires on any absolute path or `~` regardless of whether the command can write, and `escapes_worktree` then asks for anything outside the worktree. Verified:

- `cat /Users/johanzander/Downloads/bess-debug.json` → **ask**
- `head -50 /Users/…/bess-manager/CHANGELOG.md` → **ask**
- `ls ~/Downloads` → **ask**
- `grep -rn foo /usr/local/lib` → **ask**

Reading a user's debug bundle from `~/Downloads` is a core workflow here (`visualize-debug-log`), so this trades one stall class for another. Gating on write-capable commands would keep the guard while letting reads through.

.claude/hooks/auto-allow-worktree-destructive.sh
```diff
@@ -25,15 +56,110 @@ if [ -z "$command" ]; then
...
+  # the closing step of implement-issue, so a run that no longer stalls on
+  # `rm -f` would instead stall at the finish line, and `gh issue comment`
+  # would stall it in the middle. Authoring a PR/issue on this repo IS the
+  # work product. What stays behind a prompt is everything that publishes,
+  # merges, or reconfigures: pr merge, release, repo edit/delete, secret,
+  # workflow run, and any non-GET `gh api`.
+  local gh_all gh_safe
+  # `grep -c` counts LINES, and a command is normally one line, so both
+  # counts would be 1 for that compound. Count occurrences with -o | wc -l.
+  gh_all=$(printf '%s' "$1" | grep -oE "${CMD_START}gh[[:space:]]" | wc -l | tr -d ' ' || true)
```
bess-agent: **Low — the `gh` safe-list omits read-only/routine forms that previously passed.**

Now prompting where settings.json previously allowed or ignored them: `gh --version`, `gh pr checkout 561`, `gh repo clone o/r`, and `gh api graphql -f query=…` (read-only GraphQL is indistinguishable from a mutation here).

Conversely, bare `gh` with no trailing space matches neither counter and is allowed.

.claude/hooks/tests/auto-allow-worktree-destructive.test.sh
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/env bash
```
bess-agent, line 1: **Low — nothing invokes this test file.**

`grep -rn "hooks/tests\|auto-allow-worktree" scripts/ .github/` returns no hits. A 186-line expectation matrix guarding a security-adjacent decision function, with no runner wired into the quality gate or CI, will silently rot the first time these regexes change — and given findings 1–5, the matrix is exactly what should have caught them.

.claude/hooks/tests/auto-allow-worktree-destructive.test.sh
```diff
@@ -0,0 +1,186 @@
...
+
+HOOK=$(cd "$(dirname "$0")/.." && pwd)/auto-allow-worktree-destructive.sh
+[ -f "$HOOK" ] || { echo "hook not found: $HOOK" >&2; exit 1; }
+
+# Deliberately NOT under mktemp's default root: the hook counts /tmp and
+# the macOS $TMPDIR as disposable, so a fixture there would be auto-allowed
+# and the "reaches the main checkout" cases could not be expressed at all.
+# `pwd -P` because `git rev-parse --show-toplevel` reports a resolved path
+# and the hook compares the two as strings.
+TMP=$(mktemp -d "${HOME}/.cache/bess-hook-test-XXXXXX") || exit 1
```
bess-agent, line 28: **Low — `mktemp` under `~/.cache` fails where that directory doesn't exist.**

On a fresh macOS install or most CI runners `~/.cache` is absent, so this exits 1 with only mktemp's stderr — indistinguishable from a real test failure. `mkdir -p "$HOME/.cache"` first, or use a fixture root the hook doesn't treat as disposable via a `TMPDIR` override.

.claude/settings.json
```diff
@@ -79,6 +79,27 @@
         ]
       }
     ],
+    "WorktreeCreate": [
```
bess-agent, line 82: **Low (design) — the two halves of this PR partly cancel.**

The link hook exists so `settings.local.json` rules apply inside worktrees, but a hook `allow` supersedes settings rules — so any Bash `deny`/`ask` the user sets locally is overridden inside a worktree by the blanket allow. Only `defaultMode` and non-Bash rules actually survive the trip.

Worth stating in CLAUDE.md so a future local Bash `deny` isn't assumed effective inside a worktree.

bess-agent (commented):
Code review of the worktree permission hooks. I extracted the hook and exercised it against a throwaway repo plus a linked worktree (~70 probe commands), so every claim below has a verified command → decision behind it.

**Blocker: finding 1** — an env-var prefix converts an explicit settings.json `deny` into a silent, unprompted `allow`.

Other findings are inline. Confirmed sound while probing: main-checkout fall-through (including through a symlinked path — `rev-parse --show-toplevel` and `worktree list` both report the physical path), `..`-through-`/tmp` traversal, quoted absolute paths inside `python3 -c`, compound `gh pr view && gh pr merge`, `git worktree remove` lone-vs-chained, `--force-with-lease` handling, the `awk` `$2`-vs-`sub()` path-with-spaces fix in both scripts, and `set -f` restoration around the word-split loop.

PR #564, opened 2026-08-13
fix: surface a missing consumption sensor instead of stalling on "Initializing"

johanzander:
## Summary

- A missing consumption sensor now reports a critical health error instead of an endless "Initializing" dashboard
- The `sensor` strategy can no longer be selected without the sensor it requires, and is no longer the default
- No consumption fallback added — the failure stays explicit, it just becomes visible

## Root cause

The reporter's system (`solax_modbus_growatt_min`, v10.0.2) had `consumption_strategy: "sensor"` with `48h_avg_grid_import` **"Not configured"**. Every optimization run therefore aborted:

```
ERROR | core.bess.battery_system_manager:740 - Failed to update battery schedule:
       Configuration error in 48h_avg_grid_import sensor not available
```

15 occurrences between 00:00:39 and 00:25:00 in the bundle; **Total Schedules: 0**. `backend/api.py:746-755` returns `{"error": "initializing"}` for as long as `schedule_store` is empty, so `DashboardPage.tsx:156` spun forever with no error code. Demo mode changed nothing — `set_demo_mode` only flips write suppression, reads still hit HA, which is why the reporter saw it in both modes.

**Why nothing warned.** `check_prediction_health` already added `get_estimated_consumption` to the checked methods under this strategy (`sensor_collector.py:850`), but the component was declared `is_required=False`, so `required_methods` was empty (`health_check.py:162`), an unmapped sensor became `SKIPPED` (`health_check.py:189-201`), the component reported OK, and `_critical_sensor_failures` stayed empty (`battery_system_manager.py:3163`). The bundle reports **"Overall Status: OK, Critical Issues: 0"** on a system that could not optimize at all, while `refresh_health_check` kept retrying the schedule build believing every sensor was healthy.

**Why the reporter hit it.** `sensor` was the default in `settings.py:216`, `settings_store.py:456` and `SetupWizardPage.tsx:73`, while `HomeFormSection.tsx` disabled the `influxdb_7d_avg` and `ha_statistics` options when their backing sensor was missing — but not the default one. Finishing the wizard without hand-building a 48h template helper produced a system that could never build a schedule.

## Fix

| area | change |
|---|---|
| `health_check.py` | `perform_health_check` gains an optional `required_methods` subset, so a component can require one method without promoting its optional siblings. Defaults to `all_methods`, so every existing caller is unchanged |
| `sensor_collector.py` | `check_prediction_health` passes `is_required=(strategy == "sensor")` and names only `get_estimated_consumption` as required — the solar forecast stays optional under every strategy |
| `HomeFormSection.tsx` | the "Home Assistant sensor" option gets the guard the other two already had, plus an amber note naming the sensor |
| `settings.py`, `settings_store.py`, `SetupWizardPage.tsx`, `SettingsPage.tsx` | default `sensor` → `fixed`, the one strategy that needs no sensor and cannot stall |
| `USER_GUIDE.md` | the "Strategy 1: `sensor` (default)" heading was invalidated by this change — now labelled legacy, with the requirement stated |

**Existing installs are unaffected**: persisted settings are applied with `.update(**settings["home"])`, which only sets keys present in the file, so a stored `consumption_strategy` always wins over the new default.

**Deliberately not changed:** no fallback consumption value for the `sensor` strategy. Per `rules.md` the fix is explicit failure plus preventing the invalid configuration, not silently substituting a number the user did not ask for. `ha_statistics` keeps its documented fixed-profile fallback because that is a *temporary data-accumulation* state, not a misconfiguration.

**Scope assessment:** local, within the existing owners. No parameter, flag, extra trigger or second construction site was added to route around an ordering/timing/dependency problem — `required_methods` expresses a domain fact (which method is load-bearing for the active strategy) inside the function that already computes exactly that list.

## Test plan

- [x] Fast suite: 1762 passed, 29 skipped
- [x] Slow suite: 534 passed, 4 skipped
- [x] `black --check` / `ruff` clean; `tsc --noEmit` clean; eslint 0 errors; vitest 124 passed
- [x] **Live stack** (`docker-compose.ci.yml` + mock-HA, real HTTP against the real backend), reporter-shaped settings — `consumption_strategy: "sensor"`, `48h_avg_grid_import` stripped:
  - `GET /api/system-health` → `Energy Prediction | required=True | ERROR`, detail `Average Hourly Power Consumption | ERROR | Method call failed: Configuration error in 48h_avg_grid_import sensor not available` — the reporter's exact string
  - `GET /api/dashboard-health-summary` → `hasCriticalErrors: true`, `systemMode: degraded`, `Energy Prediction` listed in `criticalIssues`
  - **Control run**, same stack with the stock fixture (`consumption_strategy: "fixed"`) → `Energy Prediction | required=False | OK`, absent from `criticalIssues`. The two runs differ only in the strategy, so the escalation is keyed to the strategy and not blanket-applied
  - Served JS bundle contains the new guard text (`48h Avg Grid Import`, `Without it no schedule can be built at all.`)
- [ ] **The new Playwright spec was NOT executed locally.** The local Playwright browser install is still the corrupted one from #556 (`npx playwright install chromium` did not complete). The spec parses and registers (`[chromium] › settings-page.spec.ts:46:7`), but its assertions are unverified until CI runs them.

Written RED first, then mutated. Tree confirmed clean afterwards.

- Before the fix existed: **4 failed, 23 passed** — including the log line the reporter saw, `✓ [OPTIONAL] Energy Prediction: OK`
- Reverted: `is_required=sensor_strategy_active` → `is_required=False` in `sensor_collector.py`
- Result: **4 failed, 23 passed** — `test_error_when_estimated_consumption_unavailable`, `test_error_when_estimated_consumption_not_configured`, `test_component_is_required_under_sensor_strategy`, `test_unconfigured_sensor_becomes_a_critical_failure`
- Restored: 32 passed, tree clean

The guard tests (`test_solar_forecast_failure_is_only_a_warning_under_sensor_strategy`, the `fixed`/`ha_statistics` non-required tests) pass both before and after, so they pin that the change does not over-require. One pre-existing test, `test_extended_horizon.py:557`, genuinely depended on the old default — it mocks `controller.consumption_forecast`, which only the `sensor` strategy reads — and now sets the strategy explicitly rather than having its isolation silently ignored.

## Outcome-level coverage

- `TestCriticalFailurePromotion` asserts the end state the reporter experienced: the unmapped sensor reaching `get_critical_sensor_failures()`, which is what the dashboard renders as a degraded banner. It builds the component from the real `check_prediction_health` output and runs it through the real promotion path in `_run_health_check`, rather than asserting an internal flag.
- Live HTTP evidence above covers the API surface the frontend actually consumes (`/api/system-health`, `/api/dashboard-health-summary`), with a control run isolating the strategy as the cause.
- No DP/intent/control-mapping behaviour changes, so no golden re-pin and no `R == P` scenario is implicated; no fixture was added, so the selector-golden and VPP-baseline meta-tests are untouched.

Closes #558

### DIFF
--- a/core/bess/health_check.py
+++ b/core/bess/health_check.py
@@ -141,6 +141,7 @@ def perform_health_check(
     is_required: bool,
     controller,
     all_methods: list[str],
+    required_methods: list[str] | None = None,
 ) -> dict:
     """Generic health check function that can be used by any component.
 
@@ -156,11 +157,22 @@ def perform_health_check(
             optional components show WARNING.
         controller: The controller instance with validate_methods_sensors method
         all_methods: List of all method names this component uses
+        required_methods: Subset of ``all_methods`` that is required, for
+            components where only some methods are load-bearing. Only
+            meaningful with ``is_required=True``; defaults to all of them.
+            Energy Prediction uses this: under the ``sensor`` consumption
+            strategy the consumption sensor is mandatory while the solar
+            forecast stays genuinely optional (#558).
 
     Returns:
         Health check result dictionary
     """
-    required_methods = all_methods if is_required else []
+    if is_required:
+        required_methods = (
+            all_methods if required_methods is None else list(required_methods)
+        )
+    else:
+        required_methods = []
     health_check = {
         "name": component_name,
         "description": description,
--- a/core/bess/health_check.py
+++ b/core/bess/health_check.py
@@ -171,6 +171,16 @@ def perform_health_check(
         required_methods = (
             all_methods if required_methods is None else list(required_methods)
         )
+        # A required method that is not also checked is never returned by
+        # validate_methods_sensors, so required_total stays 0 and the
+        # "specified but none configured" branch below reports ERROR for
+        # something that was never looked at.
+        unchecked = set(required_methods) - set(all_methods)
+        if unchecked:
+            raise ValueError(
+                f"required_methods {sorted(unchecked)} are not in all_methods "
+                f"for component '{component_name}'"
+            )
     else:
         required_methods = []
     health_check = {
--- a/core/bess/sensor_collector.py
+++ b/core/bess/sensor_collector.py
@@ -845,17 +845,27 @@ class SensorCollector:
         Only validates the ``get_estimated_consumption`` sensor when the
         ``sensor`` strategy is active — other strategies (fixed,
         influxdb_7d_avg, ha_statistics) do not rely on that HA sensor.
+
+        Under the ``sensor`` strategy that sensor is *required*: every
+        optimization run reads it and aborts without it, so no schedule can
+        ever be built and the dashboard sits on "Initializing" indefinitely
+        (#558). The solar forecast stays optional under every strategy, so
+        only the consumption method is named as required.
         """
         all_methods = ["get_solar_forecast"]
-        if consumption_strategy == "sensor":
+        sensor_strategy_active = consumption_strategy == "sensor"
+        if sensor_strategy_active:
             all_methods = ["get_estimated_consumption", *all_methods]
 
         return perform_health_check(
             component_name="Energy Prediction",
             description="Solar and consumption forecasting for optimization",
-            is_required=False,
+            is_required=sensor_strategy_active,
             controller=self.ha_controller,
             all_methods=all_methods,
+            required_methods=(
+                ["get_estimated_consumption"] if sensor_strategy_active else None
+            ),
         )
 
     def check_health(self, consumption_strategy: str = "sensor") -> list:
--- a/frontend/src/components/settings/HomeFormSection.tsx
+++ b/frontend/src/components/settings/HomeFormSection.tsx
@@ -19,6 +19,7 @@ interface Props {
 
 export function HomeFormSection({ form, onChange, sensors }: Props) {
   const haStatsSensorConfigured = Boolean(sensors?.['lifetime_load_consumption']);
+  const avgGridImportSensorConfigured = Boolean(sensors?.['48h_avg_grid_import']);
   const localLoadSensorConfigured = Boolean(sensors?.['local_load_power']);
   const chargeRateSensorConfigured = Boolean(sensors?.['battery_charging_power_rate']);
   const currentSensorsConfigured = form.phaseCount === 1
@@ -34,7 +35,7 @@ export function HomeFormSection({ form, onChange, sensors }: Props) {
           'Data source',
           [
             { value: 'fixed', label: 'Fixed value' },
-            { value: 'sensor', label: 'Home Assistant sensor' },
+            { value: 'sensor', label: 'Home Assistant sensor', disabled: !avgGridImportSensorConfigured },
             { value: 'influxdb_7d_avg', label: 'InfluxDB (requires InfluxDB integration)', disabled: !localLoadSensorConfigured },
             { value: 'ha_statistics', label: 'HA Statistics (7-day hourly profile)', disabled: !haStatsSensorConfigured },
           ],
@@ -45,6 +46,13 @@ export function HomeFormSection({ form, onChange, sensors }: Props) {
           <p className="text-xs text-amber-600 dark:text-amber-400 pt-1">
             InfluxDB requires the <strong>Local Load Power</strong> sensor to be configured in the{' '}
             <strong>Sensors</strong> tab. This sensor is not available on all inverter platforms (e.g. Growatt SPH).
+          </p>
+        )}
+        {!avgGridImportSensorConfigured && (
+          <p className="text-xs text-amber-600 dark:text-amber-400 pt-1">
+            Home Assistant sensor requires the <strong>48h Avg Grid Import</strong> sensor to be
+            configured in the <strong>Sensors</strong> tab under Consumption Forecast. Without it
+            no schedule can be built at all.
           </p>
         )}
         {!haStatsSensorConfigured && (
--- a/frontend/src/components/settings/HomeFormSection.tsx
+++ b/frontend/src/components/settings/HomeFormSection.tsx
@@ -51,8 +51,7 @@ export function HomeFormSection({ form, onChange, sensors }: Props) {
         {!avgGridImportSensorConfigured && (
           <p className="text-xs text-amber-600 dark:text-amber-400 pt-1">
             Home Assistant sensor requires the <strong>48h Avg Grid Import</strong> sensor to be
-            configured in the <strong>Sensors</strong> tab under Consumption Forecast. Without it
-            no schedule can be built at all.
+            configured in the <strong>Sensors</strong> tab under Consumption Forecast.
           </p>
         )}
         {!haStatsSensorConfigured && (
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
+- **A missing consumption sensor no longer leaves the dashboard stuck on "Initializing"** — with the `sensor` strategy selected and no `48h_avg_grid_import` entity, the health check now reports a critical error, that strategy can't be selected without its sensor, and new installs default to `fixed`. ([#558](https://github.com/johanzander/bess-manager/issues/558))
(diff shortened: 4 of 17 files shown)


core/bess/sensor_collector.py
```diff
@@ -845,17 +845,27 @@ def check_prediction_health(self, consumption_strategy: str = "sensor") -> dict:
...
-        if consumption_strategy == "sensor":
+        sensor_strategy_active = consumption_strategy == "sensor"
+        if sensor_strategy_active:
             all_methods = ["get_estimated_consumption", *all_methods]
 
         return perform_health_check(
             component_name="Energy Prediction",
             description="Solar and consumption forecasting for optimization",
-            is_required=False,
+            is_required=sensor_strategy_active,
```
bess-agent, line 863: **Medium.** Setting `is_required=True` for the whole "Energy Prediction" component contradicts this PR's own goal of keeping the solar forecast optional, because `PreflightCheckDialog.toRowStatus` (`frontend/src/components/PreflightCheckDialog.tsx:19-24`) keys off the *component's* `required` flag, not the per-method one: any non-`OK` status on a required component maps to `error`, which sets `allPassed=false` and disables the "Enable Live Control" button.

Scenario: a user on the `sensor` strategy with a working 48h sensor but a Solcast entity returning NaN/None gets component status `WARNING` (correct, and pinned by the new `test_solar_forecast_failure_is_only_a_warning_under_sensor_strategy`) together with `required: true` — so the preflight dialog renders a red X and hard-blocks leaving demo mode. Before this PR that same state was a non-blocking amber warning.

Suggested fix: have `toRowStatus` block only on `status === 'ERROR'`, or expose the required-subset result rather than a component-wide flag.

bess-agent, line 863: **Low.** The same flag makes *transient* unavailability critical. `perform_health_check` deliberately distinguishes `entity_unavailable` (-> check-level `WARNING`, "temporarily unavailable") from truly broken, but `determine_health_status` only counts `status == "OK"` toward `required_working`. So during an HA restart, the 48h sensor being momentarily unavailable now yields component `ERROR` -> `_run_health_check` logs "SYSTEM DEGRADED", `/api/dashboard-health-summary` shows the critical banner, and `refresh_health_check` skips the initial-schedule retry.

This is consistent with how every other required component already behaves, so it may well be acceptable — flagging it because it is a new user-visible behaviour for *every* `sensor`-strategy install, not just misconfigured ones.

frontend/src/components/settings/HomeFormSection.tsx
```diff
@@ -47,6 +48,13 @@ export function HomeFormSection({ form, onChange, sensors }: Props) {
             <strong>Sensors</strong> tab. This sensor is not available on all inverter platforms (e.g. Growatt SPH).
           </p>
         )}
+        {!avgGridImportSensorConfigured && (
```
bess-agent, line 51: **Low.** This amber note renders whenever `48h_avg_grid_import` is unset, regardless of the selected strategy. On a fresh install — which now defaults to `fixed` — the Home tab shows "Without it no schedule can be built at all." about a strategy the user is not using and cannot select. Alarming, and factually wrong for their configuration.

The two sibling notes have the same unconditional shape, but their wording is neutral; this one is not. Suggested fix: gate it on `form.consumptionStrategy === 'sensor'`, or soften the wording.

core/bess/health_check.py
```diff
@@ -141,6 +141,7 @@ def perform_health_check(
     is_required: bool,
     controller,
     all_methods: list[str],
+    required_methods: list[str] | None = None,
```
bess-agent, line 144: **Low.** `required_methods` is not validated as a subset of `all_methods`. A future caller naming a method absent from `all_methods` gets a silently ineffective requirement: `validate_methods_sensors` never returns it, `required_total` stays 0, and the `required_methods and required_total == 0` branch then reports ERROR for a method that was never actually checked.

The current caller is correct; an `assert set(required_methods) <= set(all_methods)` here would keep it that way.

bess-agent (commented):
Code review of #564 — the core change looks sound. `perform_health_check` gains an optional `required_methods` subset, every existing caller passes keywords and defaults to the old behaviour, `determine_health_status` already handles a required-but-unmapped sensor (`not_configured` is no longer skipped, so `required_working < required_total` -> ERROR), the `48h_avg_grid_import` key the frontend guard reads does resolve through `getActiveSensorsFlat` (it lives in `sensors.shared`), the e2e fixture matches what the new Playwright spec asserts, and the changed defaults only touch `_bootstrap_defaults` (first boot) plus dataclass/form defaults, so persisted `sensor` installs really are unaffected.

Five findings, four of them inline below. The fifth has no anchor in this diff:

**`backend/api.py:3159` — Low.** The "can no longer be selected without the sensor it requires" guarantee is client-side only. `PATCH /api/settings` passes `payload.consumptionStrategy` straight into `HomeSettings.update()` with no validation — not even against the four known values (an unknown string only blows up later at `battery_system_manager.py:1224`). A user editing via the API, or any path that writes `bess_settings.json` directly, can still land in exactly the reported stall state. The new health error does surface it, so this is defence-in-depth rather than a regression.

One observation, not a finding: the wizard already auto-enables fuse protection when its sensors are discovered (`SetupWizardPage.tsx:226`), so there is precedent for auto-selecting `ha_statistics` when `lifetime_load_consumption` is found — otherwise every new install now silently ships with a flat 3.5 kWh/h forecast, which is what the USER_GUIDE calls the least-good option.

core/bess/sensor_collector.py
bess-agent: Confirmed and fixed in 10b80e5a. Verified the mechanism: `toRowStatus("WARNING", required=true)` returns `error`, which sets `allPassed=false` and disables the button — a real regression from making Energy Prediction required.

Took the first suggested option, blocking only on `ERROR`, because it is precise rather than merely looser: `determine_health_status` returns ERROR whenever `required_working < required_total`, so a required component reporting WARNING can only mean an *optional* member failed. A required sensor that is temporarily unavailable still yields component ERROR (the check-level WARNING does not count toward `required_working`), so the gate does not weaken for the case it exists to catch.

Optional components keep their existing mapping — my first attempt checked `ERROR` before the `required === false` branch, and the existing suite caught it immediately: `does not show an optional component in ERROR as passing` and `still enables the button when only an optional component has ERROR status` both failed. Final shape is OK → ok; optional → NOT_CONFIGURED ? ok : warning; required → ERROR ? error : warning. New test `enables the button when a required component has WARNING status` pins the fixed case; 8/8 in that file pass.

bess-agent: Accurate description, deliberately not changed — reasoning recorded here rather than silently.

A required sensor going `entity_unavailable` does now produce component ERROR, the "SYSTEM DEGRADED" log, and the critical banner. That is identical to how Battery Monitoring, Energy Monitoring and Battery Control already behave for their sensors, so the alternative would be a special case for this one component.

On the skipped retry specifically: `refresh_health_check` only retries when no schedule exists yet, and in that state the build reads the same unavailable sensor and raises `SystemConfigurationError` anyway. Skipping it loses nothing; the retry fires on the next health check once the sensor returns.

The visible cost is a banner during an HA restart for `sensor`-strategy installs. Given the default moved to `fixed` and the UI now blocks selecting `sensor` without its sensor, that population is small and explicitly opted in — and a banner while the forecast source is down is true, not noise.

frontend/src/components/settings/HomeFormSection.tsx
bess-agent: Fixed in 10b80e5a, by softening the wording rather than gating on the strategy.

Gating on `form.consumptionStrategy === "sensor"` would hide the note exactly when it is needed: the option is disabled, so a user who wants it can never select it, and the note is the only thing explaining why. Kept it unconditional and matching its two neutral siblings — it now stops at "configured in the Sensors tab under Consumption Forecast", dropping the "Without it no schedule can be built at all" sentence that was alarming and, on a fresh `fixed` install, describing a strategy the user is not using.

core/bess/health_check.py
bess-agent: Fixed in 10b80e5a. Raises `ValueError` rather than `assert` — asserts are stripped under `python -O`, and this is a contract violation worth failing loudly on:

```python
unchecked = set(required_methods) - set(all_methods)
if unchecked:
    raise ValueError(
        f"required_methods {sorted(unchecked)} are not in all_methods "
        f"for component {component_name!r}"
    )
```

Placed inside the `is_required` branch, since `required_methods` is ignored entirely when the component is optional.

PR #568, opened 2026-08-13
fix: defer Growatt MIN TOU writes until a segment is about to take effect

johanzander:
## Summary

- A TOU segment is written only once its start is within `WRITE_HORIZON_MINUTES` (45). A segment hours away is no longer rewritten every time the plan shifts around a marginal period.
- A segment that is already *running* is no longer rewritten every 15 minutes — its start time was being truncated forward each cycle.
- Measured end to end: **26 → 13** hardware writes over 96 real optimization cycles with four separated windows. The issue's 13:45 window takes **1 write where production did 11**.

## Root cause

`sync_to_hardware` had no notion of *when* a segment starts. It wrote any segment whose content differed from the fresh hardware read, filtered only by `end_minute(segment) >= effective_minute` — "hasn't ended yet" — with no upper bound. So a marginal boundary flip five hours out failed the content match and was written, every cycle.

The reporter's worked example: the 13:45 `battery_first` window rewritten 11 times between 08:16 and 13:45, ending on the value written at 08:16.

## Fix

`_select_hardware_intervals` already owned *"which intervals reach hardware now, rest deferred to a later cycle"* for the 9-slot cascade. The write horizon is that same idea keyed on time, so it goes there rather than into a new mechanism. Already-active intervals have `start <= now`, so they can never be deferred, and a restart pushes the imminent plan immediately as an emergent property.

Three things had to follow, because "eligible to write" and "in the plan" stop being the same set once writes are deferred:

1. **`evaluate_intents`** decides whether `sync_to_hardware` runs at all, and compared only the whole daily plan — while already computing the hardware-eligible set and discarding it (`candidate_intervals, _active = ...`). Without this, a segment deferred at 08:15 is never written once the plan stabilises. This also fixes the same latent gap for the 9-slot cascade, where a pending interval could never be promoted on a stable plan.
2. **The disable side** must diff against the full plan, not the eligible subset — otherwise a correctly-programmed future segment reads as *unplanned* and gets cleared, then rewritten when it comes back into range. Deferral applies to updates only; disables stay prompt, so #551's immediate orphan cleanup is preserved exactly.
3. **Slot assignment** must reserve slots for planned-but-deferred segments, or the next imminent write lands on top of one and destroys it with no disable ever issued. A documented reclaim policy (furthest-out segment yields first) keeps a table left full by the old always-write behaviour from deadlocking after an upgrade.

Separately, `_group_periods_by_mode` now reports the group covering the current period from its *true* start. The plan is rebuilt from the current period each cycle, which truncated an in-progress segment's `start_time` forward to "now", renaming it every 15 minutes — 16 writes for a stable two-hour window.

`pending_write` is narrowed to mean "due but missing" rather than "not on hardware", so the dashboard does not paint most of the day amber `Pending Write`. Backend-only; the frontend is untouched.

## Scope assessment

**Local.** Everything is inside `GrowattMinController`, within the target methods' existing contracts. No new class, no new owner. The diff adds no parameter, flag, or extra trigger whose job is to route around a problem — `_assign_hardware_slots` gains `planned_tou` because slot reservation genuinely needs to know what the plan still wants, which is the same information the disable side now uses.

**Consequence worth noting:** at a 45-minute horizon at most `floor(45/15) + 1 = 4` intervals are ever eligible (verified against a worst case alternating mode every 15 minutes all day), so the 9-slot cap can no longer bind end to end. The cap is kept and still enforced as the inverter's hard contract; its guard is now exercised directly on `_select_hardware_intervals` rather than through a plan that can no longer reach it.

## Test plan

- [x] `pytest -m "not slow"` — 1824 passed, on the tree merged with `origin/main`
- [x] `pytest -m slow` — 534 passed, same tree
- [x] black / ruff clean; mypy byte-identical to the `origin/main` baseline for this file (19 pre-existing errors, none added)
- [x] **Observed through the real HA API layer** against the mock-HA container (`docker-compose.ci.yml`), driving the issue's own 13:45 window across four cycles with `HomeAssistantAPIController` → `growatt_server/update_time_segment`:

  | Cycle | Real HA service calls |
  |---|---|
  | 08:15 (5.5h before) | none — deferred |
  | 11:00 (2.75h before) | none — deferred |
  | 12:30 (1.25h before) | none — deferred |
  | 13:15 (30 min before) | 1 × `update_time_segment` |

- [x] **Full-day in-process run** (96 real optimization cycles, real DP → `BatterySystemManager` → controller), four separated windows: **26 writes before, 13 after**

Ten mutations, each reverting one behaviour; every one reddens its intended test. Source verified restored intact after each.

| Mutation | Test(s) that failed |
|---|---|
| `WRITE_HORIZON_MINUTES` → 24h (deferral off) | far-future-flip, gate-fires, deferred-becomes-eligible |
| `evaluate_intents` eligible-set check → `if False and …` | gate-fires-when-imminent |
| Disable side diffs the eligible subset again | far-future-left-alone, keeps-its-slot, reclaim |
| Clear-all branch keyed on eligible subset again | far-future-left-alone |
| Never disable anything | #551 unplanned-segments-are-disabled |
| Remove `[: self.max_intervals]` | slot-cap-still-applies |
| Slot protection back to eligible-only | keeps-its-hardware-slot, reclaim |
| Reclaim nearest slot instead of furthest | reclaimed-furthest-first |
| Deferred segments flagged pending again | not-flagged-pending-write |
| In-progress start clamped to "now" again | running-window-not-rewritten, restart-still-programs |

Two of these caught genuinely vacuous tests during development rather than after: the reclaim test initially used nine *contiguous* hours, which merge into one interval, so the reclaim path never fired and the mutation survived.

## Outcome-level coverage

Asserted on **the inverter's segment table** — what is programmed, and when — via the `_SimulatingController` stub that models the 9 hardware slots, not on command arguments.

A plan-faithfulness (`R == P`) scenario is deliberately **not** the primary test here: the diff touches only the TOU hardware-write layer, and deferral is designed to have exactly zero effect on battery behaviour, so `R == P` is unchanged by construction and would prove nothing. `inverter_simulator` has no notion of segment writes. The correct outcome for this layer is "what the segment table holds at the moment each window becomes active", which is what these tests pin.

## Notes for review

Two things found during review and deliberately left alone, both pre-existing and out of scope:

- The remaining writes are dominated by a window's *end* extending as the plan re-solves each cycle — that is the near-tie plan flipping of #485, which this issue explicitly composes with rather than subsumes.
- The three retry attempts inside the horizon are serially correlated (they share failure modes that outlast 45 minutes). The exposure is narrow: only periods needing a non-default mode are at risk, and the fallback is `load_first` — normal self-consumption, a lost optimization opportunity rather than unsafe behaviour.

`docs/SOFTWARE_DESIGN.md`'s "Schedule generation" list is updated — it documented this exact mechanism. `bess-knowledge.md` mentions nothing this diff touches.

Closes #554


johanzander (commented):
Code review of the real PR scope (`merge-base(origin/main, head)...head`). The core deferral mechanics hold up: `new_tou ⊆ planned_tou` under the same `effective_period`, the reclaim path is self-healing, the `len(planned_tou) == 0` clear-all keying is correct (keying on `new_tou` would have wiped the table whenever everything was deferred), the `start_period >= len(intents)` walk-back guard is genuinely required by the new `mode_at()` indexing, and the carry-forward of past strategic intents it depends on exists (`battery_system_manager.py:2298`).

Four findings, all in the new `needs_hardware_reconciliation` gate rather than the deferral logic itself — 2 medium, 2 low, left inline.

One non-finding: the `RuntimeError` in `_assign_hardware_slots` is now unreachable (`k + n <= 9` from the `max_intervals` cap makes `len(reclaimable) >= shortfall` always). Harmless retained invariant guard, not a bug.

core/bess/growatt_min_controller.py
```diff
@@ -685,11 +810,79 @@ def evaluate_intents(
...
+
+        def live(segments: list[dict]) -> set[tuple]:
+            return {
+                (s["start_time"], s["end_time"], s["batt_mode"])
+                for s in segments
+                if s.get("enabled", True)
+                and self._time_to_minutes(s["end_time"]) >= effective_minute
+            }
+
+        missing = live(self.active_tou_intervals) - live(
```
johanzander: **medium — orphan segments are never detected.**

`needs_hardware_reconciliation` computes `live(active_tou_intervals) - live(hardware)`, so it only detects *missing* segments, never extra ones. `evaluate_intents` is plan-vs-plan, and `sync_to_hardware` — the only place the disable side runs — now fires far less often by design.

Concrete: the inverter restores a stale slot 5 holding `18:00-20:00 grid_first` that the plan does not contain. The plan is stable, so no sync is triggered, and the battery exports for two hours against the plan.

The function's own docstring cites #551 as evidence the inverter alters its table unprompted — this is the half of that failure mode left uncovered, and the reduced write frequency makes the exposure window longer than before this PR. The orphan set is cheap to compute from the `planned_tou` the disable side already builds.

core/bess/growatt_min_controller.py
```diff
@@ -685,11 +810,79 @@ def evaluate_intents(
...
+            self._read_segments_from_hardware(controller)
```
johanzander: **medium — no attempt cap or backoff on a persistent mismatch.**

Any *persistent* readback mismatch becomes a write every optimization cycle for the rest of the segment's life.

Concrete trigger: `_normalize_segments` (line 1329) defaults `enabled` to `False` when the vendor payload omits the key. `live(hardware)` is then empty, every programmed segment reads as missing, and the controller rewrites the whole eligible set to the Growatt cloud every 15 minutes — strictly worse than the pre-PR behaviour, and the exact API-load condition #554 exists to reduce.

Same loop for any firmware that echoes a normalized `end_time` (e.g. `14:45` for our `14:44`), since `live()` compares the time strings verbatim.

core/bess/battery_system_manager.py
```diff
@@ -2505,6 +2505,30 @@ def _should_apply_schedule(
...
+                        self._controller, period
+                    )
+                )
+            except Exception as e:
+                # Not being able to look is not a reason to abandon the cycle.
+                # An unconfigured or unreachable inverter surfaces through the
+                # write path below, which records it as a pending write, and
+                # through the health checks — whereas raising here would kill
+                # the optimization itself, which needs no inverter at all.
+                logger.warning("Could not check the inverter for drift: %s", e)
```
johanzander: **low — the swallow comment's reasoning does not hold.**

It justifies itself with "surfaces through the write path below, which records it as a pending write", but that write path is only reached when `evaluate_intents` reports a difference. On a stable plan it never runs.

So an unconfigured `growatt_device_id` (`_read_segments_from_hardware` raises `SystemConfigurationError`) or an unreachable cloud yields a `logger.warning` every cycle and nothing else — `_hardware_write_pending` stays `False`. The behaviour is safe (health checks cover the user-visible part), but the stated reasoning is wrong and it produces ~96 warnings/day on an unconfigured install.

core/bess/battery_system_manager.py
```diff
@@ -2505,6 +2505,30 @@ def _should_apply_schedule(
...
 
+        # A write that raised is covered above. This covers the write that did
+        # not raise and still is not there: the schedule comparison below only
+        # ever compares plan against plan, so on a platform that skips the
+        # write while the plan holds steady, nothing would look at the inverter
+        # again until the plan changed (issues #551, #554).
+        if self._controller is not None and not prepare_next_day:
+            try:
+                drifted, drift_reason = (
+                    self._inverter_controller.needs_hardware_reconciliation(
```
johanzander: **low — duplicate cloud read per cycle.**

The drift check adds an unconditional `read_time_segments` cloud call on every optimization cycle (~96/day), and on cycles that do write, `sync_to_hardware` performs a *second* independent read at `growatt_min_controller.py:1367`.

Given #554's own premise that Growatt 500s on this endpoint under load, reads are worth counting alongside writes — the fresh read could be threaded through to the sync rather than repeated.

Issue #571, opened 2026-08-13
bug: discharge gate reads a raw snapped V grid difference while the policy uses _interpolate_value — shadow price 42% above the optimizer's own objective


johanzander:
Reported by @Frank-Leysen on #520 — Growatt MOD 5000TL3-XH, BESS 10.1.0b7, [debug bundle](https://gist.github.com/Frank-Leysen/1e80aa1269d9058934610cf3974025e3).

## Symptom

2026-08-13, SOLAR_STORAGE / Load First, EV charging + pool pump. `ems_discharging_rate` was 100% for every period 13:00→14:15, **0% for period 58 (14:30) only**, then 100% again at 14:45 — one isolated period importing ~1.1 kW from grid with SOC at 65% and 8 kWh of headroom.

## The seam

Policy selection (`_best_action_at_continuous_state`, `dp_battery_algorithm.py:1518`) and reported `future_value` (`:1774`, `:2021`) evaluate the value function through `_interpolate_value` at the continuous SOE. `_record_marginal_value` (`:913`) alone uses a raw backward difference between snapped grid cells:

```python
shadow_price = (V[t, i] - V[t, i - 1]) / SOE_STEP_KWH   # :913
allowed      = buy_price_t * efficiency_discharge >= shadow_price   # :915
```

For SOLAR_STORAGE the planned discharge rate is 0, so `max(planned, gate)` (`battery_system_manager.py:2686`) makes this the entire hardware decision.

`_record_marginal_value`'s own docstring cites the P1 mirrored-implementation bug class as the reason both *call sites* route through one function. The two **value estimators** were never unified the same way.

## Evidence

Replay at the production-faithful horizon (134 periods, terminal value 0.20289 via `_calculate_terminal_value`, `battery_system_manager.py:1923-2013`) reproduces the recorded shadow price bit-for-bit. The result is identical across horizon 38/134 × terminal 0/0.20289.

**1. Production's own objective contradicts it.** Finite-differencing `reward_objective_cost` over the exact cell the shadow price claims to measure, *on the production grid itself*:

| cell | production objective | `_record_marginal_value` reports |
|---|---|---|
| [9.8750, 9.9000] | **+0.17978** | **+0.25554** |

A **+42% error** — not a coarse-vs-fine comparison, an internal inconsistency.

**2. Concavity violation in `V[0]`.**

```
cell [9.8250, 9.8500] -> +0.23602
cell [9.8500, 9.8750] -> +0.23602
[...]

PR #579, opened 2026-08-14
fix: price the discharge gate off the cell the battery is actually in

johanzander:
## Summary

- The sub-period discharge gate read its value estimate from a different point on the value function than the policy walks, so it could hold the battery for a single period while the periods either side discharged.
- Adds `_value_slope_below`, the left one-sided derivative of the interpolant `_interpolate_value` defines, and routes `_record_marginal_value` through it.
- No planned energy moves: 142 of 2168 golden gate booleans flip, with `actions`, `intents`, `soe_trajectory` and cost bit-identical.

## Root cause

`_record_marginal_value` did its own index arithmetic (`dp_battery_algorithm.py:908`):

```python
i = round((soe - battery_settings.min_soe_kwh) / SOE_STEP_KWH)
shadow_price = float((V[t, i] - V[t, i - 1]) / SOE_STEP_KWH)
```

`round()` snaps to the **nearest** grid point. `_interpolate_value` (`:1436`) and `_local_value_slope` (`:1461`) both floor. So for any SoE in the lower half of a cell the gate was priced off the cell **below** — a region of `V` the battery is not in — while the policy walked a different curve.

Visible directly on the reporter's data, sweeping `initial_soe` at production resolution. The reported slope changes at `frac` 0.4→0.6, i.e. half-way **between** rungs, not at the rung boundaries:

```
initial_soe    idx      frac   shadow
9.8750     323.0000     1.00   0.19499
9.8800     323.2000     0.20   0.19499   <- same cell as below, different answer
9.8850     323.4000     0.40   0.19499
9.8900     323.6000     0.60   0.20806
9.8950     323.8000     0.80   0.20806
9.9000     324.0000     1.00   0.20806
```

`9.8800` and `9.8950` are both inside cell `[323, 324]` and must price the same kWh identically. That is the plateau-centred-on-the-rung signature of `round()`.

For SOLAR_STORAGE the planned rate is 0, so `max(planned, gate)` (`battery_system_manager.py:2688`) makes the gate the entire hardware decision, which is why a misread surfaces as an isolated 0% period rather than a small cost wobble.

**Not reproduced:** the issue's specific numbers (shadow 0.25554, gate CLOSED at 9.9000) do not reproduce in my replay — I get 0.20806 and an OPEN gate, presumably a horizon/terminal-value difference from the reporter's production run. The *defect* reproduces exactly; that one closed period does not, and periods 60/61 were never in the bundle. The mechanism and the corpus-wide effect below stand on their own evidence.

## Fix

`_value_slope_below(V_row, soe, battery_settings) -> float | None` returns the **left** one-sided derivative of the same piecewise-linear interpolant the policy walks. Left-sided because the gate authorizes energy *leaving* the battery, so it must price the value given up going down.

Deliberately **not** `_local_value_slope`: that is the right-sided reading, correct for the tie detector's noise magnitude and wrong here. Measured on the corpus it flips 395 decisions against this fix's 142, nearly all opening — a systematic over-open, not a fix.

`tie_detection` / the PWL re-solve are deliberately not wired in. `epsilon_for_period` is a currency band on two candidate *values*, not a EUR/kWh *level* — converting means dividing by `SOE_STEP_KWH`, which is the hand-derived epsilon P5 forbids — and the PWL window's `V` is solved with its exit SoE pinned (`:2111`), so its slope prices an artificial terminal constraint.

One thing found while implementing, worth a reviewer's eye: a one-sided derivative is discontinuous at each grid point, so it needs a tolerance the continuous interpolant does not. A battery resting exactly on its reserve floor yields `idx = +1e-16` rather than `0.0` in several fixtures; without `_GRID_POINT_TOLERANCE` that ulp prices a kWh below the floor that does not exist and reopens the ceiling #526 closed. The existing `test_discharge_gate_authorization_526` caught this on the first implementation (6 floor periods wrongly authorized).

`soe_levels` became unused in `_record_marginal_value` and was dropped from it and its two call sites.

## Test plan

- [x] `./scripts/quality-check.sh` passes (1853 passed, 31 skipped; Black/Ruff/frontend/TS/ESLint clean)
- [x] `.venv/bin/pytest -m slow` passes (537 passed, 6 skipped, 3m47s)
- [x] Observed end-to-end through the real hardware-write path, not just tests: took a real optimizer-produced flip (`realworld_2026_04_24_090423`, period 47) and drove the decision through the real `BatterySystemManager._apply_period_schedule`:

```
Gate    : False  ->  True   (pre-fix -> post-fix)
  pre-fix  gate=False -> discharge_rate = 0
  post-fix gate=True  -> discharge_rate = 100
```

  That is the reporter's symptom — the register that was writing 0% now writes 100%.

- [x] Full E2E stack (`docker-compose.ci.yml`, real backend + mock-HA) brought up and healthy; prices fetched, health checks green. The scheduler could not be driven to a fresh optimization there because the scenario pins a past `mock_time` and an accelerated container clock does not produce optimization runs in this project, so the observation above was taken through the real BSM path instead.

- Reverted: `lo = int(np.ceil(...)) - 1` → `lo = min(round(idx), len(V_row) - 1) - 1`, i.e. the pre-#571 nearest-grid-point snap.
- Result: `test_shadow_price_is_constant_within_one_grid_cell` FAILED (obtained 0.19499, expected 0.20806), and 20 of 38 `test_action_selector_parity` golden tests FAILED.
- Restored: tree clean, 55 passed.

Stated plainly: `test_gate_slope_stays_on_the_policys_interpolant` does **not** redden under that mutation — at the offsets it uses, `round()` and `ceil-1` pick the same cell. It is a standing guard against the two estimators drifting apart again, not a reproduction of this bug. The real-data cell test and the goldens are what catch it.

## Outcome-level coverage

- **Action-selector goldens** — `intra_period_discharge_allowed` re-pinned across 21 fixtures. Measured delta: **142 of 2168 periods flip, 141 opening a wrongly-closed gate, 1 closing**. `actions`, `intents`, `soe_trajectory` and cost are bit-identical on all 35 existing fixtures; I verified the regeneration itself touched no other field.
- **New fixture** `regression_2026_08_13_145213` built from the reporter's debug bundle via `from_debug_log.py`, plus its golden and a VPP baseline entry (`--add-new` only, no full re-baseline).
- **No `run_scenario_realized` / R == P test**, deliberately: the gate moves no planned energy, and `inverter_simulator.py:194-196` computes `deficit = max(0, home - solar)` for the load_first modes SOLAR_STORAGE/SOLAR_EXPORT map to, which is structurally zero at point-forecast resolution for exactly the intents that select them (the file argues this itself at `:118-126`). An R == P test would pass identically with the gate open or closed and would be coverage theatre. Reviewed and agreed independently.

## Scope assessment

Local — stays inside `_record_marginal_value`'s existing contract (same inputs, same two fields written); the new helper is a peer of the two existing value estimators in their established owner file. No parameter, flag, default-fallback, second construction site, or extra trigger routing around an ordering/timing problem: the diff *removes* a mirrored value estimator, which is what P1 exists to prevent.

## Documentation

`docs/agents/bess-knowledge.md` updated — it described `shadow_price` as a "backward difference between adjacent SoE grid levels", the exact mechanism this replaces. Added how the slope is now read and the bundle signature of the old bug. `docs/SOFTWARE_DESIGN.md` mentions neither `shadow_price` nor the gate, so nothing to change there.

Refs #571 — **does not close it.** The reported period (58, `soe = 9.9000`)
is unchanged by this PR: `idx = 324.0` sits exactly on a grid point, so
`round()` and `ceil()-1` select the same cell and the half-cell correction is
identically zero there. Verified against this branch's head; see the
reproduction in the comments. #571's own mechanism is the concavity violation
at that cell, which is a separate fix.

No longer blocked on Phase 4a: 4a has landed on `main`, and this branch is
merged up to it. The relocation is orthogonal to this diff —
`intra_period_discharge_gate` (the rate mapping) now lives in
`core/bess/execution_model.py`, while the value estimate this PR fixes stays in
`dp_battery_algorithm.py`, which `execution_model.py:90` names as its owner.

## Update after merging `main`

- Merged `origin/main` (`e1d334ae`). One conflict, in
  `vpp_baseline_v10_0_2.json`: resolved to main's copy, then this branch's
  `regression_2026_08_13_145213` entry re-added with
  `capture_vpp_baseline.py --add-new` — no full re-baseline, so the recorded
  v10.0.2 drift is untouched.
- `main` brought in fixture `regression_2026_08_12_202906`, whose golden was
  captured before this fix. Re-pinned: **5 of its 111 periods flip, 4 opening a
  wrongly-closed gate and 1 closing**, with `actions`, `intents`,
  `soe_trajectory` and `battery_solar_cost` bit-identical — the same gate-only
  signature measured on the other 21 fixtures. Corpus total is now 147 of 2279.
- Re-verified after the merge: `./scripts/quality-check.sh` green,
  `.venv/bin/pytest -m slow` 541 passed / 7 skipped, and CI green on all seven
  required checks.


bess-agent:
@claude-bot review

bess-agent:
Thanks — approved, and both points land.

**Root-cause gap (#571 stays open):** agreed, and that is deliberate. `Refs`, not `Fixes`. The concavity violation in the backward recursion's `next_i_cover = np.round(...)` is a separate mechanism in a separate code path and needs its own fix; this PR does not touch it.

**Redundant `_snapped_grid_index` call:** not taking this one, and the reason is the rule the PR exists to enforce. `has_value_cell_below` is the single owner of "is there a cell below this SoE" — it is exported precisely so tests stop re-deriving it, which is how the two estimators drifted apart in the first place (its docstring says so). Replacing the call with an inline `idx > 0.0` puts a second copy of that predicate in the file, which is the P1 mirrored-implementation shape this diff is removing. The cost is one subtraction, one division and a `round()` on a pure function; I would rather pay it than re-derive the rule one line after naming its owner.

No code change from this round.

PR #585, opened 2026-08-14
fix: stop blaming sensor configuration for a transient price-source outage

johanzander:
## Summary
- A transient 500 from the HA Nordpool integration told the user to fix a configuration that was correct
- The expected daily "tomorrow's prices aren't published yet" call was recorded as a user-visible runtime failure

## Root cause

From `docs/bess-debug-2026-08-14-080410.md` (v10.1.0b8), after a restart at 07:49:

```
08:00:09 WARNING .../nordpool/get_prices_for_date failed on attempt 1/4: 500 ...
08:01:26 ERROR   ... failed on final attempt 4/4: 500 ...
08:01:27 ERROR   ⚠️ SYSTEM DEGRADED: Critical sensor failures detected in required components: Electricity Price Data
08:01:27 ERROR   ⚠️ Please fix sensor configuration for full functionality.
08:04:05 INFO    Fetching Nordpool prices for 2026-08-14 using official integration
08:04:06 INFO    Successfully fetched 96 prices from official Nordpool integration
```

Nothing was misconfigured — config entry present, area SE4, same call succeeded three minutes later. The 500s are time-correlated, not configuration-correlated: every one in the last three bundles lands at `HH:00:08–09`, i.e. the hourly price fetch colliding with whatever HA does at the top of the hour. Usually attempt 2 succeeds.

**Not a regression.** These 500s appear in bundles back to v9.6.1 (June), and "Electricity Price Data" degraded first shows up 2026-08-12 in *both* 10.0.2.4/10.0.2.5 (stable) and 10.1.0b7/b8 (beta) — the frequency of the upstream failure changed, not our code.

1. `OfficialNordpoolSource` funnelled every failure through `except Exception -> ValueError`, so `perform_health_check` could not distinguish "not configured" from "transiently failed". `_run_health_check` then escalated any required component in ERROR to the config-blaming banner.
2. `_api_request`'s final-attempt branch called `record_failure_once(...)` unconditionally, ignoring the `suppress_retry_warnings` flag two lines above it.

## Fix

`official_nordpool_source.py` now raises the typed exceptions its two sibling price sources already use — `SystemConfigurationError` for a missing config entry, `PriceDataUnavailableError` for availability failures. `price_manager.get_price_data` already special-cased exactly those two types, so this aligns an outlier with the existing convention rather than inventing one.

The health status **stays ERROR** for an availability failure, because without prices the system genuinely cannot optimize.

> **Correction (review finding 4).** An earlier version of this section justified that by claiming health checks are not scheduled. That is wrong: `backend/app.py:442` registers `refresh_health_check` on `CronTrigger(minute="*/5")`, so a WARNING *could* have been re-escalated by a later run. The outcome is unchanged, but it now rests on the correct reason — a required price source with no data is an error at the moment it is observed, not a warning — rather than on a false premise about scheduling.

What changes is the explanation — and correspondingly the degraded banner no longer presumes the cause:

```
- ⚠️ Please fix sensor configuration for full functionality.
+ ⚠️ See the Health page for which check failed and why.
```

`ha_api_controller.py` gates the runtime-failure record on `not suppress_retry_warnings`, and the docstring now says so.

## Test plan
- [x] `./scripts/quality-check.sh` passes locally (1854 passed, 31 skipped)
- [x] `.venv/bin/pytest -m slow` passes (538 passed, 5 skipped, 315s)
- [x] Live `docker-compose.ci.yml` stack (mock-HA + real backend), price fetch failing for real:
  ```
  "name": "Nordpool Service Call",
  "status": "ERROR",
  "error": "Nordpool price fetch failed, will retry on the next fetch. If this persists, check that the Nordpool integration is still loaded in Home Assistant and re-run the setup wizard if it was re-added: ...",
  "value": "Unavailable"
  ```
  and in the container log: `⚠️ SYSTEM DEGRADED: required components failing: Electricity Price Data, Energy Monitoring` followed by `⚠️ See the Health page for which check failed and why.` — no configuration blame.
- [x] Defect 2 driven against a throwaway HTTP server returning real 500s through the real controller + `RuntimeFailureTracker`:
  ```
  today    -> raised HTTPError: 500 Server Error ... /api/services/nordpool/get_prices_for_date
  today    -> runtime failures recorded: 1
  tomorrow -> raised HTTPError: 500 Server Error ... /api/services/nordpool/get_prices_for_date
  tomorrow -> runtime failures recorded: 0
  ```

Written RED first, before any production change:

```
4 failed, 2 passed
FAILED test_official_nordpool_source.py::...::test_service_call_failure_raises_price_data_unavailable
FAILED test_official_nordpool_source.py::...::test_missing_config_entry_raises_system_configuration_error
FAILED test_official_nordpool_source.py::...::test_transient_service_failure_reports_warning_not_error
FAILED test_ha_api_controller.py::TestApiRequest::test_suppressed_failure_is_not_recorded
```

After the design changed from WARNING to ERROR-with-accurate-text in response to code review, the revised tests were re-run RED against the then-current code:

```
2 failed, 10 passed
FAILED test_official_nordpool_source.py::...::test_transient_service_failure_is_reported_as_unavailable_not_misconfigured
FAILED test_bsm_settings_and_lifecycle.py::TestRefreshHealthCheck::test_degraded_message_does_not_blame_sensor_configuration
```

- The BSM test failed on the literal `⚠️ Please fix sensor configuration for full functionality.` still being emitted
- Restored: all 161 tests in the three touched files pass; tree clean

Two tests passed RED **by design** — `test_missing_config_entry_still_reports_error` and `test_successful_call_reports_ok` pin behaviour that must *not* change, so that the fix cannot blunt the genuine-misconfiguration signal.

## Outcome-level coverage

- `test_degraded_message_does_not_blame_sensor_configuration` asserts on the operator-facing output of `refresh_health_check()` — the actual thing the user saw — not on an internal flag
- `test_transient_service_failure_is_reported_as_unavailable_not_misconfigured` asserts the health payload's status *and* text, and that it does not contain configuration language
- No plan-faithfulness (`R == P`) pin: the diff touches neither the DP, intent classification, nor control/rate mapping, so the `run_scenario_realized` requirement does not apply. Confirmed against the diff by the review agent.

## Scope assessment

Local. Neither fix leaves its current owner: the exception/health change stays inside the price source's existing health-check contract, the tracker change inside `_service_call_with_retry`'s existing final-attempt branch. The banner wording is a one-line change in the block that already prints it.

**Workaround check:** the diff adds no parameter, flag, default-fallback, second construction site, extra trigger, or branch whose job is to route around an ordering/timing/dependency problem. Fix 2 *deletes* an unconditional side effect and reuses the flag that already encodes the intent. Nothing suppresses the upstream 500 — it still logs, still retries 4×, still degrades the system. The HA-side hourly collision is not something this repo can fix; this is only about reporting it honestly.

## Documentation

Grepped `docs/agents/bess-knowledge.md` and `docs/SOFTWARE_DESIGN.md` for everything this diff touches: neither documents the price source's exception types, the health-status mapping, or the degraded-mode wording. The `SYSTEM DEGRADED` mention at `SOFTWARE_DESIGN.md:550` is about 404 sensors from #549, which still reports ERROR and still degrades — unaffected.

Closes #583

---

## Review round 2 (findings addressed)

**A stale config entry reported as a transient outage (medium).** Confirmed as a real regression this PR introduced — the pre-fix code said "fix sensor configuration", which was *correct* for that case. The suggested remedy (classify 4xx / "config entry not found" as `SystemConfigurationError`) does not work. Probed against a live HA with an unknown config entry:

```
HTTP 500
500 Internal Server Error
Server got itself in trouble
```

Byte-identical to the transient outage, no machine-readable reason — status and body cannot distinguish the two causes. So the message now asserts neither: it states the fetch failed and will be retried, and names the integration/wizard as what to check if it does not heal. That covers the stale-entry case without resurrecting #583's false accusation.

**Untyped `ValueError` on the unsupported-date guard (low).** Documented in the `Raises:` block, including that `price_manager.get_price_data` wraps it as `PriceDataUnavailableError` and that callers should not rely on that path. Left untyped deliberately — it is a caller bug, not an availability condition, and retyping it would make the wrapping *more* plausible-looking, not less.

**Config-entry test could not fail (low).** Correct — it asserted only the absence of a string in a hardcoded literal. Rewritten to assert what the message says (`"config entry"`, `value == "Missing"`). Both changed tests were mutation-checked:

| Mutation | Result |
|---|---|
| `"value": "Missing"` → `"N/A"`, error text → `"Setup incomplete"` | `test_missing_config_entry_error_names_the_configuration_problem` FAILED |
| dropped the wizard/integration pointer from the service-failure text | `test_transient_service_failure_is_reported_as_unavailable_not_misconfigured` FAILED |

Restored: 161 passed across the three touched files, `quality-check.sh` green.

**`suppress_retry_warnings` encodes which call, not which failure (low/medium).** Not changed — reasoning in the thread.


core/bess/official_nordpool_source.py
```diff
@@ -197,6 +211,22 @@ def perform_health_check(self):
             service_check.update(
                 {"status": "OK", "value": f"{len(prices)} hourly prices available"}
             )
+        except PriceDataUnavailableError as e:
```
johanzander, line 223: **medium — a stale config entry now reports as a transient outage**

This branch labels *every* non-empty-config failure "temporarily unavailable, will retry on the next fetch", but `config_entry_id` is only ever checked for emptiness, never validity. If the user deletes and re-adds the Nordpool integration in HA — a routine operation that mints a new entry id — the service call is rejected permanently, and that now surfaces on the Health page as a transient outage that will heal itself. The degraded banner has also been stripped of any configuration hint, so nothing anywhere tells the user to re-run the wizard.

That is the exact inverse of #583: a genuine misconfiguration reported as a transient outage. Distinguishing on the HA response (4xx / "config entry not found" → `SystemConfigurationError`) would keep both signals honest.

core/bess/ha_api_controller.py
```diff
@@ -1169,8 +1170,11 @@ def _api_request(
                         str(e),
                     )
 
-                    # Record runtime failure if failure tracker is available
-                    if self.failure_tracker:
+                    # Record runtime failure if failure tracker is available.
+                    # A suppressed failure is an expected condition (see
+                    # suppress_retry_warnings above) — recording it would surface
+                    # a routine daily event as a user-visible failure (#583).
+                    if self.failure_tracker and not suppress_retry_warnings:
```
johanzander, line 1177: **low/medium — `suppress_retry_warnings` encodes which *call*, not which *failure***

It is set at the call site purely by `target_date == tomorrow_date` (official_nordpool_source.py:98). Suppressing the runtime-failure record therefore also swallows auth failures, connection refusals, and a permanently broken config entry on the tomorrow path.

Partly mitigated: the today-date call is not suppressed and runs every 5 minutes via the health-check cron, so a whole-integration outage still gets recorded. But a failure specific to the tomorrow request now leaves no trace at all.

core/bess/official_nordpool_source.py
```diff
@@ -51,12 +52,16 @@ def get_prices_for_date(self, target_date: date) -> list[float]:
             List of hourly prices per kWh (VAT-exclusive)
 
         Raises:
-            ValueError: If prices cannot be fetched
+            SystemConfigurationError: If the integration is not configured
+            PriceDataUnavailableError: If prices cannot be fetched right now
```
johanzander, line 62: **low — the docstring is now wrong for one path**

The "only today and tomorrow supported" guard further down still raises a bare `ValueError`, and it sits *outside* the `try`, so it escapes the new classification untouched. This `Raises:` block documents only the two typed exceptions.

Downstream, `price_manager.get_price_data` special-cases exactly those two types and wraps everything else — so this programming error becomes a `PriceDataUnavailableError`, i.e. an availability label on what is a caller bug.

core/bess/official_nordpool_source.py
```diff
@@ -197,6 +211,22 @@ def perform_health_check(self):
             service_check.update(
                 {"status": "OK", "value": f"{len(prices)} hourly prices available"}
             )
+        except PriceDataUnavailableError as e:
+            # Still ERROR — without prices the system genuinely cannot optimize.
```
johanzander, line 224: **low — the stated rationale rests on a false premise**

The PR body (and this comment's reasoning) justifies keeping ERROR rather than WARNING on the grounds that health checks are not scheduled — only `start()`, the debug export, and the API call them — so nothing would ever re-escalate a persistently dead price source.

That is not the case: `backend/app.py:441` registers `self.system.refresh_health_check` on `CronTrigger(minute="*/5")`. The chosen outcome may well still be right, but it was argued from a fact that does not hold, and the same wrong premise will mislead whoever revisits this. Worth correcting in the PR body before merge.

core/bess/tests/unit/test_official_nordpool_source.py
```diff
@@ -0,0 +1,109 @@
...
+        assert result["status"] == "ERROR"
+        service_check = next(
+            c for c in result["checks"] if c["name"] == "Nordpool Service Call"
+        )
+        assert service_check["status"] == "ERROR"
+        assert "temporarily unavailable" in service_check["error"].lower()
+        assert "retry" in service_check["error"].lower()
+        assert "not configured" not in service_check["error"].lower()
+
+    def test_missing_config_entry_error_names_the_configuration_problem(self):
```
johanzander, line 86: **low — this test cannot fail**

It asserts only `"temporarily unavailable" not in config_check["error"]`, where that error is the hardcoded literal `"No config entry ID configured"` produced by the early return. It passes identically against pre-fix code and cannot fail for any realistic regression — it does not test the "distinguishable by their text" property its docstring claims.

Asserting the positive — that the text names the config entry, or points at the wizard — would make it discriminate.

johanzander (commented):
Code review. The core change is sound — typed exceptions replacing the generic `ValueError` funnel, and gating `record_failure_once` on `not suppress_retry_warnings`. Every caller of `get_prices_for_date` / `get_tomorrow_prices` / `check_health` catches broadly (`price_manager.get_price_data`, `_get_price_data`, `debug_data_exporter._serialize_price_data`), so the exception-type change breaks nothing, and there are no stale references to the removed log strings anywhere in the repo.

Five findings inline, most important first. One (the PR body's rationale for keeping ERROR) is a factual correction worth making before merge.

core/bess/official_nordpool_source.py
bess-agent: Confirmed as a regression this PR introduced — the pre-fix code said "fix sensor configuration", which was *correct* for a stale entry.

The proposed discriminator does not work, though. Probed against a live HA with an unknown config entry (both a malformed id and a well-formed ULID that does not exist):

```
HTTP 500
500 Internal Server Error
Server got itself in trouble
```

Byte-identical to the transient outage, and the body carries no machine-readable reason. Neither status code nor body can separate the two causes, so any classification here would be a guess wearing a type name.

Fixed by not asserting either cause. The message now states what is known and names what to check if it does not heal:

> Nordpool price fetch failed, will retry on the next fetch. If this persists, check that the Nordpool integration is still loaded in Home Assistant and re-run the setup wizard if it was re-added: ...

That restores the configuration hint for the stale-entry case without resurrecting the #583 accusation for the transient one. Validating the entry properly would need a `config_entries/get` WS round-trip in the `Configuration Entry` check — real, but a bigger change than this PR, and worth its own issue if you want it.

core/bess/ha_api_controller.py
bess-agent: Not changing this one — pushing back with reasoning rather than expanding scope.

The observation is accurate: the flag is set by `target_date == tomorrow_date`, so it does describe the call, not the failure. But the failure modes it could hide are already covered or are the expected case:

- **Auth failure, connection refused, whole-integration outage** — the today-date call is unsuppressed and runs on the same 5-minute `refresh_health_check` cron, so any of these is recorded within five minutes, from the same controller and the same service domain.
- **A failure specific to the tomorrow request** — that is precisely "the market has not published tomorrow yet", which is the expected daily condition the flag exists to describe. Recording it was the bug in #583.

Narrowing the flag to describe the failure rather than the call would mean classifying on the HA response, and the probe in the other thread shows the response does not carry enough to classify on. So the change would add a mechanism without adding a signal.

Happy to revisit if you would rather have the tomorrow path record failures once past the publication hour (~13:00 CET) — that is a real distinction and would be a small follow-up, but it is a behaviour change beyond this fix.

core/bess/official_nordpool_source.py
bess-agent: Correct on both points. The `Raises:` block now documents it:

```
ValueError: If target_date is neither today nor tomorrow. This is a
    caller error, raised before the fetch is attempted, and is
    deliberately left untyped — ``price_manager.get_price_data``
    wraps it as PriceDataUnavailableError, which mislabels a
    programming error as an availability problem, so callers should
    not rely on that path.
```

Left untyped on purpose. Retyping it to `PriceDataUnavailableError` would make the downstream mislabel *consistent* rather than fix it — a caller passing a date three days out is a bug, not an availability condition, and the honest options are either a distinct exception type or the current documented wart. Not worth widening this PR for; the docstring now stops it being a silent trap.

core/bess/official_nordpool_source.py
bess-agent: Verified and the premise was wrong. `backend/app.py:442`:

```python
self.scheduler.add_job(
    self.system.refresh_health_check,
    CronTrigger(minute="*/5"),
    misfire_grace_time=30,
)
```

So a WARNING would have been re-escalated by a later run, and the reason I gave for rejecting it does not hold. Corrected in the PR body with an explicit note rather than a silent edit, since the wrong premise is the kind of thing that gets inherited.

The outcome stands on its own reasoning: a required price source with no data is an error at the moment it is observed, not a warning. The 5-minute cron changes how quickly it self-corrects, not what it is.

core/bess/tests/unit/test_official_nordpool_source.py
bess-agent: Correct — it asserted the absence of a string from a hardcoded literal, so it passed against pre-fix code and could not fail. Rewritten to assert what the message says:

```python
config_check = result["checks"][0]
assert config_check["status"] == "ERROR"
assert "config entry" in config_check["error"].lower()
assert config_check["value"] == "Missing"
```

Mutation-checked, along with its sibling:

| Mutation | Result |
|---|---|
| `"value": "Missing"` → `"N/A"`, error text → `"Setup incomplete"` | `test_missing_config_entry_error_names_the_configuration_problem` FAILED |
| dropped the wizard/integration pointer from the service-failure text | `test_transient_service_failure_is_reported_as_unavailable_not_misconfigured` FAILED |

Restored: 161 passed across the three touched files.

PR #588, opened 2026-08-14
chore: make the OS sandbox the safety boundary; drop the Bash guard hooks

johanzander:
Makes the OS sandbox the safety boundary, cuts approvals to the two commands whose effects escape the repository, and removes the two `Bash` permission hooks that were standing in for both. An agent should run `implement-issue` end-to-end and stop exactly once, at the design confirm gate.

> [!NOTE]
> **Verified.** A fresh-session `verify-sandbox.sh` run passes every check, `gh` and `podman` included. Check 0 did not fire, so it was a genuinely sandboxed run.
>
> **The sandbox captures its policy once, at activation, and ignores later edits.** This bit twice. It *looks* like live-reload — editing `.claude/settings.json` does activate the sandbox, so the editing session finds itself sandboxed and concludes the config re-reads. It does not: a probe write to a path added to `allowWrite` minutes earlier still returned `Operation not permitted`. Iterating on `sandbox.*` in one session produces confident results measured against the original config. One edit, one fresh session, one run — never a second edit in the same session.

## The whole permission surface, after this PR

| | Setting |
|---|---|
| **sandbox** | `enabled`, `allowWrite: ["."]`, `excludedCommands: ["gh"]` |
| **defaultMode** | `auto` — unlisted commands go to a classifier, never to you |
| **ask** | `git push --force *`, `sudo *` |
| **deny** | `podman machine rm`, `podman system reset`, every mutating `git stash` form |

That is the entire list. `rm -rf`, `git reset --hard`, `rebase`, `merge`, `git branch -D`, `git worktree remove` all run unattended.

## Why `rm -rf` is safe to leave unattended

The sandbox confines every Bash write to the repository, decided by the OS from the actual syscall rather than guessed from a command string. Outside the repo the macOS profile denies `file-write-create` and `file-write-unlink` in a single rule, so `rm -rf ~` cannot create *or* unlink.

**`allowWrite` must name the repo root, and that is the whole trick.** Writes are `allowOnly` minus `denyWithinAllow`, and the built-in `allowOnly` is only `/dev/*`, `/tmp/claude`, `~/.npm/_logs`, `~/.claude/debug` — the repository is not in it.

The abandoned `chore/sandbox-replaces-worktree-hook` branch set `allowWrite: [".claude", ".git", "scripts"]`: three paths that are *all on the built-in deny list*, never opening the repo root. It therefore allowed essentially nothing, every git operation failed, and its tip commit concluded the sandbox was unusable here. The conclusion was wrong; the config was. This PR supersedes that branch.

## What stays denied, and what it costs

There is no allow-within-deny primitive for writes (reads have one, which is why `allowRead` differs), so no `allowWrite` entry overrides the built-in `denyWrite` list.

| Denied | Consequence |
|---|---|
| `.git/config`, `.git/worktrees` | Create worktrees with **`EnterWorktree`**, never `git worktree add` from Bash — the harness is not sandboxed |
| `.claude/**`, `scripts/**`, `.github/**` | Edit with the **Edit/Write tools**, which the sandbox does not govern; Bash writes there fail |
| `package.json`, `package-lock.json` | `npm install` fails. `worktree-setup.sh` symlinks `node_modules`, so this bites only when a branch needs its own deps |
| — | A `git merge origin/main` touching `.claude/` or `scripts/` fails; finish it with Edit/Write |
| — | `.git/objects`, refs and index are **not** denied → commit, branch, reset, reflog work normally |

## Hooks

| Hook | Fate |
|---|---|
| `auto-allow-worktree-destructive.sh` | **removed** — a cwd-conditional auto-allow; the sandbox does this job from the syscall |
| `link-worktree-local-settings.sh` | **removed** — symlinked `settings.local.json` into worktrees to undo the asymmetry the first hook created |
| `check-worktree-path.sh` | **kept** |
| `pretest-on-commit`, `optimizer_core_principles`, `lint-edited-python` | untouched (quality, not permissions) |

The two removed hooks cost six distinct false positives in a single day, every one introduced by the fix to the previous one, including blocking a live beta release. They tried to infer which files a command would touch by parsing the command string; the sandbox is what that code was failing to be.

`check-worktree-path.sh` stays because it is a different shape: it guards Edit/Write rather than Bash, compares two `git rev-parse` results as strings, never parses a command, and has never produced a false positive. It also catches what the sandbox structurally cannot — the Edit/Write tools are not sandboxed at all, so a stale absolute path writing into the main checkout while ~20 worktrees are live is *only* caught here. It caught exactly that while this branch was being written.

## `defaultMode: auto`

Load-bearing, and the gap in the first two commits. Cutting the ask list controls what *prompts*; everything **unlisted** is governed by the mode, which was set nowhere. Under the plain `default` mode that is a prompt for every command outside the `allow` list — compose, npm, python3, mkdir, cp and a dozen one-off shapes a single `implement-issue` run reaches for.

It lives in the **tracked** settings deliberately: `defaultMode` is not a Bash rule, so it travels into every worktree on its own — no hook required. Keeping it in the gitignored `settings.local.json` is precisely what made an autonomous run in the main checkout start prompting the moment it entered a worktree, and what `link-worktree-local-settings.sh` existed to paper over.

## `git stash`

Denied rather than prompted, because it is the one command that destroys *another agent's* uncommitted work rather than your own — one `refs/stash` per repository, shared by every worktree, with no owner. Every mutating form is listed; `list`/`show` still work. The rules match the command as written, so `git -C <dir> stash pop` slips past where the hook normalised it — noted in CLAUDE.md.

## `scripts/verify-sandbox.sh`

Ported from the abandoned branch and corrected. Two checks asserted the wrong outcome — `git worktree add` and a Bash write to `.claude/**` were expected to **succeed**, so a correct config scored two failures. They now assert blocked, and a green run means the sandbox behaves as CLAUDE.md claims.

Its `rm -rf` probe was dropped rather than written: proving an unlink outside the repo is blocked needs a file outside the repo that already exists, and creating one is itself blocked — so every candidate target is one of your real files. A probe that deletes `~/.zshrc` when the sandbox is off is worse than no probe. Check 0 covers it soundly, since create and unlink are denied by the same rule.

## Measured results

`scripts/verify-sandbox.sh`, run via the Bash tool with the sandbox live:

| Check | Result |
|---|---|
| write outside the repo is blocked | **PASS** — `rm -rf` is contained; this was the whole point |
| write inside the repo is allowed | **PASS** — `allowWrite: ["."]` resolves correctly |
| `git worktree add` is blocked | **PASS** (expected — use `EnterWorktree`) |
| Bash writing `.claude/settings.json` is blocked | **PASS** (expected — use Edit/Write) |
| Bash writing `scripts/**` | *allowed* (informational) |
| `gh` reaches GitHub | **FAIL** |
| `podman` reaches its VM | **FAIL** |

Two earlier claims are withdrawn as measured-false:

- **The deny list is per-file inside `.claude`, not the whole directory.** `touch .claude/.sandbox-probe` succeeds. The ported check probed exactly that path and would have reported the directory writable; it now probes `.claude/settings.json`, which is denied.
- **`scripts/**` is writable.** The claim that `scripts/`, `.github/` and the lockfiles are denied came from misreading the binary's *GitHub Actions* default config (it lists `~/actions-runner` and `GITHUB_EVENT_PATH`) as the local one.

## Why each non-default knob exists

Every one was found by reading an actual error message. Every wrong guess along the way came from reasoning about what *ought* to be blocked. Don't drop one because it looks redundant:

| Knob | The measured failure it answers |
|---|---|
| `enableWeakerNetworkIsolation` | `gh` → `x509: OSStatus -26276`. Go binary can't reach `trustd` to verify TLS. Egress was never the issue — `curl https://github.com` returned **200** in the same sandbox. Explicitly weaker: it opens `trustd`. |
| `network.allowMachLookup` (`SecurityServer`, `securityd`, `trustd`) | `gh auth status` → *"The token in keyring is invalid"*. The macOS keychain is XPC, not network. Same block made `git push` emit `failed to store: 100001` (push still landed). |
| `network.allowLocalBinding` | `podman info` → `dial tcp 127.0.0.1:64752: connect: operation not permitted`. The VM is reached over local TCP, so `allowUnixSockets`/`allowRead` were both wrong. |
| `allowWrite: [".", "~/GitHub/bess-manager"]` | `frontend/node_modules` and `.venv` are symlinks into the main checkout; writes through them resolve outside a worktree's root → `EPERM` on `node_modules/.vite-temp`, breaking `vitest`. Worktrees are nested inside the main checkout, so one entry covers both. |

The hardcoded `~/GitHub/bess-manager` is ugly in a tracked file. It's there because the alternative — a user-level `allowWrite` — is refused by the auto-mode classifier, correctly: that would be a session widening its own containment.

`sandbox.excludedCommands` is **not used and not needed**. It was tried in both project and user settings while the four knobs above were missing, appeared to do nothing, and is moot now.

## Verification

- **`scripts/verify-sandbox.sh` — all checks PASS in a fresh session**, `gh` and `podman` included.
- `./scripts/quality-check.sh` — 0 errors, 0 warnings.
- The script gains a probe it was missing: nothing exercised the **symlinked dependency trees**, yet that was a real blocker. It doesn't present as a permissions problem — it surfaces as vitest failing to load its config with `EPERM`, breaking Step 6 in every worktree for no visible reason. A check now fails loudly if that `allowWrite` entry is ever removed.

**Coverage limits, stated rather than glossed:** the run exercises `gh auth status` (keychain + TLS — most of what `gh pr create` needs) and `podman info` (the socket connection only, *not* that a compose E2E completes). A full Step 8 and a real `gh pr create` remain the first end-to-end proof.

https://claude.ai/code/session_019LeVZcjErmJnJKa4ix4BLE


johanzander:
## Code review (post-merge)

Reviewed `chore/drop-permission-guard-hooks` (merged as `0579ffb9`) — deletes the two Bash permission hooks, enables the OS sandbox, rewrites the permission lists, adds `scripts/verify-sandbox.sh`.

### High — permission patterns don't match the forms they claim to

1. **`.claude/settings.json:59`** — the `git stash` deny list enumerates literal subcommands, so option-first forms match nothing and fall through to `defaultMode: auto`: `git stash -u`, `-k`, `-p`, `--include-untracked`, `--all`. `-u` is one of the most common forms. `CLAUDE.md:256` claims the deny covers "every mutating form"; the old hook normalised the command, so this is a regression, not a pre-existing gap. Fix: add `Bash(git stash -*)` and `Bash(git stash --*)`.

2. **`.claude/settings.json:76`** — `Bash(git push --force *)` is a literal prefix match, so `git push -f origin main` and bare `git push --force` (no args) match nothing — and `Bash(git push *)` is in the **allow** list at line 43, so both run silently. The one shape the PR says still asks does not reliably ask.

3. **`.claude/settings.json:76`** — none of the old hook's GitHub-reaching guards survive: pushes moving a shared ref (`main`, `beta`, `beta-release-*`, tags) are allowed outright, and `gh pr merge` / `release` / `repo edit` / `secret` / `workflow run` / non-GET `gh api` are unlisted so `auto` decides. The sandbox provides no containment for network effects, so "the sandbox is the safety boundary" doesn't cover this class. Concretely: an autonomous run can now `git push beta main` or `gh pr merge` a release PR unattended — both of which the project's own rules require explicit instruction for.

### Medium

4. **`.claude/settings.json:76`** — the safety argument for leaving `rm`, `git reset --hard`, `git branch -D` unattended rests on "tracked content survives in the object database, discarded commits survive by SHA in the reflog". But `git gc --prune=now`, `git reflog expire --expire=now --all` and `git tag -d` previously asked and are now unattended, and `.git/objects`/refs are deliberately not sandbox-denied. The recovery mechanism the risk model depends on can itself be destroyed without a prompt.

5. **`scripts/verify-sandbox.sh:85`** — the worktree probe decides success from `[ -d "$probe_wt" ]` rather than `git worktree add`'s exit status. `git worktree add` creates the target directory before writing `.git/worktrees`, and `.claude/worktrees/<name>` is not on the deny list (the script itself notes at line 118 that only specific `.claude` files are denied). So a correctly-blocked run can leave the directory behind → `wt=created` → FAIL reported on a correct config, `git worktree remove --force` then fails silently, stray directory left behind. `wt_err` is captured on line 84 but never printed in the `blocked` branch, so the real error is lost either way.

6. **`scripts/verify-sandbox.sh:141`** — check 4c is the only guard against dropping the `~/GitHub/bess-manager` `allowWrite` entry, but it can't discriminate when run from the main checkout, which is the default place to run it: there `frontend/node_modules` is a real directory inside `.`, so the probe passes regardless. The claim that "a check now fails loudly if that `allowWrite` entry is ever removed" only holds inside a worktree. It should assert it's running in a linked worktree (or skip loudly) before treating a PASS as meaningful.

7. **`scripts/verify-sandbox.sh:165`** — the podman remedy text still says "add it to `sandbox.filesystem.allowRead` or exclude podman", which is the wrong knob this PR spent several commits disproving (the fix is `network.allowLocalBinding`, per CLAUDE.md). When this check next fails it sends the reader down the exact dead end the PR documented.

### Low

8. **`scripts/verify-sandbox.sh:156`** — same for the `gh` remedy: it recommends `sandbox.excludedCommands` and "move the exclusion to `~/.claude/settings.json`", while CLAUDE.md now records `excludedCommands` as unused/ineffective and notes a user-level widening is refused by the auto-mode classifier.

9. **`CLAUDE.md:238`** — "Worktrees are nested inside the main checkout, so one entry covers both" is false for the sibling-folder layout (`../bess-manager-feature/`), which `CLAUDE.md:180` still presents as first-class. A sibling worktree is under `~/GitHub/`, not `~/GitHub/bess-manager`; it only works because `"."` resolves to the session's own cwd, and a main-checkout session operating on a sibling is silently blocked.

10. **`scripts/quality-check.sh:150`** — the gate that ran the permission-decision test matrix is removed with nothing replacing it. `verify-sandbox.sh` can't fill the gap: it exits 2 unless run by the Bash tool in a sandboxed session, so it can never be a CI/pre-commit check. The permission surface is now entirely unguarded by automation — the condition #562's "a gate that cannot run its checks must not report success" principle was added to prevent.

---

Findings 1–3 are the ones worth fixing first; they're straightforward pattern additions to `.claude/settings.json`.

PR #596, opened 2026-08-15
fix: close the permission-pattern gaps; assert rules by command string

johanzander:
Closes the permission-pattern gaps found by review on #588, plus four rounds of follow-up review on this PR. The rules in #588 matched the command **as written** — prefix globs, no normalisation — so a series of them read as covered while matching nothing.

## Final permission surface

| | Rules |
|---|---|
| **deny** | every mutating `git stash` form, incl. `-*` / `--*` **and `git -* stash`**; `podman machine rm`; `podman system reset` |
| **ask** | every `git push` · every `gh api` · `gh pr merge` / `release` / `repo edit` / `secret` / `workflow run` · `git gc` / `prune` / `repack` / `maintenance` / `reflog expire` / `reflog delete` / `update-ref` / `tag -d` / `--delete` / `-f` · a `git -* <verb>` twin for each of the above · `sudo` |

The `git -*` twins name a **verb** deliberately. A blanket `Bash(git -*)` was tried and did two bad things at once: it gated read-only inspection (`git -C sub status`, `git -C <wt> diff` — stalling `rules.md`'s cross-checkout recipe), and it matched `git -C x stash pop` while no *deny* rule did, silently downgrading the stash prohibition to a one-keystroke prompt.

`rm`, `git reset --hard`, `rebase`, `merge`, `git branch -D`, `git worktree remove` stay unattended — the sandbox contains them.

## The root cause, stated once

Three separate findings came from the same mistake, so it's now in CLAUDE.md rather than fixed case-by-case:

**An `ask` that fails to match does not fall through to the `auto` classifier.** It falls through to whatever `allow` covers the command — `Bash(git push *)` here, `Bash(git *)` in user settings — and runs unattended. Precedence is `deny > ask > allow` **by category**, not by specificity. So a too-narrow `ask` is worse than no rule: it reads as covered while changing nothing.

## What was leaking

**Markers at an arbitrary argument position** — unreachable by prefix globs, so `git push` and `gh api` are now guarded bluntly:
```
git push origin main --force            # --force not adjacent to "push"
git push origin +beta-release-9.9       # force via refspec
git push origin --delete release-9.9    # destroys a shared ref
git push origin v9.9.0                  # publishes a release tag
gh api repos/o/r/pulls/N/merge -X PUT   # merges, bypassing `gh pr merge`
gh api <path> -f key=val                # any -f/-F makes it a POST
```

**git's global options before the subcommand** — every rule anchored on `git stash` / `git push` / `git gc` was sidestepped:
```
git -C ../bess-manager-feature stash pop      # takes another agent's stash
git -C .claude/worktrees/x push origin main
git --git-dir=/tmp/r/.git stash drop
git -c push.default=current push beta main
git --no-pager gc --prune=now
```
The deleted hook normalised for exactly this, and CLAUDE.md *teaches* `git -C .claude/worktrees/<name>` as the cross-checkout idiom — so it's the spelling an agent reaches for first. `Bash(git -*)` covers the class in one rule, including options nobody has thought of yet.

**Long-option spellings**: `git tag --delete` ran unattended while `git tag -d` asked. `git prune` / `git repack -d` join the same category.

## The method change

Presence checks — "is rule X in the list" — passed for four consecutive review rounds while real spellings leaked, because they answer the wrong question.

`quality-check.sh` now carries **56 real command strings** in three lists, using the same prefix-glob semantics the harness applies:

| List | Checked against | Catches |
|---|---|---|
| `MUST_BE_DENIED` (20) | `deny` **only** | an `ask` certifying a command that must be blocked |
| `MUST_BE_GUARDED` (29) | `deny` + `ask` | rules too **narrow** to match real spellings |
| `MUST_NOT_BE_GUARDED` (7) | must match **nothing** | rules too **broad**, gating read-only work |

The single-list version was itself part of the bug: it merged `deny + ask`, so it reported `✅ Permission surface intact` on the stash downgrade described above — the gate written to catch that class certified it.

Both directions verified to discriminate rather than pass vacuously. Re-adding the blanket rule:
```
❌ read-only command is gated by Bash(git -*): git -C sub status --short
❌ read-only command is gated by Bash(git -*): git -C .claude/worktrees/x diff -- file.py
```
And it closes a mutation the reviewer demonstrated: nine stash deny rules — including `clear` and `drop`, the two that irreversibly destroy another agent's entries — could previously be deleted with the gate still green.

New bypass spellings go into the right list first, then the rule is fixed until the gate is green.

It also regained the two podman denies it was missing — the only rules guarding a shared VM whose destruction is unrecoverable, and its own comment said *"a rule absent from here is a rule that can be silently removed."*

## Other fixes

- **`quality-check.sh` ran the gate as a bare heredoc under `set -e`**, so a missing rule aborted the whole script: `ERRORS` never incremented, the `if [ $? ]` below was dead code, and the scenario check, markdown check and summary never printed. Wrapped in `if ! … then`; verified by deleting a rule and watching the run continue to `Errors: 2`.
- **`verify-sandbox.sh` worktree probe** read `[ -d $dir ]` instead of exit status. `git worktree add` creates the directory *before* the denied write, so a correctly-blocked run scored a false FAIL and left a stray directory. Its remedy also printed `Error was:` with nothing after it — `wt_err` is populated only on the branch where the remedy never prints.
- **`verify-sandbox.sh` dangling symlink**: an `-e`→`-L` swap made a broken link (main checkout's `node_modules` reinstalled while a worktree still points at it) report a sandbox misconfiguration and send the reader to `allowWrite` — the wrong knob. Now three explicit states, verified against all four filesystem shapes.
- **`Bash(git push *)` removed from `allow`** — dead under category precedence, but the most misleading line in the file for anyone auditing it.
- **CLAUDE.md corrected** where it contradicted the shipped rules: it claimed only two shapes ask, and that `--force-with-lease` runs unattended.

## `--force-with-lease` asks, deliberately

It can't be carved out: any exemption pattern is prefix-anchored, so it would miss `git push origin main --force-with-lease`, while a pattern loose enough to catch that also catches plain `--force`.

The cost is near zero. Neither `implement-issue` nor `release` mentions `rebase`, `--force` or `--force-with-lease`, and this project merges the target branch before a PR rather than rebasing — a merge-based flow never force-pushes. **If the prompt ever fires, that is the finding**: an agent has gone off-script. Answer it on its merits; don't narrow the push rule, which is how findings 1–3 got introduced.

## Also fixed in review

- **`git maintenance run --task=gc`, `git reflog delete`, `git tag -f`, `git update-ref -d`** all fell through to the user-settings `Bash(git *)` allow — same enumeration-gap class, in the category whose whole job is protecting the recovery mechanism.
- **`docs/agents/rules.md`** still cited "the permission hook" for the stash prohibition; the hooks were removed in #588. It now names `permissions.deny` and states that the `git -C` spelling used two paragraphs below is covered.
- **CLAUDE.md's category table** omitted `prune`, `repack` and the `git -*` twins, and claimed every mutating stash form was denied while the downgrade was live.

## Verification

- `quality-check.sh` — `✅ Permission surface intact (56 command shapes checked, 20 require deny, 7 must stay unattended)`, with both failure directions proven to discriminate.
- The **frontend suite fails `EPERM`** in the session that wrote this: that session captured `allowWrite: ["."]` at sandbox activation, before the main-checkout entry existed. Not a regression from this PR — a fresh session verified the widened entry. Worth confirming on a fresh run rather than taking my word for it.
- `verify-sandbox.sh` changes (worktree probe, symlink states) have **not** been re-run under a sandbox. The next fresh-session run is their first real test.

https://claude.ai/code/session_019LeVZcjErmJnJKa4ix4BLE


.claude/settings.json
```diff
@@ -70,10 +70,29 @@
...
       "Bash(git stash branch *)",
       "Bash(git stash create)",
-      "Bash(git stash store *)"
+      "Bash(git stash store *)",
+      "Bash(git stash -*)",
+      "Bash(git stash --*)"
     ],
     "ask": [
-      "Bash(git push --force *)",
+      "Bash(git push --force*)",
```
johanzander: **Dropping the space widened this glob onto `--force-with-lease`.** The old pattern `Bash(git push --force *)` did not match `git push --force-with-lease ...` (no space after `--force`); `--force*` does.

CLAUDE.md states `--force-with-lease` is deliberately auto-allowed because git itself refuses the dangerous case. With this pattern, every post-rebase `git push --force-with-lease` in an autonomous `implement-issue` run stalls on a prompt — the exact stall the exclusion exists to prevent.

Fix: `"Bash(git push --force)"` + `"Bash(git push --force *)"`, or update CLAUDE.md to say lease-pushes now ask.

.claude/settings.json
```diff
@@ -70,10 +70,29 @@
...
+      "Bash(git push -f*)",
+      "Bash(git push origin main*)",
```
johanzander: **This only matches when the args are literally `origin main...`.** It misses `git push -u origin main`, `git push origin HEAD:main`, and a bare `git push` on a `main` checkout — all of which fall through to the `allow` entry `Bash(git push *)` under `defaultMode: auto`.

An autonomous run doing `git push -u origin main` therefore publishes to the shared ref unattended, which is the hole this change is meant to close.

.claude/settings.json
```diff
@@ -70,10 +70,29 @@
...
+      "Bash(git push *beta-release-*)",
+      "Bash(git push --tags*)",
+      "Bash(git push *--tags*)",
+      "Bash(git push *refs/tags/*)",
+      "Bash(gh pr merge*)",
+      "Bash(gh release*)",
+      "Bash(gh repo edit*)",
+      "Bash(gh secret*)",
+      "Bash(gh workflow run*)",
+      "Bash(gh api -X*)",
```
johanzander: **These two only match when the flag directly follows `gh api`, and they miss gh's implicit-POST form entirely.** Unmatched, and therefore running under `auto`:

- `gh api repos/o/r/releases -f tag_name=v1` (any `-f`/`-F` makes it a POST)
- `gh api <path> --input body.json`
- `gh api repos/o/r -X DELETE` (flag after the path)

That makes the `gh release*` / `gh repo edit*` / `gh secret*` guards bypassable through the raw API path.

scripts/quality-check.sh
```diff
@@ -150,6 +150,43 @@ else
...
+# Replaces the hook-matrix gate deleted with the hooks (#588). verify-sandbox.sh
+# cannot fill that role -- it exits 2 unless the Bash tool runs it in a
+# sandboxed session, so it can never be a CI or pre-commit check. What IS
+# statically checkable is that the rules which stand in for the deleted hooks
+# are still present. Every entry below was a real regression at some point:
+# option-first `git stash` forms fell through to `auto` because the deny list
+# enumerated literal subcommands, and the GitHub-publishing guards were dropped
+# entirely -- effects the sandbox cannot contain, since it bounds the
+# filesystem, not the network.
+python3 - <<'PY'
```
johanzander: **This heredoc runs as a bare statement under `set -e` (line 6), so a failure aborts the whole script.** `ERRORS=$((ERRORS + 1))` never executes and the `if [ $? -ne 0 ]` block below is dead code; the scenario-discovery check, the markdown check, and the final `Errors: N / Warnings: N` summary never print.

Concretely: drop `Bash(gh release*)` from settings and the run stops mid-script with no summary — unlike every other failure mode in this file, all of which are wrapped in `if ! ...`.

```bash
if ! python3 - <<'PY'
...
PY
then
    ERRORS=$((ERRORS + 1))
fi
```

scripts/verify-sandbox.sh
```diff
@@ -80,14 +80,23 @@ check "write inside the repo is allowed" allowed "$in" \
...
+# Decide from the EXIT STATUS, not from whether the directory exists.
+# `git worktree add` creates the target directory BEFORE writing .git/worktrees,
+# and .claude/worktrees/<name> is not itself denied (only specific .claude
+# files are -- see check 4). So a correctly-blocked run still leaves the
+# directory behind, and a `[ -d ]` test reports "created" on a correct config:
+# a false FAIL, plus a stray directory when `git worktree remove` then fails.
 probe_wt=".claude/worktrees/sandbox-probe-$$"
-wt_err=$(git worktree add -q -b "sandbox-probe-$$" "$probe_wt" HEAD 2>&1 >/dev/null)
-if [ -d "$probe_wt" ]; then
+if wt_err=$(git worktree add -q -b "sandbox-probe-$$" "$probe_wt" HEAD 2>&1 >/dev/null); then
```
johanzander, line 90: **Moving the command into the `if` condition makes `wt_err` non-empty only on the branch where it is never printed.** `wt_err` captures stderr, so it is populated only when the command *fails* (`wt=blocked`) — but the check's remedy string (`... Error was: %s` with `$wt_err`) is emitted only when the check FAILS, i.e. when `wt=created`, the branch where the command succeeded and `wt_err` is empty.

So the one FAIL path of this check now always prints `Error was:` with nothing after it — losing the diagnostic exactly when the sandbox turns out looser than documented.

johanzander (commented):
Code review findings (6). Five are anchored inline; the sixth is on a line this PR doesn't touch:

**`CLAUDE.md:214` — the Permissions section contradicts this PR.** It still reads *"Only two shapes still ask ... `git push --force` and `sudo`. **That is the entire list.**"* while this PR adds 17 `ask` entries (gh merge/release/secret/workflow, shared-ref pushes, `git gc`, `git reflog expire`, `git tag -d`). That paragraph is what agents are told to trust, so the next session reasons from a list that is now wrong — and may "clean up" the new asks as unauthorized additions.

johanzander:
## Code review — 5 findings

Note on mechanism: permission precedence is **deny > ask > allow by category**, not rule order or specificity. So the issue below is not that `allow` "wins" — it's that the `ask` patterns don't *match* these command shapes at all, leaving them to the blanket allow.

### HIGH — `.claude/settings.json`: force-push guard has a pattern-shape hole

`Bash(git push --force *)` only matches `--force` immediately after `push`. Neither of these matches any `ask` pattern:

```
git push origin main --force
git push origin +beta-release-9.9
```

Both fall through to `Bash(git push *)` in `allow` and run unattended — defeating the force-push guard this PR exists to restore. `CLAUDE.md:250`'s claim that such shapes are "left to the `auto` classifier, which reads the whole command" does not hold for `git push`.

### MEDIUM — tag / shared-ref pushes uncovered

The only tag rules are `Bash(git push --tags)` / `Bash(git push --tags *)`, but no release path uses that form:

- `.claude/skills/release/SKILL.md:85` — `git push origin v<version>`
- `.claude/skills/release/SKILL.md:98` — `git push origin --delete release-X.Y`

Both publish or destroy a shared ref unattended, while the release skill requires explicit user approval for exactly those steps. `beta-release-*` pushes to `origin` are uncovered for the same reason.

### MEDIUM — no `gh api` rule

The PR body lists "non-GET `gh api`" among the restored guards, but no `gh api` entry exists in `ask`. These reach GitHub with the same effect as `gh pr merge` / `gh repo edit` while bypassing every `gh` entry:

```
gh api -X PUT  repos/johanzander/bess-manager/pulls/N/merge
gh api -X PATCH repos/johanzander/bess-manager/rulesets/17568935
```

### LOW — `scripts/quality-check.sh:178`: the gate doesn't cover its own failure mode

`REQUIRED` omits `Bash(git push --tags*)`, `gh repo edit*`, `gh secret*`, `gh workflow run*`, and `git tag -d*`. Deleting exactly those entries — the same regression class the gate was written to catch ("the GitHub-reaching guards were gone entirely") — still passes the check.

### LOW — `scripts/verify-sandbox.sh:160`: dangling symlink is misdiagnosed

Swapping `[ -e frontend/node_modules ]` for `[ ! -L ... ]` changes behaviour for a *dangling* link. `-e` follows the link, so a broken link previously produced the honest `SKIP … not present (run worktree-setup.sh)`. With `-L`, the probe runs, `touch` fails on the missing target, and the script reports `FAIL … add the main checkout to sandbox.filesystem.allowWrite` — pointing the reader at the wrong knob. Scenario: the main checkout's `frontend/node_modules` was removed or reinstalled while a worktree still links to it.

---

The three permission findings share one root shape: the `ask` patterns anchor too tightly for a rule set sitting under a blanket `Bash(git push *)` allow.

johanzander:
## Code review

### High

**1. `.claude/settings.json:59` — `git -C <path> stash pop` bypasses the entire stash deny**

Every deny entry starts with the literal `git stash`, and patterns match the command as written (the PR's own premise). The hook this replaced normalised precisely this: `.claude/hooks/auto-allow-worktree-destructive.sh:84` says *"a literal `git`+space+`push` match is sidestepped by `git -C sub push`"*, and line 269 names `git -C sub stash drop` as a hole it had to close. CLAUDE.md actively teaches `git -C .claude/worktrees/<name> …` as the cross-checkout idiom, so this is the spelling an agent is most likely to reach for.

Concrete: an agent doing cross-worktree work runs `git -C ../bess-manager-feature stash pop`, matches no deny, falls through to `Bash(git *)` in user settings, and takes another agent's stash entry — the exact scenario the deny exists to prevent. Same for `git --git-dir=… stash`.

**2. `.claude/settings.json:78` — same bypass for the newly-restored push guard**

`git -C <worktree> push origin main`, `git -C <path> push beta main --force`, and `git -c foo=bar push` all miss `Bash(git push)` / `Bash(git push *)` and hit the `Bash(git push *)` allow / `Bash(git *)` user allow instead. The PR argues the blanket spelling is required because prefix globs can't reach an arbitrary argument position — but the same argument applies to arguments *before* `push`, which the blanket rule still cannot reach. An unattended run can publish to `beta`/`main` from a worktree with no prompt.

### Medium

**3. `.claude/settings.json:89` — `Bash(git tag -d*)` misses `git tag --delete v9.9.0`**

`git tag -d` is the only tag rule; `git prune` / `git repack -d` in the same "destroys the recovery mechanism" category are uncovered too. This is an enumeration gap of exactly the type the PR body calls out as "a too-narrow ask is worse than no rule": deleting a published release tag runs unattended via the long-option spelling. Use `Bash(git tag -d*)` **plus** `Bash(git tag --delete*)`.

**4. `CLAUDE.md:454` (and the same claim at `CLAUDE.md:218`) — prose now contradicts the shipped rules**

Line 218 says `git push --force-with-lease` "runs unattended" and line 454 says it is "deliberately not in the `ask` list", while the new `Bash(git push *)` ask captures every push including `--force-with-lease`. An agent reading this section will plan an autonomous flow around a prompt that will in fact fire (implement-issue Step 4/9), and a future editor reading line 454 as normative may "fix" the settings by narrowing the push rule back — re-opening findings 1–3 of the PR body.

**5. `scripts/quality-check.sh:186` — `REQUIRED` map omits the two podman `deny` entries**

`Bash(podman machine rm)` and `Bash(podman system reset)` are missing, even though the block's own comment states *"a rule absent from here is a rule that can be silently removed."* Those are the only rules guarding a resource the sandbox cannot contain (a shared VM whose destruction is unrecoverable, and which CLAUDE.md flags as never-without-asking). A future edit dropping them passes the new gate with "✅ Permission surface intact".

### Low

**6. `.claude/settings.json:43` — `Bash(git push *)` now appears in both `allow` (line 43) and `ask` (line 79)**

Per the PR's own precedence note the `ask` wins, so the allow entry is dead — but it is the single most misleading line for anyone auditing this file, and it is the entry the PR text cites as the reason narrow asks fell through. Remove it.

### Verified non-issues

- `set -uo pipefail` (no `-e`) in `verify-sandbox.sh` makes the new `[ -n "$wt_err" ] && printf …` safe.
- The `if ! python3 - <<'PY' … PY then` construct in `quality-check.sh` is valid bash and correctly increments `ERRORS` under `set -e`.
- The three-state `node_modules` symlink branch is logically exhaustive and correct.
- CI (`issue-fix.yml`, `pr-review.yml`) runs `--permission-mode bypassPermissions`, so the new `ask` entries do not break the bot pipeline.
- All four hooks referenced in `settings.json` still exist at the PR head.

johanzander:
## Code review — 8 findings

Findings 1, 3 and 4 compound: the bypass exists, and both gates that should catch it pass.

**1. `.claude/settings.json:77` — `Bash(git -*)` downgrades every mutating `git stash` form from deny to a one-keystroke `ask`.**
Deny beats ask, but no deny rule matches `git -C <path> stash pop` or `git --git-dir=… stash drop`, while `Bash(git -*)` in `ask` does. `CLAUDE.md:231` claims "Denied outright: … every mutating `git stash` form", and `git -C <worktree>` is the exact spelling CLAUDE.md teaches as the cross-checkout idiom — so the most likely bypass is the one that got weakened. An agent following the patch-across-checkouts recipe reaches for `git -C .claude/worktrees/x stash pop`, gets a prompt instead of a hard deny, approves it, and takes another agent's entry.
Fix: add `Bash(git -* stash)` / `Bash(git -* stash *)` to `deny`.

**2. `.claude/settings.json:77` — `Bash(git -*)` also puts read-only git behind a prompt**, contradicting this PR's "runs end-to-end without approving anything" goal. It matches `git -C sub status --short`, `git --no-pager log --oneline`, `git -C .claude/worktrees/x diff -- f`, and `git -c user.email=… commit` — the last shadowing the `Bash(git commit *)` allow. An autonomous run executing rules.md's cross-checkout procedure (`git diff -- f | git -C <wt> apply`, then `git -C <wt> diff -- f` to verify) stalls twice on read-only operations.
Consider narrowing to the dangerous verbs: `Bash(git -* push*)`, `Bash(git -* gc*)`, `Bash(git -* stash*)`, `Bash(git -* reflog expire*)`, `Bash(git -* tag -d*)`.

**3. `scripts/quality-check.sh:257` — the gate conflates `deny` and `ask` (`guards = deny + ask`)**, so it certifies "Permission surface intact" for commands policy requires to be *denied*. The two `MUST_BE_GUARDED` entries at lines 217 and 219 (`git -C … stash pop`, `git --git-dir=… stash drop`) are currently satisfied *only* by the `ask` rule `Bash(git -*)`. The gate written to catch finding 1 reports green on it.
Fix: check stash/podman shapes against `deny` only, and the rest against `deny + ask`.

**4. `scripts/quality-check.sh:213` — `MUST_BE_GUARDED` omits most stash subcommands**, so nine deny rules can be deleted with the gate still green. Mutation-tested: removing `Bash(git stash clear)`, `…drop)`, `…drop *)`, `…apply)`, `…apply *)`, `…branch *)`, `…store *)`, `…save)`, `…save *)` from `deny` still prints "✅ Permission surface intact (30 command shapes guarded)" and exits 0. `git stash clear` and `git stash drop` are the two forms that irreversibly destroy other agents' entries. The file's own comment says "a rule absent from here is a rule that can be silently removed" — that is exactly what happens.
Add `git stash clear`, `git stash drop`, `git stash apply`, `git stash save x`, `git stash branch b`, `git stash store <sha>`.

**5. `.claude/settings.json:87-92` — the "destroys the recovery mechanism" category enumerates three verbs and misses equivalents.** Unmatched by any deny/ask rule: `git maintenance run --task=gc` (runs gc with prune), `git reflog delete HEAD@{0}`, `git tag -f v9.9.0 <sha>` (moves an existing tag), `git update-ref -d`. Each falls through to the user-settings `Bash(git *)` allow and runs unattended. Same enumeration-gap class this PR was written to close, and none appear in `MUST_BE_GUARDED`.

**6. `CLAUDE.md:225-231` — the "closed list" table diverges from the rules it documents.** It omits `git prune`, `git repack` and `git -*`, all of which are in the `ask` list and in the gate's `REQUIRED`; and line 231's "every mutating `git stash` form is denied" is false given finding 1. `quality-check.sh:184` says "keep this list in sync" — the doc is the copy that is out of sync.

**7. `docs/agents/rules.md:21` — stale reference to the deleted hook.** Still reads "The permission hook denies every mutating form outright", but the hooks were removed in #588 and this PR does not update it. An agent reading rules.md first (as CLAUDE.md instructs) will believe a mechanism exists that does not.

**8. `CLAUDE.md:491` — self-referential attribution error.** "…which is how findings 1–3 of PR #596 were introduced" — #596 *is* this PR; those findings were introduced by #588.

---

Checked and found correct: the `verify-sandbox.sh` worktree probe (exit-status branching, `2>&1 >/dev/null` ordering, cleanup on both paths); the three-state `-L`/`-e` node_modules logic under `set -uo pipefail`; the `if ! python3 …` wrapper protecting `set -e`; the `matches()` prefix-glob emulation; and `&&`/`;`/`|` subcommand decomposition, so `git status; git -C sub gc` is not a bypass.

johanzander:
## Code review

Verified by extracting the gate's own Python matcher from `quality-check.sh` and replaying it against spellings the gate does *not* list. The gate itself passes on this branch ("56 command shapes checked") — these are coverage gaps, not gate failures.

### 1. `.claude/settings.json:76` (high) — `Bash(git -* stash *)` denies the read-only stash commands the docs promise still work

`git -C sub stash list` and `git -C .claude/worktrees/x stash show` both match this **deny** rule (verified with the gate's matcher). `deny` has no override, so inspecting another checkout's stash — the cross-checkout idiom CLAUDE.md:345 teaches ("`git stash list`/`show` still work") — is hard-blocked. The hook this replaced explicitly allowed it (`.claude/hooks/tests/auto-allow-worktree-destructive.test.sh:191,297`).

**Fix:** deny per-verb twins (`git -* stash pop*`, `drop*`, `clear*`, …) instead of blanket `stash *`, and add `git -C sub stash list` / `stash show` to `MUST_NOT_BE_GUARDED`.

### 2. `.claude/settings.json:81` (medium) — `Bash(git -* push*)` matches any global-option git command containing " push" anywhere

The pattern compiles to `git -.* push.*`, so `git -c user.name=x commit -m "fix: push guard"` and `git -C sub log --grep push` both prompt, defeating the `Bash(git commit *)` allow. This is the exact over-broad shadowing the CLAUDE.md rationale added in this PR cites as the reason the twins name a verb — the `push*` twin reintroduces it.

**Fix:** anchor as `Bash(git -* push)` + `Bash(git -* push *)`, and add a commit-message probe to `MUST_NOT_BE_GUARDED`.

### 3. `.claude/settings.json:86` (medium) — `gh repo delete` and other unrecoverable `gh` mutations are unguarded while `gh repo edit` asks

`gh repo delete owner/repo --yes` matches no deny and no ask; with `Bash(gh *)` in user settings and `autoAllowBashIfSandboxed: true`, it runs unattended. It escapes to GitHub and git cannot undo it — this PR's own stated standard for the ask list — yet the far milder `gh repo edit` is guarded. `gh pr close`, `gh issue delete`, `gh cache delete` are in the same hole.

**Fix:** add `Bash(gh repo delete*)` at minimum, plus a `MUST_BE_GUARDED` entry.

### 4. `CLAUDE.md:346` (medium) — stale text contradicts the rule this PR ships

The stash paragraph still reads "The rules match the command as written, so a prefixed form such as `git -C <dir> stash pop` slips past — don't reach for one." That is now false: `Bash(git -* stash)` / `Bash(git -* stash *)` deny it, and `docs/agents/rules.md:22` (edited in this same PR) asserts the opposite. An agent reading CLAUDE.md concludes the prohibition is unenforced for the very spelling the PR closes.

### 5. `scripts/quality-check.sh:272` (low) — no compound-command probe

Under the PR's own "match the command as written" semantics, `cd frontend && git push origin main` and `git status && git push` match nothing in deny or ask (verified with the gate's matcher) and fall through to the `Bash(git *)` user allow. If the harness does split on `&&`/`;`, the gate should assert that rather than leave it assumed; if it doesn't, this is a live hole of exactly the class the PR closes.

**Fix:** add one compound spelling to `MUST_BE_GUARDED` and let the result settle the question.

### 6. `.claude/settings.json:100` (low) — `Bash(git -* prune*)` catches `git -C <main> worktree prune`

Same greedy-`.*` mechanism; also hits `git -C sub remote prune origin`. CLAUDE.md:495 argues worktree pruning must stay unattended ("Step 4's prune loop alone would have hit ~24 of them"). Low impact today only because `implement-issue/SKILL.md:122` happens to use the plain spelling.

---

**Reviewed and found correct:** the `verify-sandbox.sh` exit-status worktree probe (`if wt_err=$(git worktree add …)` — stderr capture is correct given `2>&1 >/dev/null` ordering, and the `[ -n "$wt_err" ] &&` short-circuit is safe under `set -uo pipefail` with no `set -e`); the three-state `node_modules` branch (covers dangling link, real dir, absent, live link); and the `if ! python3 … <<'PY' … PY then` rewrap in `quality-check.sh`, which does fix the `set -e` abort described in the PR body.

PR #599, opened 2026-08-15
fix: gate Growatt MIN TOU writes on when a change takes effect (#589)

### LINKED ISSUE #589: Gate TOU writes on when a change takes effect, not on when the segment starts
Follow-up to #554, which cut Growatt MIN TOU writes by deferring segments that have not started. Measurements taken while verifying that PR show where the remaining overhead is, and that the deferral rule can be generalised to remove most of it.

## The floor

The inverter has to be told about each distinct non-`load_first` window that actually executes: one call to program it, one to clear it. An oracle that knew the day’s final realized schedule in advance would spend exactly that and nothing more.

For the four-window test day used below, the realized schedule contains 2 windows, so the floor is **4 writes**.

## Where #554 leaves us

Full day, 96 real optimization cycles through the DP → `BatterySystemManager` → `GrowattMinController`, four separated price windows:

| write horizon | writes | floor | overhead |
|---|---|---|---|
| none (pre-#554) | 33 | 4 | +29 |
| 15 min | 12 | 4 | +8 |
| 30 min | 12 | 4 | +8 |
| 45 min (shipped) | 12 | 4 | +8 |
| 60 min | 12 | 4 | +8 |

Two things worth noting from this sweep:

- **The horizon value does not affect the write count.** Identical at 15 and 60 minutes. The horizon buys retry opportunities and bounds how many segments are eligible at once; it does not trade against writes. Choosing 45 costs nothing.
- **The residual +8 is end-boundary churn.** Eligibility is gated on when a segment *starts*. Once a window is running, the DP keeps nudging its *end* as it re-solves — `02:00-03:14`, then `03:29`, then `03:44` — and each of those is written immediately, even though the change cannot take effect for hours.

## Proposal

Gate on **when the disagreement takes effect, not when the segment starts**:

> Write only when the earliest period at which the programmed table and the plan would execute *different modes* falls within the horizon.

This strictly generalises the shipped rule. A not-yet-started segment’s first divergence is at its start, so its behaviour is unchanged. A running window whose end moves has its first divergence at the earlier of the old and new end — hours away — so it defers, and collapses to a single write when the end is finally imminent.

I believe this is optimal for a given horizon: writing earlier is only ever a bet that the plan will not change again, and writing later breaks correctness. "Act at the last responsible moment" is the boundary. The only other lever is making the plan itself more stable, which is #485 (hysteresis on near-tie flips) — independent of this and multiplicative with it.

## Open design question

Computing "first divergence" needs to know what is currently programmed. `_select_hardware_intervals` — which owns eligibility today — is reached from `_build_candidate`, which is deliberately close to pure and has no controller reference.

The practical route is to diff against the last-committed set (`active_tou_intervals`) rather than a fresh hardware read: since #552 the write path already diffs against real hardware, and since #554 `needs_hardware_reconciliation` catches the case where the two have genuinely diverged. That keeps the computation where it is. Worth confirming before implementing.

## Prerequisites already in place

This builds directly on #554 and is **inert without part of it**. In particular the `_group_periods_by_mode` walk-back: before that fix, a running segment’s `start_time` was truncated forward every cycle, which is itself a divergence — divergence-gating would detect it and write every 15 minutes forever. The slot reservation, full-plan disable diff, hardware-eligible write gate and drift reconciliation from #554 are likewise required for any deferral policy, not just the shipped one.

## Verification

The harness used for the table above is a 96-cycle in-process run with the clock walked forward (see `docs/agents/memory/` on multi-run snapshots) counting `set_inverter_time_segment` calls, plus an oracle floor derived from the realized intent sequence. Same approach should show the overhead dropping from +8 toward 0.

johanzander:
## Summary

- Gate Growatt MIN TOU writes on when a plan change first **takes effect**, not on when the segment **starts**
- **Read the real-day replay below before merging: on the one real day measured, this saves zero writes.** The mechanism is proven in isolation (15 → 1) and on a synthetic day (106 → 82), but #554 appears to have already captured the churn on real data
- `active_tou_intervals` becomes the hardware *target*: a deferred update keeps the version actually running on the inverter

## Root cause

From #589, verified against the code before changing it. Eligibility was decided by start time (`_select_hardware_intervals`, pre-change):

```python
if self._time_to_minutes(interval["start_time"]) <= horizon_minutes:
```

A running window has `start_time <= now` unconditionally, so it is eligible on every cycle forever. The DP keeps nudging its *end* as it re-solves — `02:00-03:14`, then `03:29`, then `03:44` — and each nudge changes the content key, so the differential update disables and rewrites it immediately, hours before the change can matter.

## Fix

Write when the earliest period at which the programmed table and the plan would execute **different modes** falls within `WRITE_HORIZON_MINUTES`. This strictly generalises #554 rather than replacing it: a not-yet-started segment first diverges, from an inverter that does not hold it, at its own start — so far-future deferral is unchanged by construction.

The divergence is computed over the interval's own span **plus the span of any programmed segment it overlaps**, because the disagreement is not always inside the new interval. A window whose end moved back from `06:59` to `05:59` matches what is programmed everywhere it covers; the disagreement is the hour beyond it that the inverter still holds.

Deferring an *update* leaves the previous version running, so that version is what the selection returns in its place. Two consequences, both deliberate:

- `active_tou_intervals` is the hardware target, not a subset of the plan — which is already how the disable path and `reconcile_hardware` read it.
- `planned_tou` in `sync_to_hardware` is widened to plan ∪ target, or the disable side would clear a correctly-running window mid-run to avoid rewriting it.

The dashboard's `pending_write` now asks the same `_write_is_due` helper the write path does, so a segment that is merely not due yet cannot be painted amber by one path while the other defers it.

Disabled entries are excluded from the programmed baseline: a real MIN reports unused slots as disabled ones and `initialize_from_tou_segments` seeds from that read, so counting them would carry a stale slot's mode over in place of the plan's — and since the carry-over lands back in `active_tou_intervals`, it would be its own evidence next cycle and never wash out.

## Test plan

- [x] `.venv/bin/pytest -m slow` — 538 passed, 5 skipped
- [x] **96-cycle in-process run**, real DP → `BatterySystemManager` → `GrowattMinController`, clock walked forward, counting `set_inverter_time_segment` calls. Identical harness on this branch and on its parent `160a67c8` (#554 shipped):

  | | writes |
  |---|---|
  | `160a67c8` (#554) | 106 |
  | this branch | **82** |

  Realized schedule byte-identical between runs (same 4 windows, same times) — the saving is writes, not behaviour. The cadence shows the mechanism directly: baseline rewrites the night window every cycle (00:00, 00:15, 00:30, 00:45…), this branch every second cycle (00:00, 00:30, 01:00…). Biggest single win 06:00-08:45, 9 writes → 4.

- [x] **Isolated mechanism check** — the same end-churn test run against the parent commit. 7 cycles of a running window's end flipping while the change is >2h away:

  | | writes |
  |---|---|
  | `160a67c8` (#554) | **15** |
  | this branch | **1** |

  The baseline output is the churn verbatim — `disable 02:00-05:59` / `enable 02:00-06:59` / `disable` / `enable`, seven times, for a change that cannot take effect until 06:00. This is the cleaner evidence of the mechanism: where the issue's premise (a change hours away) actually holds, 14 of 15 writes disappear.

  Two honest caveats on the full-day harness. Its mock holds SOC constant and has no InfluxDB, so the DP believes the battery never charges and walks the window end forward 15 min per cycle indefinitely; a real day converges instead. And the absolute counts are inflated relative to production. **The defensible claim is the relative −23% on an identical workload**, not the absolute 82. 100% of the residual is end-moves on windows whose end is genuinely within the horizon — correct by design.

- [x] **Real-day replay — and it shows no benefit.** `docs/bess-debug-2026-08-12-202906.md` records all 83 optimization runs of a real day, each with the strategic intent chosen for every period. Replaying that recorded sequence through the controller (no DP re-solve, so the harness can neither invent churn nor suppress it):

  | commit | TOU writes |
  |---|---|
  | `3914557f` (before #554) | 149 |
  | `160a67c8` (#554 shipped) | **32** |
  | this branch (#589) | **32** |

  The replay is faithful — pre-#554 it reproduces the 13:45 window being rewritten 17 times, matching the churn #554's own test comment documents from this same bundle. And it is demonstrably sensitive, since it separates pre-#554 from #554 by 117 writes.

  **On this day #554 already captured everything, and #589 adds nothing.** The residual after #554 is not far-future end-churn; it is changes whose divergence is genuinely inside the horizon, which are correct to write. The issue's premise — that ~+8 of deferrable end-movement remains — does not hold here.

  Caveat: this is one day. It is the only bundle I could replay, because bundles since #555/#567 use a compact delta encoding (`predicted_periods_delta`) the replay script does not decode yet.

- Reverted: the `enabled` filter in `_programmed_baseline`
- Result: `test_disabled_hardware_slots_are_not_carried_over_as_programmed` FAILED, 68 passed
- Reverted: `segment.copy()` → `segment` in the carry-over
- Result: initially **nothing failed** — the fix was unguarded. Added `test_carried_over_segment_does_not_alias_the_plan`, which then FAILED on the same mutation, 69 passed
- Both RED-first: `test_moving_the_end_of_a_running_window_issues_no_writes` and `test_a_deferred_end_change_leaves_the_running_window_programmed` failed before the fix, with the log showing `Disabling 02:00-05:59` + `Setting 02:00-06:59` at 02:15 for a change that cannot bite until 06:00
- Restored: tree clean, 70 passed

## Outcome-level coverage

`TestWritesAreGatedOnWhenAChangeTakesEffect` asserts the inverter's segment table at the moment a change must take effect, plus `set_inverter_time_segment` call counts — via `_SimulatingController`, which models the 9-slot table. No `R == P` plan-faithfulness assertion: this path has no execution-model cost to compare, and it is the test shape #554 established for the same code.

`test_carried_over_segment_does_not_alias_the_plan` asserts object identity rather than behaviour, and says so in its docstring: the corruption is only observable on a cycle that happens to assign a different slot, but the ownership property must hold on every cycle.

## Docs

`docs/SOFTWARE_DESIGN.md` rule 6 rewritten — it documented the start-based rule this replaces. `docs/agents/bess-knowledge.md` mentions nothing this change touches.

## Follow-up found while verifying (not in this PR)

`core/bess/tests/conftest.py:258` — `MockHomeAssistantController.read_inverter_time_segments()` always returns `empty_slot_table()`, never recording writes. Two effects: redundant writes are unobservable (`existing_match` can never be true), and the disable path is entirely dead (every slot reads back `enabled: False`, so the differential loop skips all nine). This derailed the first two measurement runs here, and is the second time a TOU write-count investigation has hit it.

No current test is weakened — every multi-cycle write-count test already uses `_SimulatingController`, and the single-cycle consumers are correctly modelling a fresh inverter. The exposure is prospective. Worth promoting `_SimulatingController`'s ~15 lines into conftest as a separate change.

Closes #589


johanzander:
Superseded by #601, which closes the same gap in ~30 lines instead of ~250.

Both were measured against a replay of `docs/bess-debug-2026-08-12-202906.md` (83 recorded runs): 149 writes before #554, 32 after, and 32 on **both** this branch and #601. Identical outcome, so the extra machinery here — carrying the previously-programmed version, redefining `active_tou_intervals` as the hardware target, widening the disable protection — buys nothing. It also introduced two bugs of its own in review (stale disabled slots carried over as programmed; aliased dicts mutated by slot assignment).

Closing in favour of #601. The write-count measurement problem that made this hard to evaluate is recorded in TODO.md there.

PR #601, opened 2026-08-15
fix: stop rewriting a running TOU window's end before the change matters (#589)

### LINKED ISSUE #589: Gate TOU writes on when a change takes effect, not on when the segment starts

johanzander:
## Summary

- Stop rewriting a **running** TOU window's end before the change can take effect
- 30 lines in the controller, one predicate, no change to how segments are selected
- Supersedes #599, which did the same thing at selection time for ~250 lines and an identical write count

## Root cause

#554 defers a segment until its **start** is imminent:

```python
if self._time_to_minutes(interval["start_time"]) <= horizon_minutes:
```

A window that is already running has `start_time <= now`, so it is never deferred. Every nudge the DP makes to its end as it re-solves is written the moment it appears — `02:00-05:59` becoming `02:00-06:59` at 02:15 costs a disable plus a write, four hours before the only quarter-hour that differs.

## Fix

Until the earlier of the two ends, the inverter behaves identically either way, so that write is churn. The diff now treats a programmed segment and a planned one as **already matching** when they share start and mode and differ only in an end whose change is beyond `WRITE_HORIZON_MINUTES` — one predicate, used by both the disable and update sites.

**Why here and not in `_select_hardware_intervals`.** Deferring at selection time means the previously-programmed version must be carried in `active_tou_intervals`, which turns that field from "the slice of the plan eligible to be written" into "the hardware target", and forces the disable path to be widened to spare it. That version exists and works — #599 — but it is ~250 lines against ~30, it introduced two bugs of its own in review (stale disabled slots carried over as programmed; aliased dicts mutated by slot assignment), and it produces an identical write count on every scenario tested.

## The #551 guard, resolved rather than relaxed

Hardware holds `13:45-14:14`; the plan holds `13:45-13:59`. `_ORPHAN_RANGES` counted that as a segment nobody planned, to be cleared at 07:30. It is not — same start, same mode, and identical behaviour until 14:00.

So it comes out of `_ORPHAN_RANGES` **and gains its own deadline test**: it must still be cleared before 14:00, and is, at 13:15. The set is split with a comment saying why, not quietly shortened.

## Test plan

- [x] `./scripts/quality-check.sh` passes
- [x] `.venv/bin/pytest -m slow` — 538 passed, 5 skipped
- [x] Full TOU suite — 69 passed, including every #554 deferral test unchanged

Pinned in **both** directions, because one alone would let a vacuous rule through:

- Reverted the tolerance to a strict end comparison → `test_moving_the_end_of_a_running_window_issues_no_writes` and `test_a_deferred_end_change_leaves_the_running_window_programmed` FAILED, 67 passed
- Made the tolerance unconditional (never write the end change) → `test_a_shortened_window_is_cleared_before_the_cut_takes_effect` and `test_moved_end_reaches_hardware_before_it_takes_effect` FAILED, 67 passed
- Restored: 69 passed

Isolated mechanism, 7 cycles of a running window's end flipping >2h out: **15 writes → 1**.

## What this does not do

**It does not reduce writes on real data.** `docs/bess-debug-2026-08-12-202906.md` records 83 real optimization runs. Replaying that sequence through the controller:

| commit | TOU writes |
|---|---|
| before #554 | 149 |
| #554 shipped | **32** |
| this branch | **32** |

The replay is faithful (pre-#554 it reproduces the 13:45 window being rewritten 17 times, as #554's own test comment documents from this bundle) and sensitive (it separates pre-#554 by 117 writes). On that day the residual after #554 is entirely changes taking effect at or before the current period — plan reshaping, which no horizon can defer.

So this is **correctness hardening, not an optimization**: it closes a real gap in #554's rule that did not happen to fire on the day measured. Merge it on that basis, or not at all.

## Follow-up recorded in TODO.md

Three harnesses were tried and each mismeasured this for a different reason — a mock that never records writes (`conftest.py` returns an empty slot table always, which also makes the disable path dead), a frozen SOC that invents churn, a single-snapshot fixture that suppresses it. TODO.md records all three, plus what to build: promote `_SimulatingController`'s slot modelling into conftest, make the bundle replay a checked-in script that understands the post-#555 `predicted_periods_delta` encoding, and re-measure across several real days before doing any further write-gating work.

Closes #589


bess-agent:
@claude-bot review

PR #600, opened 2026-08-15
feat: close the implement-issue review loop with the Stage 4 bot

johanzander:
Automates the loop that is currently driven by hand from a second session: open PR → `@claude-bot` review → fix findings → re-review, 2–4 rounds until only nits remain.

## Summary

- **New Step 11 in `implement-issue`** — "Independent review loop", after Step 10 (watch to green) and before the hard constraints. Requests the Stage 4 review, fixes findings, repeats. Hard cap 3 rounds.
- **New `scripts/request-pr-review.sh`** — posts `@claude-bot review` as `bess-agent` and blocks until a new review lands, printing `VERDICT <STATE> <submittedAt> <author>`. Exit 2 on a 15-minute timeout, after dumping recent `PR Review` runs.
- **`pr-review.yml` gate widened** — accepts `bess-agent` alongside the repo owner.

## Why the loop runs in the main session, not a subagent

The waiting is mechanical; the fixing is not. Triaging a finding requires the Step 2 diagnosis and the Step 3 scope assessment to tell a real bug from a reviewer objecting to a decision that was made deliberately. A fresh subagent has neither and will implement whatever the reviewer says — the exact failure `superpowers:receiving-code-review` exists to prevent.

The cost argument for backgrounding applies to *polling and narrating* through the wait, not to the fixing. So the wait is a `run_in_background` bash script — zero tokens while it blocks — and the fixing stays in the session that holds the context.

## Blast radius on the bot gate

Stages 1–3 and 5 stay owner-only: those spend money on work nobody has asked for yet. Stage 4 is the only stage that reviews a diff the requesting session already opened, and the 3-round cap bounds the spend at ~$1.50–6 per issue.

## Test plan

- `./scripts/quality-check.sh` — green (0 errors, 0 warnings).
- Workflow YAML parses and the widened `if` expression renders correctly (checked with `yaml.safe_load`).
- The verdict `jq` was run against real PRs: `#596` → `COMMENTED 2026-08-15T06:59:53Z by johanzander`; `#594` (no reviews) → empty, i.e. the loop keeps waiting rather than false-positiving.
- `bess-agent` login confirmed as literally `bess-agent` via `scripts/gh-agent.sh api user --jq .login`, not assumed.

**Not yet verified, and cannot be before merge:** `issue_comment` workflows always run the copy of the file on the default branch, so the widened gate does not take effect until this merges. Until then a `bess-agent` comment is silently ignored. See the verification comment below for the pre-merge half of the check.

## Docs

[...]

PR #603, opened 2026-08-15
chore: mark an approved PR ready for review in implement-issue Step 11

johanzander:
## Summary

- `implement-issue` Step 11 now runs `gh pr ready <n>` when the Stage 4 bot returns `APPROVED`, instead of leaving a reviewed, green PR sitting in draft.
- The merge constraint is separated out and hardened: no agent ever merges, and nothing flips out of draft before an approval.

## Root cause

Not a bug — a gap in the skill, found by running it. Step 11 (added in #600) drives the independent review loop to `APPROVED`, but Steps 10 and 12 were written before that loop existed, back when the skill handed over an unreviewed PR. So they still said *"do not take it out of draft"* and *"Draft PR only"*, and the skill stopped at "approved, green, still a draft".

That end state makes the maintainer re-derive something the loop already established: notice the PR, judge whether it is finished, flip it, then merge. Three manual steps standing in for one.

## Fix

- **Step 11, `APPROVED` branch** — runs `gh pr ready`, with the reasoning for why an approval is what earns the flip. Guarded: if commits landed after the approving review (a `TODO.md` nit, a `git merge origin/main`), re-check `gh pr checks` and `mergeable` first and report which commits landed after the review and whether they touched reviewed code. That is the one way this step could mislead.
- **Step 12** — first bullet is now "never merge, ever", stated on its own and explicitly not weakened by a green CI run or an approval. Draft-until-approval is a separate bullet, so the two rules stop being one sentence that had to be read as both.
- **Step 10** — no longer says "do not take it out of draft" flatly; says it stays a draft *there*, because nothing has reviewed it yet.
- **CI mode (row 11)** — unchanged behaviour, now explicit: CI opens a draft and leaves it, because Step 11 never runs there and there is no approval to earn the flip.
- **CLAUDE.md** — "PRs are always opened as drafts" clarified to "opened as drafts, ready on approval, never merged by an agent". It was the one line elsewhere that could be read as "must stay draft forever".
- Plus the matching Quick Reference row, one Rationalizations row, and two Red Flags (handing over an approved-but-draft PR; flipping ready early or merging at all).

## Test plan

- [x] **The change was executed by hand before being written down.** This session took #601 through Step 11: the bot returned `APPROVED`, `gh pr ready 601` flipped it, and #601 is now green, `MERGEABLE`/`CLEAN`, ready, unmerged — exactly the end state this diff codifies. The `TODO.md` nit parked on #601 after its approval is also the concrete case the "commits landed after the review" guard is written for.

Not applicable, and not skipped silently: this diff is prose in `.claude/skills/` and `CLAUDE.md` with no executable behaviour, so there is no mutation that could redden a test. The equivalent evidence is the manual execution above — the flow was run against a real PR first, and the wording follows what that run actually needed (including the post-approval-commit guard, which only surfaced because it happened).

## Outcome-level coverage

None, because no optimizer behaviour changes — no DP, intent, control-mapping or economics code is touched. The affected files are `.claude/skills/implement-issue/SKILL.md` and `CLAUDE.md`.

## Documentation check

Grepped `docs/agents/bess-knowledge.md` and `docs/SOFTWARE_DESIGN.md` — neither mentions PR draft state or the review loop, so nothing to update there. `docs/agents/workflow.md` and `docs/agents/rules.md` mention drafts only in the CI/Stage-3 sense, which this change deliberately leaves alone. `CHANGELOG.md` is intentionally not touched: agent tooling has no user-visible effect, matching #600, which changed the same skill and added no entry.

### DIFF
--- a/.claude/skills/implement-issue/SKILL.md
+++ b/.claude/skills/implement-issue/SKILL.md
@@ -500,12 +500,25 @@ On the verdict:
   never merge it, and never flip it early — an unreviewed or red PR stays a
   draft, no matter how confident you are in the diff.
 
-  If you push anything after the approval (parking a nit in `TODO.md`, a
-  `git merge origin/main`), re-check `gh pr checks` and
-  `mergeable`/`mergeStateStatus` before flipping, and say in your report
-  which commits landed after the approving review and whether they touched
-  the reviewed code. `gh pr ready` on a PR whose approved diff has since
-  changed underneath it is the one way this step can mislead.
+  **If you push anything after the approval, the approval covers a diff that
+  no longer exists, and what you do next depends on what you pushed.** Ask
+  whether the new commits touched code the reviewer actually reviewed:
+
+  - **They didn't** (parking a nit in `TODO.md`, a `CHANGELOG.md` line, a
+    clean `git merge origin/main` that changed nothing under review):
+    re-check `gh pr checks` and `mergeable`/`mergeStateStatus`, then flip.
+    Name those commits in your report and say why they fall outside the
+    reviewed diff, so the maintainer can check that call rather than take
+    it on trust.
+  - **They did:** do NOT flip. Go back and run another review round — the
+    approval you are holding was for different code, and `gh pr ready` on a
+    stale-approved diff is the one way this step can actively mislead. A
+    round costs minutes; a wrongly-ready PR spends the maintainer's trust in
+    the ready flag, which is the only thing that makes it worth setting.
+
+  The round cap counts this round like any other. If it lands you on the cap,
+  say so and hand over a draft PR with the outstanding state — an honest
+  draft beats a ready flag that means less than it claims.
 - **`CHANGES_REQUESTED` / `COMMENTED`** — collect the findings:
 
   ```bash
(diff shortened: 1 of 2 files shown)


bess-agent:
@claude-bot review

bess-agent:
Findings 1 and 3 were right and are fixed in 92e51364.

**1 — CLAUDE.md / CI-mode contradiction.** Confirmed by reading it back: the bullet sits under "General bot rules", which governs Stages 1–5, so it claimed ready-on-approval for exactly the flow this PR says skips Step 11. A Stage 3 PR approved by a manual Stage 4 review had nothing that would ever flip it. Now split in two — interactive runs flip their own; a CI-opened PR stays a draft for the owner, who is triggering that review by hand anyway.

**3 — post-approval guard.** Agreed, and the sharper way to put it is that disclosure isn't mitigation. The guard now branches on the answer instead of just reporting it: if the post-approval commits fall outside the reviewed diff, flip and justify which commits and why; if they touched reviewed code, don't flip, run another round, because the approval being held is for different code. The round cap counts it, and hitting the cap hands over an honest draft.

**2 — `gh pr ready` left unattended: deliberate, and now documented.** Not adding `Bash(gh pr ready*)` to `ask`, for the reason CLAUDE.md's own standard gives: *"the effect escapes the repo **and** git cannot undo it."* Both halves have to hold. `gh pr ready` clears the first and fails the second — `gh pr ready --undo` puts it straight back, it changes no content, and the diff was already public because the `git push` that created it prompted. Compare the row above it: `gh pr merge` is the same category and *is* gated, because nothing undoes a merge to `main`. That is the line, and `gh pr ready` sits on the safe side of it.

The other half is that gating it would defeat the change. This flip is the codified endpoint of Step 11's loop, which exists to reach that state without the maintainer in the loop; a prompt there stalls the one flow whose whole point is not stalling. You were right that it shouldn't be silent, though, so the permissions section now states the call and its reasoning rather than leaving the absence to read as an oversight.

Worth adding: the guard from finding 3 has already been exercised for real. This session flipped #601 after parking a `TODO.md` nit *following* its approval — outside the reviewed diff, hence legitimate under the rule as now written, which is also where the guard came from.

bess-agent:
@claude-bot review

Issue #602, opened 2026-08-15
bug: terminal value is quantity-linear (bang-bang midnight SOE) and double-charges cycle_cost


johanzander:
## Summary

Two defects in end-of-horizon valuation, found while investigating #595 and verified against that bundle's real inputs:

1. **The terminal row is linear and unbounded in quantity**, so the DP's midnight SOE is a *step function* of `terminal_value_per_kwh` — it flips from "carry nothing" to "carry a full battery" across a few öre. No scalar produces a moderate carry.
2. **`cycle_cost_per_kwh` is charged twice** on energy held to the boundary — once on the way in, again in the terminal formula.

They interact: (2) currently parks the slope on the "carry nothing" side of (1)'s cliff. **Fixing (2) alone turns a 2.4 kWh undershoot into a 10 kWh overshoot.** They must be fixed together.

Repro script: https://gist.github.com/johanzander/2002f14f6281cc7236df5612b4a24e89 (run against the debug bundle attached to #595).

## Defect 1 — quantity-linear terminal row

`core/bess/dp_battery_algorithm.py:1215-1218`:

```python
if terminal_value_per_kwh > 0.0:
    for i, soe in enumerate(soe_levels):
        usable_energy = soe - battery_settings.min_soe_kwh
        V[horizon, i] = max(0.0, usable_energy) * terminal_value_per_kwh
```

One constant slope, no quantity limit. The DP compares that slope against the shadow price, so the decision is all-or-nothing.

Sweep on #595's real inputs (20 kWh battery, `min_soe` 2.0, `cycle_cost` 0.4, `initial_soe` 2.40, 62-period horizon from period 34):

| `terminal_value_per_kwh` | boundary SOE (kWh) |
|---|---|
| 0.000 – 0.650 | 2.02 (flat across the whole range) |
| 0.690 | 2.26 |
| **0.720** | **5.29** |
| 0.786 | 14.37 |
| 0.850 | 18.22 |

The entire transition happens between 0.69 and 0.79. Worse, at the top of that band the parameter is knife-edge against the SOE grid discretization — deterministic across repeated runs, but:

```
terminal 0.786087 -> boundary SOE 14.37
terminal 0.786100 -> boundary SOE 16.32
```

A **1.3×10⁻⁵** change in the terminal value moves midnight SOE by 2 kWh. This parameter is not tunable.

[...]

Issue #595, opened 2026-08-15
Safe speculations to not miss future needs


ridax67:
[bess-debug-2026-08-15-084345.md](https://github.com/user-attachments/files/31095798/bess-debug-2026-08-15-084345.md)

<img width="1104" height="1361" alt="Image" src="https://github.com/user-attachments/assets/f5228220-ac91-417b-b8cf-be7cbf8f71c9" />

<img width="1200" height="1924" alt="Image" src="https://github.com/user-attachments/assets/39eddaf1-cfe9-4f96-ba72-740b7236bd3d" />

## What happened
Bess is leaving cheap spots to export because it is not seeing beyond midnight befor it gets tomorrows pricing.  Would it not be better to "fill in" certain slots with safe speculations?  Like hour 0-6 wil be little sunshine and the house power need we can estimate already.  The only thing we don't know is the pricing for those hours.  The price could be assumed to the same as 23.00-23.59.

As it is now it often plans for soc min at 00.00 just because it doesn't know there is a tomorrow.  I know (or hope) there will be a tomorrow.  I don't know the pricing, but I know it will not justify buying from grid instead of charging with solar now in the morning.

## Debug bundle
Please drag-and-drop **bess-debug-2026-08-15-084345.md** from your Downloads folder into the editor below before submitting.
The file contains your version, settings, recent logs, and the raw HA discovery dump that the maintainers need to reproduce the issue.

Issue #345, opened 2026-07-18
Terminal value unconditionally zeroed at extended (today+tomorrow) horizon boundary — may contribute to #126's repeated 'horizon drift'


johanzander:
## Summary

`_calculate_terminal_value()` (`core/bess/battery_system_manager.py:1840-1915`) only estimates a value for battery energy left at the end of the DP's optimization horizon when the horizon is **today-only**. As soon as the horizon extends into tomorrow (tomorrow's prices become available, typically early-mid afternoon), it unconditionally returns `0.0`:

```python
# core/bess/battery_system_manager.py:1877-1884
today_period_count = get_period_count(time_utils.today())
remaining_today = today_period_count - optimization_period
total_horizon = len(buy_prices)

# If horizon extends past today, DP has explicit tomorrow data
if total_horizon > remaining_today:
    logger.info(
        "Horizon extends past today (%d > %d remaining), terminal value = 0.0",
        total_horizon,
        remaining_today,
    )
    return 0.0
```

The comment's premise ("DP has explicit tomorrow data") is only true *up to* the new horizon boundary (midnight tomorrow) — it does not extend past it. The day-after-tomorrow's prices are essentially never available (no provider publishes them this far ahead), so the DP has exactly as little visibility past midnight-tomorrow as it had past midnight-today before the multi-day horizon feature existed.

This means the exact failure mode that motivated adding terminal value in the first place — commit `a3435fff` ("Extend DP optimization horizon with tomorrow's prices and terminal value fallback"): *"The DP algorithm treated energy in the battery at end-of-horizon as worthless, causing it to dump battery energy late in the day when holding for tomorrow would be more economical"* — is structurally reintroduced at the new (today+tomorrow) horizon edge, just unconditionally suppressed instead of estimated.

This isn't a rare edge case: it is the normal state every single day, for the several hours between when tomorrow's prices land (~13:00-15:00 typically) and midnight.

## Why this matters given #126

[...]

PR #605, opened 2026-08-15
test: retrofit the scenario corpus onto production terminal values

johanzander:
## Summary

- Retrofits all 38 pinned scenario fixtures off `terminal_value_per_kwh = 0.0` onto production-computed terminal values, closing the blindness `TODO.md` records as a **blocking prerequisite for #602**.
- Extracts the terminal-value formula, which existed twice, into `core/bess/terminal_value.py` — verified bit-identical to the previous implementation.
- Adds two guards so the corpus cannot go blind again, and two re-pin scripts so the artefacts it now owns are regenerable.

This is groundwork, not the #602 fix. It deliberately changes no production behaviour.

## Root cause

`optimize_battery_schedule()` defaults `terminal_value_per_kwh=0.0`, and at 0.0 the terminal-row branch never executes at all — `V[horizon]` stays all zeros (`dp_battery_algorithm.py:1216-1219`). Exactly one of 38 fixtures overrode that default. So the pinned corpus could neither regress the terminal value nor validate it, in either direction; that is how #345's terminal-value fix went green without a single fixture reaching the code path.

## Fix

**The formula is now shared.** It existed in `BatterySystemManager._calculate_terminal_value` and again, hand-cloned including the #359 carve-out, in `simulation/verification.py:108-121`. The test path needed it a third time. Sharing it is load-bearing rather than tidiness: the forecast-robustness harness was free to judge plans by a different objective than the one that produced them.

**#422's cap scoping is reproduced without timestamps** as the last `24/period_duration` periods. That yields `regression_frank_debug_before`'s independently-pinned `0.143013413` exactly, where the unscoped array gives the pre-#422 `0.195488259`. Without it the retrofit would have quietly reintroduced the bug #422 fixed into all 38 fixtures.

**`capture_vpp_baseline.py` gains `--repin-current`** because it had no legal re-pin path: `--add-new` is a no-op on an unchanged fixture set, leaving a full re-baseline — which the script's own docstring warns destroys the v10.0.2 drift signal — as the only apparent remedy. The v10.0.2 half is byte-identical here; 22 fixtures moved on the current half only, verified programmatically.

Two mutations, both run.

**1. The corpus now detects a terminal-value formula change.**

- Mutated: dropped the `- cycle_cost_per_kwh` term from both `buy_based` and `sell_cap` in `terminal_value.py` — #602's actual candidate change.
- Result: `test_recorded_terminal_values_still_match_the_production_formula` FAILED (1 failed, 43 passed).
- Then ran the re-capture it points to: all 38 recorded values moved, and `test_all_scenarios` went to **24 failed, 20 passed**.
- Restored: tree clean.

**The first mutation is the reason the second guard exists.** Before it, that same mutation left `test_scenarios.py` at **43 passed** — the fixtures kept replaying their recorded constants and nothing noticed. Explicit values bought transparency at the cost of formula-change detection; the guard buys it back. This was caught by running the mutation, not by reading the diff.

**2. The declaration guard is not vacuous.**

- Mutated: removed `terminal_value_per_kwh` from `synthetic_seasonal_spring.json`.
- Result: `test_every_fixture_declares_a_terminal_value` FAILED.
- Restored: tree clean.

## Outcome-level coverage

- `expected_results` on all 38 fixtures — re-pinned on the 19 that moved, each delta printed and reviewed.
- The action-selector goldens (23 regenerated) pin `actions`, `intents`, `intra_period_discharge_allowed` and the SoE trajectory bit-exactly.
- The Growatt VPP baseline's current half (22 regenerated), pinning commands, realized cost and SoE trajectory.
- R == P is intact: `test_plan_faithfulness.py::test_realized_matches_planned_across_all_fixtures` runs `run_scenario_realized` over the whole corpus and passes at the retrofitted values.

## Measured deltas

Largest economic movements, all explained:

| fixture | `battery_solar_cost` | note |
|---|---|---|
| `realworld_2026_04_29_195900` | 43.57 → 47.59 | see below |
| `synthetic_2025_01_12_evening_peak_with_solar` | −9.92 SEK realized (VPP) | improvement |
| everything else | < 0.5 SEK | tie-breaking at the horizon |

**`realworld_2026_04_29_195900` looks like a 4 SEK regression and is not.** In-horizon cost rises 4.014 SEK, terminal credit on the carried energy is 4.075, so the true objective **improves by 0.061 SEK**. `battery_solar_cost` excludes the terminal credit, so a correct decision reads as lower savings. Worth knowing before #602 makes carries larger — the same mechanism will surface as user-visible "savings dropped" reports with no bug behind them.

## A finding this surfaced (logged, not fixed)

Under production-realistic terminal values, only **four** of 38 fixtures still flag a tie window, and the largest hybrid advantage among them is **0.0043 SEK** — against the +0.0600 SEK `test_hybrid_resolution_improves_on_grid_dp` was written on. A nonzero terminal row adds a horizon value gradient that dissolves the near-ties #450's PWL splice resolves.

Three test files now pin `terminal_value_per_kwh = 0.0` explicitly so they keep exercising the mechanism they were written for. That is honest — a near-tie rig must run where near-ties exist — but it means the hybrid path's *production* value is now measured by nothing. Logged to `TODO.md`; it belongs to #450, and the question changes again once #602 lands.

I did **not** lower the 0.01 threshold to chase the surviving 0.0043 fixture. That would have kept the test green while weakening it tenfold.

## A determinism defect this surfaced (#606, carved out)

`regression_2026_08_13_145213` selects between **two near-equal-value plans by floating-point details**, so the choice changes with the interpreter: period 18 is `-1.1625 kW` on Python 3.12.13 and `-0.6875 kW` on 3.13.15 — same commit, same pinned numpy, deterministic within each environment. That is a **P2 violation**: a tie resolved outside `tie_policy.py`. Filed as **#606**.

The retrofit surfaced it rather than caused it: the fixture only diverges once it runs at its production terminal value instead of the `0.0` default, and it is the corpus's most terminal-value-sensitive fixture (8.41 kWh of boundary SOE under #602's candidate change). `origin/main`'s golden for it reproduces *exactly* on 3.12, so the goldens were not already environment-fragile.

Carved out narrowly: only the plan-shaped arrays and the 1e-9 cost pin, in the parity gate and the VPP current-plan pin. **The economics stay covered** — `test_all_scenarios` asserts `battery_solar_cost` at 0.001 SEK and passes on both interpreters, which is itself the evidence the two plans are a tie rather than one being wrong. What is uncovered until #606 closes is the plan, not the money.

Also logged in #606: `.python-version` pins 3.12.13 while CI runs 3.13.15, so a divergence of this kind cannot be seen locally.

## Test plan

- [x] `pytest -m slow`: 543 passed, 7 skipped
- [x] Fast suite: 1943 passed, 50 skipped
- [x] Extraction A/B'd against `origin/main`'s implementation over 76 real price-array pairs from every fixture: **max delta 0.000e+00, bit-identical**
- [x] Ran the real `BatterySystemManager._calculate_terminal_value` on a bundle-derived fixture and observed its actual log line: `Terminal value: 0.143/kWh (buy_based=0.310, sell_cap=0.143, cap_applied=True, ...)`, returning `0.143013413` — matching #422's independent pin, with the cap binding as intended. Confirms the breakdown-dict delegation and logging work at runtime, not just under test.
- [x] Verified the VPP baseline's v10.0.2 half is unchanged on every fixture
- [x] Both capture scripts verified idempotent (`--check` clean on a second run)

Mock-HA stack was brought up (`docker-compose.ci.yml`) but its scheduler needs the setup wizard before it optimizes; the in-process production-method run above is the substituted observation.

## Documentation

- `docs/agents/bess-knowledge.md` — the terminal-value section pointed at a stale `battery_system_manager.py:1841-1921`; now points at the shared module and notes all three callers share one objective. Formula description unchanged, because the formula is unchanged.
- `docs/agents/testing.md` — records that `terminal_value_per_kwh` is now required on every fixture, and the three-artefact regeneration order.
- `docs/SOFTWARE_DESIGN.md` — checked, its one mention is generic and still accurate.

## CHANGELOG

Deliberately none: no user-visible behaviour changes. Per the repo convention that pure internal work stays out of the curated sections.

## Scope assessment

**Structural**, per `rules.md` Debugging Protocol step 8. The formula moves modules and `BatterySystemManager` keeps ownership of fetching its inputs and reporting — the DP must not learn to fetch forecasts. No new class; `terminal_value_breakdown` returns a plain dict.

**Workaround check.** The diff adds no parameter, flag, fallback or trigger to route around an ordering or dependency problem. The two judgement calls are stated rather than hidden: the empty-input case is owned by `BatterySystemManager` and **raises** in the shared function (the extraction had silently given `verification.py` a `0.0` fallback it never had — caught in review, fixed, contra `rules.md` no-silent-fallbacks); and the day-boundary precondition on cap scoping is documented as unenforceable from fixture data rather than guarded by a length check that would be wrong (`n % periods_per_day == 0` is false for every correct multi-day fixture here).

Related: #602 (this unblocks it), #595, #422, #450, #345.


bess-agent:
@claude-bot review

johanzander:
## Code review

Ran the suites against the PR tree on Python 3.12.13: fast suite 1506 passed / 50 skipped, slow suite 542 passed / 8 skipped. Nothing red. The findings below are correctness/coverage gaps, not failures.

### Medium

**1. `docs/agents/testing.md:283` — doc contradicts the capture script**

The doc says a fixture that must pin a specific terminal value "keeps its recorded value; the script leaves any already-recorded value alone." `scripts/capture_scenario_terminal_values.py:65` does the opposite: it overwrites any recorded value that differs from the computed one by ≥1e-9. And the new `test_recorded_terminal_values_still_match_the_production_formula` makes a hand-pinned value fail outright — its own docstring says such a fixture "does not belong in this corpus".

Failure path: a maintainer follows the doc, hand-pins a terminal value to reproduce an upstream-computation defect, then either runs the documented capture script (pin silently overwritten) or pushes (slow suite red in CI, no supported remedy). The doc paragraph should say the recorded value is *derived*, not preserved.

**2. `core/bess/tests/unit/test_vpp_regression_baseline.py:112` — `#606` skip is wider than intended**

The skip is placed before all four assertions, but only the first (`capture_plan(name) == entry["current_plan"]`) depends on interpreter-dependent plan selection. The remaining three replay the *recorded* plan through the VPP simulator and are fully deterministic — verified on this fixture: `commands` match exactly and the `realized_cost` delta is `0.0`.

So the carve-out also drops VPP execution-model coverage (commands, realized cost, SoE trajectory) for `regression_2026_08_13_145213`, contrary to the PR's stated "only the plan-shaped arrays and the 1e-9 cost pin" scoping. A simulator regression on this fixture would go unnoticed. Move the skip to guard only the `capture_plan` assertion.

### Low

**3. `scripts/capture_vpp_baseline.py:125` — `--repin-current` misses simulator-side drift**

It writes an entry only when `capture_plan(name) != entry["current_plan"]`. If a change moves the *simulator* rather than the planner, every plan compares equal, the script prints "no fixture's plan changed; nothing written" and exits 0 — while `test_current_plan_is_pinned`'s `commands`/`realized_cost`/`soe_trajectory` assertions stay red. That leaves a full re-baseline (the thing the docstring warns against) as the only apparent remedy, which is exactly the gap `--repin-current` was added to close. Re-simulating and comparing `current` as well as `current_plan` closes it.

**4. `scripts/capture_vpp_baseline.py:115` — `--repin-current` repins the `#606` fixture**

It iterates all fixtures including `regression_2026_08_13_145213`, writing that fixture's interpreter-specific plan into the tracked baseline even though no test reads it (see finding 2). The artefact then silently records whichever interpreter last ran the script. Worth skipping the `#606` set here, or at least noting it.

**5. `core/bess/tests/unit/test_action_selector_parity.py:83` — exempt fixture passes vacuously**

For the exempt fixture the test body reduces to a full DP solve (`capture_fixture(name)`) followed by no assertion at all, and reports as PASSED. `pytest.skip` (as the VPP test uses) would make the missing coverage visible in the run report and avoid the wasted solve; as written, a future assertion added outside the two `if` guards would silently re-arm the flake.

**6. `scripts/capture_scenario_terminal_values.py:81` — cosmetic**

In `--check` mode with a clean corpus the script prints "0 fixture(s) updated", contradicting `--check`'s documented "without writing". The stale branch exits before this line, so only the clean case is misleading.

### Checked and found sound

- The extracted `terminal_value.py` is behaviourally identical to both prior copies — same floors, same `export_prices_vary` cap gate, same empty-input raise semantics, with the `not buy_prices` short-circuit retained only in `BatterySystemManager`.
- The `#422` cap-scoping approximation in `helpers.scenario_terminal_value` holds across the corpus: all 38 fixtures enumerated, every multi-day one is "partial first day + whole terminal day" as the docstring claims. The one horizon that does not end on a day boundary (`realworld_2026_04_19_084608`) is shorter than a day, so the slice degenerates to the whole array and stays correct.
- The capture script's tolerances mirror `test_all_scenarios`' `COST_TOLERANCE_SEK` / `PCT_TOLERANCE` / `ENERGY_TOLERANCE_KWH` exactly.
- The `#571` shadow-price tests that consume the retrofitted fixture still discriminate — `test_shadow_price_still_varies_between_grid_cells` passes non-vacuously.
- CI runs `pytest -m slow`, so both new guards do execute in the PR gate.

PR #607, opened 2026-08-16
fix: offer exact load cover wherever the lattice cannot express it (#352 Shape B, Phase 4b)

### LINKED ISSUE #352: bug: low-rate BATTERY_EXPORT writes grid_first at a forced sub-load discharge rate, causing avoidable grid import on any load spike

johanzander:
## Summary

- A period whose house load falls between two discharge steps had **no action that covered it**. The step below under-covers and imports the difference at the buy price; every step above is either excluded as unexecutable (#497) or exports for real at the sell price. With buy > sell that is a forced loss both ways, and the DP took the cheaper one — the export.
- That export is what classified the period `BATTERY_EXPORT`, which is what wrote `grid_first` at a sub-load rate. `grid_first` does not load-follow, so the rate is a hard commitment and any load spike is imported while the battery sits full — @pookey's original report.
- Phase 4b of the optimizer plan; closes #352 **Shape B**. (Shape A was #520/#524.)

## Root cause

From the design doc §2, which is normative here:

> At the field-evidenced period the action space offers **2.70 kW** or **3.30 kW**, and nothing between: the house needs **2.80 kW**; 2.80 is off the percent lattice (15 kW battery, 1% = 0.15 kW); 2.85 / 3.00 / 3.15 are removed by #497's unexecutable band; `_residual_cover_p` offers exact cover **only below the smallest lattice step**, so it does not fire here.
>
> **The export was never the goal.** It is the least-bad option, not a choice.

`_residual_cover_p` already planned exact cover — as a *delivery* under a ceiling command, which load-following hardware executes as `min(ceiling, actual load)` — but only below the smallest lattice step, where it read as a narrow #466 sunrise special case. It is the same gap at any size.

## Fix

1. **`action_selector._residual_cover_p`** — the below-the-lattice restriction is removed; exact cover is offered wherever the lattice cannot express covering the load. The gates that remain are executability ones: the classifier's noise floor and magnitude threshold, and "a covering ceiling exists on this platform".
2. **`execution_model.discharge_command_index`** — one planned-power → written-percent conversion, rounding by what the number *means*: **up** for a ceiling (`load_first` on load-following hardware, where rounding down under-delivers the plan — 539 of 1377 corpus deficits round down), **nearest** for a target (`grid_first` anywhere, native SolaX). `inverter_controller` and the simulator's mirror both call it, so they cannot round apart — #282's and #497's shared failure shape.
3. Everything else follows: both backward passes already consume `_residual_cover_p`, so the DP and the PWL window stay on one action set (P1).

Chosen over the off-lattice alternative per the design doc: the plan *is* the lattice command, so the lossy conversion that would need a second fix does not exist.

## Measured

38-fixture corpus, DP re-optimising, measured on this branch against its own merge base.

**Re-measured 2026-08-16 after #605 and #608 landed.** The first version of this PR was measured at `terminal_value_per_kwh = 0.0`; #605 retrofitted the corpus onto production terminal values and #608 removed the interpreter-dependent plan selection that exposed. Every number below is re-derived on the merged base — the earlier ones (−3.33 SEK, 174 → 129) described a corpus that no longer exists.

| | pre | post |
|---|---|---|
| corpus cost | 1795.37 SEK | **1791.09 SEK** (−4.27) |
| partial-rate committing exports | 160 | **118** |
| fixtures moved | — | 30 / 38 |

The fix reads *better* at production terminal values than at zero: −4.27 SEK against −3.33, and 42 committing exports removed against 33.

**Three fixtures report a higher cost** (+0.18 total, largest +0.1023 on `realworld_2026_04_27_184643`) — down from five before the retrofit. Stated rather than smoothed over: this is not the objective-vs-reported pricing split, since the objective moved by the same amount. On the largest, the new plan ends the horizon with more energy stored, which the reported scalar does not price but the DP's own objective does, so those are not like-for-like comparisons. It remains the one claim here I have bounded rather than fully closed (P6).

## Deliberate re-pins

First phase where golden churn is expected (design §6). Re-pinned with **main's** tooling, not this branch's:

- **20 fixtures' `expected_results`** — via `scripts/capture_scenario_expected_results.py`.
- **Action-selector goldens** — `actions`, `intents`, `intra_period_discharge_allowed`, SoE.
- **VPP baseline, `current` half only** — via `capture_vpp_baseline.py --repin-current`; the v10.0.2 reference half is untouched.

Note for the reviewer: #605 independently added *both* of those tools while this branch was open — including a `--repin-current` flag with the same name and semantics as the one this PR originally introduced. Main's versions are kept and this branch's duplicates dropped.

## Test plan

- [x] `pytest -m slow` green — 546 passed, 7 skipped (re-run on the merged base)
- [x] **Step 8, real stack** — `docker-compose.ci.yml` with mock-HA, the reproduction bundle optimized by the production DP and pushed through the real `GrowattMinController` + HA API controller, read back from mock-HA's service log:

| | intent | mode | rate | commanded | plan | house |
|---|---|---|---|---|---|---|
| pre-fix | `BATTERY_EXPORT` | **`grid_first`** | 22% | 3.30 kW (committed) | 3.30 | 2.80 |
| post-fix | `LOAD_SUPPORT` | **`load_first`** | 19% | 2.85 kW (ceiling) | 2.80 | 2.80 |

  Re-confirmed on the merged base: p99 still plans `LOAD_SUPPORT` at exactly −0.7000 kWh, zero export, zero import.

mock-HA recorded `number.growatt_battery_discharging_power_rate = 19` (was 22). Pre-fix exported 0.125 kWh and committed the inverter to a fixed 3.30 kW — the configuration where the reporter's oven load was imported.

Two mutations, both run against the final code:

- **Reverted:** the below-the-lattice restriction in `_residual_cover_p` (the pre-4b gate).
  **Result:** `test_352_low_rate_export_is_not_planned_when_exact_cover_exists` FAILED — `period 99: planned a 0.1250 kWh export against 0.7000 kWh of home consumption`. 1 failed, 1 passed.
- **Reverted:** the ceiling rounding direction in `discharge_command_index` (`ceil` → `round`).
  **Result:** `test_exact_cover_is_delivered_in_the_round_down_band` FAILED — the cover candidate is withdrawn (no covering ceiling) and the DP falls back to `[('BATTERY_EXPORT', 1.40), ('LOAD_SUPPORT', 1.20), ('BATTERY_EXPORT', 1.40)]`, the #352 mechanism reproduced in miniature. **Re-run after the merge**, where that test was relaxed from "all three periods cover" to "at least one does" (how many cover is an incidental DP outcome the terminal value moves); the mutation still empties the covering set, so the relaxation did not hollow it out.
  Worth recording: this mutation is **not** caught by the corpus R==P test, which is why that second test exists — the first draft of it passed under the mutation because its period-selection criterion matched export by-products.
- **Restored:** tree clean, both suites green.

## Outcome-level coverage

- `test_352_low_rate_export_is_not_planned_when_exact_cover_exists` — flows at p99 on the real bundle, plus whole-horizon R==P through `simulate(...)`.
- `test_exact_cover_is_delivered_in_the_round_down_band` — realized-vs-planned discharge for a deficit in the round-DOWN band, the half the reproduction does not reach.
- `test_352_fix_must_not_refuse_the_arbitrage_remainder` — the adversarial pin the design requires (25 kWh, 10 kW cap, 2/4/5 SEK): the partial-rate remainder must stay in the cheapest hour. Green before and after; adding a candidate cannot refuse a plan.
- `expected_results` on 20 fixtures, the action-selector goldens, the VPP `current` half, and `KNOWN_PLAN_EXECUTION_GAP_SEK` (unchanged, one entry).

## Review findings addressed

From `/code-review high`, all 10 confirmed findings:

1. Stale `grid_to_battery_solar_savings` pin — fixed (the slow-suite failure).
2. **Ceiling-up over-delivers when the #520 gate is closed** — real, and the docstring overclaimed. `ceil` is the *tightest admissible* ceiling, not a free one: no rounding both delivers the plan and never exceeds it, since the lattice is coarser than the deficit. Measured and documented (436/756 LOAD_SUPPORT periods get a ceiling above plan, 31 with the gate closed, each ≤ one step); nearest-rounding leaked up to half a step in the same periods *and* under-delivered in the others. Not "fixed" by making the direction gate-dependent — that is a second construction site for the same conversion.
3. The two capability flags can diverge (solax-modbus VPP declares cover=True with a target rate), safe today only via a third file — now pinned by `test_exact_cover_is_only_declared_where_the_written_rate_can_deliver_it`, with an anti-vacuity counter.
4. Write path hardcoded `max/100` — now reads `discharge_resolution_kw()`.
5. Four stale comments asserting the deleted invariant — corrected.
6. `min_discharge_gear_kw` became dead — removed.
7. Duplicate candidate — deduped (measured cost-neutral, see above).
8. Half-to-even on the target branch — documented as deliberate; biasing it would silently shift every target platform.
9. Four spellings of the lattice epsilon — consolidated on `execution_model.LATTICE_EPS` (P5).
10. Hand-written nearest-rounding — discharge sites migrated; the charge sites are 4c's scope and say so.

## Docs

`docs/agents/bess-knowledge.md`'s action-space section described the below-lattice restriction and pointed at `dp_battery_algorithm.py` for a function Phase 1 moved. Updated. `docs/SOFTWARE_DESIGN.md` mentions nothing this diff touches.

Refs #352 — not closing it from this PR, per repo policy: the reporter's issue closes on the prod release.

https://claude.ai/code/session_01GtTbxMVD8mqib3Rh1Sc34Q


bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

PR #608, opened 2026-08-16
fix: resolve exactly-tied DP plans by policy, not float noise (#606)

### LINKED ISSUE #606: bug: DP plan selection is interpreter-dependent at realistic terminal values (P2 tie not routed through tie_policy)
## Summary

Under a production-realistic terminal value, `regression_2026_08_13_145213` produces **two materially different plans depending on the Python interpreter version**, with everything else held fixed. This is a P2 violation (`docs/agents/optimizer-architecture.md`): a near-tie is being resolved by floating-point details rather than through `tie_policy.py`'s lexicographic preference table.

Found while retrofitting the pinned fixture corpus off `terminal_value_per_kwh = 0.0` (#605, the blocking prerequisite for #602).

## Evidence

Same commit, same fixture, same pinned numpy (2.4.6, `requirements.txt:25`):

| environment | period 18 action |
|---|---|
| local, Python 3.12.13 | **-1.1625 kW** |
| CI, Python 3.13.15 | **-0.6875 kW** |

A 0.475 kW difference — a different decision, not a float-representation artifact. It is **deterministic within each environment** (three identical local runs), so this is two stable plans, not flakiness.

It fails two bit-exact pins on CI:

- `test_action_selector_parity.py::test_selector_refactor_is_bit_identical[regression_2026_08_13_145213]`
- `test_vpp_regression_baseline.py::test_current_plan_is_pinned[regression_2026_08_13_145213]`

## Why this is new, and why it is P2

**It is new.** `origin/main`'s golden for this fixture reproduces *exactly* on Python 3.12 — `actions`, `intents`, `intra_period_discharge_allowed` and `soe_trajectory` all match. So the goldens were not already environment-fragile. The divergence appears only once the fixture runs at its production terminal value (0.202886574) instead of the old 0.0 default.

**It is P2.** Two plans of near-equal value exist and the selection between them is decided by interpreter-level float behaviour. P2 requires ties to resolve through one lexicographic preference table, not ad-hoc comparison — "reward shaping that flattens the objective without the matching preference-table row" is precisely the compliance failure named there.

Worth noting the tie detector does **not** flag period 18: its margin reads `1.488e-02`, orders of magnitude above float noise, while `value_slope` is `0.2945`. So whatever comparison actually flips is not the one `tie_margins` measures — that gap is itself part of the diagnosis.

## Why it matters now

This fixture is the corpus's most terminal-value-sensitive: applying #602's candidate change (dropping the double `cycle_cost` deduction) moves it **8.41 kWh** of boundary SOE. #602 raises terminal values across the board, so whatever is marginal here gets pushed harder, and #602's own write-up already predicts knife-edge behaviour near the terminal value — a `1.3e-5` change in the terminal value moving midnight SOE by 2 kWh.

Fixing this before #602 lands means #602's plan changes can be attributed to its economics rather than to interpreter noise.

## Contributing repo inconsistency

`.python-version` pins **3.12.13** while CI's algorithm-tests job runs **3.13.15**. Combined with bit-exact golden pinning, that guarantees a divergence of this kind is invisible in local development and only ever surfaces in CI. Worth resolving regardless of the outcome here — though aligning the versions would only hide this instance, not fix the selection non-determinism behind it.

## Reproduction

```bash
git fetch origin && git checkout chore/retrofit-scenario-terminal-value
.venv/bin/python -m pytest \
  "core/bess/tests/unit/test_action_selector_parity.py::test_selector_refactor_is_bit_identical[regression_2026_08_13_145213]" -q
```

Passes on Python 3.12.13, fails on 3.13.15. Inspect the decision directly:

```python
from core.bess.tests.helpers import _scenario_inputs
from core.bess.dp_battery_algorithm import optimize_battery_schedule
import json
s = json.load(open("core/bess/tests/unit/data/regression_2026_08_13_145213.json"))
d = {}
r = optimize_battery_schedule(**_scenario_inputs(s), tie_diagnostics=d)
print(r.period_data[18].decision.battery_action, d["tie_margins"][18])
```

## Suggested scope

1. Identify the comparison at period 18 whose outcome flips, and why `tie_margins` does not see it.
2. Route that selection through `tie_policy.py` so it is deterministic regardless of float ordering.
3. Then remove the carve-out #605 adds (see below) and restore the fixture's bit-exact pin.

## Current workaround

#605 excludes this one fixture from the bit-exact `actions` pin, documented and pointing here. Everything else about it stays pinned. The carve-out should be deleted as part of fixing this.

Related: #602, #605, #450, #512.

bess-agent:
## Summary

- The DP emitted a different plan depending on the Python interpreter, at bit-identical cost. `tie_policy.py` gains row 5, which resolves exactly-tied candidates by a stated order over actions instead of by a float comparison.
- Deletes #605's carve-out and restores this fixture's bit-exact plan pin.

## Root cause

Periods 17-20 of `regression_2026_08_13_145213` are four bit-identical 15-minute slots inside one hourly price block — same buy price, sell price, home consumption and solar — and all four are unconstrained `BATTERY_EXPORT`. The objective is therefore **exactly invariant** to how a fixed total discharge is split across them: only the sum reaches period 21. Every split is an optimum.

This is structural, not incidental, and the terminal value is what put the block on the plateau. At `terminal_value_per_kwh = 0.0` the boundary energy is worthless and the degeneracy never arises — which is why it only appeared once #605 retrofitted the corpus onto production terminal values, and why `origin/main`'s goldens were not already environment-fragile.

On the plateau the two candidates' values differ by **one ULP** (0.93 ULP recomputed in exact rational arithmetic over the same float reward and `V[19]` row), while `epsilon_for_period` here is 7.36e-4 SEK — the gap is ~8e11 times *inside* epsilon. Both candidates are eligible, rows 3 and 4 both stand down (`net_load_p < 0`; `sell_price_floored` False), and the table's final row — "the argmax winner stands" — hands the decision back to `>` in `action_selector.select_action`'s argmax loop. That is the P2 violation: the one row that is not a preference but a deferral to the last bit of a float.

`tie_margins` could not see it and is not malfunctioning. `_tie_margin` measures the gap to the best *behaviourally distinct* alternative, where distinctness is SOE separation alone (`TIE_DEDUP_SOE_KWH = 1.0`). The two candidates sit **exactly 0.5 kWh apart**, half the dedup distance, so the flipping candidate is skipped by construction and the reported 1.488e-02 is the gap to something else.

One correction to the issue's framing: every float op on the path producing `V` is IEEE-exact and bit-reproducible (`np.where/minimum/maximum/abs/arange/round/clip/min/max` plus scalar arithmetic — no reductions, no transcendentals, no `math.*`), and this fixture carries its prices verbatim. So "3.12 vs 3.13" is likely not the real variable; what differs is a whole environment. `test_action_selector_parity.py:15-28` already records that class of ULP divergence. The new fact is that the plan came to rest on a 0.93-ULP decision boundary.

## Fix

Row 5: among candidates whose value is within `VALUE_INDISTINGUISHABLE_SEK = 1e-12` of the argmax, pick the **most decisive action of the winner's own class** (largest magnitude), then the lowest candidate index. Keyed on `power`, never on `value`.

Two guards keep it a tie-*break* rather than a re-decision:

1. It stands down when an earlier row moved the pick — rows 3 and 4 stated deliberate preferences.
2. It never changes the action's class. Without this, a flat continuation value function turned IDLE into a full-rate discharge, and charge winners flipped to discharges — the semantic change row 3's docstring records that no phase has declared. **Both were observed failing** (`test_pwl_..._prefers_idle_when_flat_continuation`, `test_charge_winner_is_untouched`) before the guard existed, and are now pinned directly.

**The threshold is deliberately not epsilon, and that is not a P5 breach.** Epsilon is the *economic* band (~7.4e-4 SEK here) and belongs to `epsilon_for_period`. Reordering the whole epsilon-eligible set is a real preference change — measured at 0.006-0.010 SEK/fixture on several fixtures — and #606 asked for no such thing. This row reorders only what the objective declares *indistinguishable*, which needs a float-noise threshold on a quantity whose own accumulated rounding is ~1 ULP (~1e-15 at these magnitudes). 1e-12 is ~1000x that noise floor and ~1e7 finer than the smallest genuine cost movement this corpus has recorded (0.0181 SEK).

The charge and idle directions are **not exercised by any fixture**: this key and its `(power, index)` predecessor, which ordered charge ties the opposite way, produce byte-identical output on all 38. The direction is chosen for consistency with rows 3 and 4 rather than measured, stated as such in the docstring, and pinned in `test_tie_policy.py`.

### Measured delta

4 of 38 fixture plans move:

| fixture | Δcost | periods |
|---|---|---|
| `synthetic_clear_sky_ac_clipping` | **−0.03900** | 12 |
| `regression_2026_08_13_145213` | ±0.00000 | 2 |
| `realworld_2026_04_19_084608` | ±0.00000 | 2 |
| `regression_2026_08_08_143843` | **+0.00154** | 11 |

Net **−0.03746 SEK**; worst case +0.00154, inside the #450 0.05 SEK/fixture budget. `realworld_2026_04_29_220919` shifts 2.1e-12 SEK with its plan unchanged (an equal-power duplicate candidate — what the index tiebreak is for).

Zero in *model* value by construction; the residual is grid-snapping interpolation error, since an exact tie in interpolated value is not an exact tie in realized cost. Goldens and the VPP **current** half are re-pinned via `--repin-current`; the v10.0.2 half is verified byte-identical, so `test_drift_from_the_released_version_is_recorded` keeps its signal.

## Test plan

- [x] `./scripts/quality-check.sh` passes locally (1945 passed / 50 skipped)
- [x] `.venv/bin/pytest -m slow` passes (543 passed, 7 skipped, 232s)
- [x] **Observed on the reported interpreter.** Ran the fixture through the real backend container (`docker-compose.ci.yml`, py**3.13.14** / numpy 2.4.6):

  | row 5 | period 18 | `battery_solar_cost` |
  |---|---|---|
  | disabled | **−0.6875** ← the plan CI reported | −4.954319746390006 |
  | enabled | **−1.1625** ← matches py3.12.13 | −4.954319746390006 |

  The container reproduces the original cross-interpreter divergence *and* shows the fix closing it, at bit-identical cost — confirming the tie premise on the real reported interpreter rather than by local analysis alone.

Written RED first, then re-confirmed by mutation against the final code:

- Reverted: `apply_tie_policy`'s call to `_row_canonical_among_indistinguishable`, returning `chosen_index` directly (the pre-fix row 5).
- Result: `test_plan_is_stable_under_a_one_ulp_input_perturbation` FAILED — `At index 18 diff: -0.6875 != -1.1625`. 1 failed, 1 passed.
- The mutation output **is CI's reported plan**, so the test targets the reported defect rather than something resembling it.
- Restored: tree clean; independently re-confirmed inside the py3.13.14 container (table above).

The class guard has its own separate evidence: without it, 5 tests failed across `test_pwl_window_dp.py`, `test_tie_policy.py` and `test_export_limit_curtailment.py`.

## Outcome-level coverage

- `test_tie_determinism.py` drives the **real DP end-to-end** on a corpus fixture and asserts the emitted plan and the cost — not a unit test on the changed function with hand-built arguments. It perturbs one input by a single ULP (2.2e-16 relative), which makes the non-determinism observable on one interpreter; a test that only fails on py3.13 could not be run here.
- A second test in the same file pins the premise — that the two plans cost the same — so the fix cannot silently start forfeiting money.
- `expected_results` on `regression_2026_08_08_143843` re-pinned with the delta and its reason in the fixture's own `_note`.
- Selector goldens re-pin `actions`, `intents`, `intra_period_discharge_allowed` and the SoE trajectory bit-exactly for all 4 moved fixtures; `test_vpp_regression_baseline.py` re-pins realized cost and SoE through the VPP simulator.
- `test_tie_policy.py` gains 6 direct tests for the charge/idle classes, the class guard, index totality, the band boundary, and non-overrule of an earlier row — the paths no fixture reaches.

## Documentation

- `docs/agents/optimizer-architecture.md` — added the new row to the normative baseline order, plus a consequence bullet recording that the emitted plan never depends on a float's last bit, with the measured 0.93-ULP boundary.
- `docs/agents/bess-knowledge.md` — its tie-break invariant covered only plateaus created by *reward shaping*; #606's arises from identical inputs, so that category is now named, with the bundle signature to recognise it. The existing row-4 reference stays accurate (charge-early is still row 4).
- `docs/SOFTWARE_DESIGN.md` — grepped, mentions nothing this diff touches.

## Note on `.python-version`

Left alone deliberately. Aligning it to CI's 3.13 would hide this instance rather than fix it, and would make the discriminating test impossible to write in the form that actually reproduces the defect. Separately, PR #547's Python 3.13 base-image alignment turns out never to have reached `main` — it merged into `fix/numpy-pin-py311` 55s after that branch was squash-merged — so the shipped add-on still builds on Alpine 3.19 (Python 3.11). That is being recovered in its own PR.

Closes #606


bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

PR #609, opened 2026-08-16
fix: derive the signed sensor pair on read, not only at discovery (#604)

### LINKED ISSUE #604: Critical system issues detected
## What happened
The system reported the following issues:

- Battery Monitoring [ERROR]: Critical sensor configuration issue detected

Detailed Checks:

Battery State of Charge

OK

Value: 84.0 %

Battery Charging Power: OK Value: 1,271 W
Battery Discharging Power: Method returned None Value: N/A
Error: Method returned None

## Debug bundle
Please drag-and-drop **bess-debug-2026-08-15-170653.md** from your Downloads folder into the editor below before submitting.
The file contains your version, settings, recent logs, and the raw HA discovery dump that the maintainers need to reproduce the issue.

[bess-debug-2026-08-15-170653.md](https://github.com/user-attachments/files/31103232/bess-debug-2026-08-15-170653.md)

johanzander:
## Summary
- A Huawei/SolaX/Solis install whose settings were persisted before the signed-sensor pairing existed had the sign split silently switched off, reporting a discharge as negative charging and discharge power as `None`.
- The pairing is now derived on every read of the active sensor map instead of only when discovery writes it, so every existing install self-heals — no migration, no wizard re-run.

## Root cause

Native SolaX and Huawei publish battery power as one signed register (`battery_power_charge` / `storage_charge_discharge_power`, positive = charging); Solis and Huawei publish grid power the same way. `HAApiController` splits that single reading by sign, but the split is gated on **both** directional keys resolving to the same entity:

```python
charge_entity = self.sensors.get("battery_charge_power")
discharge_entity = self.sensors.get("battery_discharge_power")
return bool(self.battery_power_polarity and charge_entity and charge_entity == discharge_entity)
```

That pairing was created only by `discover_sensors_from_registry` — at *write* time. The reporter's persisted config (v10.1.0b9, `huawei_solar_luna2000`) has:

```
"battery_charge_power": "sensor.batteries_charge_discharge_power"
# battery_discharge_power: absent
```

so the predicate was `False` and the raw signed value was handed straight through. The failure is silent, not an error: `get_battery_charge_power()` returned `-375` while discharging and `get_battery_discharge_power()` returned `None`, feeding `get_net_battery_power()`, `/api/realtime_power` and the dashboard.

## Fix

`settings_store.apply_signed_pair_aliases(platform, sensors)` points the derived key (`export_power`, `battery_discharge_power`) at the one entity the integration publishes, for any platform listed in `PLATFORM_GRID_POWER_POLARITY` / `PLATFORM_BATTERY_POWER_POLARITY`. Which platforms are single-signed is a fixed integration fact already recorded in those maps, so the derivation belongs where that fact is read.

It runs inside `flatten_sensors()`, which backs `SettingsStore.get_active_sensors()` and therefore `HAApiController.sensors` — documented as a computed view, never a synced snapshot — so stale settings cannot survive it. The three inline copies in `discover_sensors_from_registry` (SolaX battery, Solis grid, Huawei both) now call the same helper, so the derived key still lands in `platform_sensors` and is reconciled off `platform_disabled`. An explicitly mapped counterpart is never overwritten.

## Scope assessment

- **Structural**, deliberately: the fix moves an existing derivation to a different owner rather than loosening the predicate in place. Patching `_is_shared_signed_battery_power()` to tolerate a missing counterpart would fix the readings in ~4 lines but leave the config still wrong everywhere else it is read (health-check `entity_id`, wizard required-sensor list, `sensor_collector` power-sensor resolution).
- **Workaround check passes**: no new parameter, flag, default-fallback, extra trigger or branch. Three copies of an existing alias become one helper, called from the read path that was missing it.
- Every duplicate of the pattern is covered (battery: SolaX native + Huawei; grid: Solis + Huawei), not just the reported one.

## Test plan
- [x] `./scripts/quality-check.sh` passes locally (1954 passed, 50 skipped; frontend 125 passed)
- [x] `.venv/bin/pytest -m slow` — 542 passed, 8 skipped
- [x] **Observed in the running app** (`docker-compose.ci.yml`, `SCENARIO=ci-wizard-huawei-luna2000`, `BESS_SETTINGS=./e2e/ci-bess-settings-huawei.json` — that fixture already has the reporter's exact shape: `battery_charge_power` mapped, `battery_discharge_power` absent). Battery at `-900 W` (discharging), via `GET /api/system-health`:

  | | Battery Monitoring | Charging Power | Discharging Power |
  |---|---|---|---|
  | with the fix | `OK` | `0 W` | `900 W` |
  | fix reverted, container restarted | `ERROR` | `-900 W` | `N/A` — "Method returned None" |

  The reverted row is the issue report reproduced verbatim in the running system.

Written RED first, before the implementation existed:

- Reverted: the entire fix absent (tests written against `flatten_sensors` and a `SettingsStore`-backed controller before `apply_signed_pair_aliases` existed)
- Result: **8 of 11 new tests FAILED** — `test_huawei_battery_discharge_derived_from_signed_charge_entity`, `test_solax_native_battery_discharge_derived_from_signed_charge_entity`, `test_solis_export_derived_from_signed_import_entity`, `test_huawei_export_derived_from_signed_import_entity`, `test_get_active_sensors_exposes_the_derived_pair`, plus all three of `TestSignedPowerOnSettingsPersistedWithoutThePair`. The 3 that passed are the negative cases (two-entity platform, explicit mapping, unmapped primary) — correctly green both before and after.
- Additionally mutated in the running container: `flatten_sensors` returning `result` instead of `apply_signed_pair_aliases(platform, result)` flipped Battery Monitoring from `OK` to `ERROR` with `-900 W` / `None`, then restored clean.

## Outcome-level coverage

- `TestSignedPowerOnSettingsPersistedWithoutThePair` asserts the **readings** (`charge == 0`, `discharge == 375`, `net == -375`, and the grid twin) through a real `SettingsStore` in the persisted per-platform shape — not the alias dict, and not a hand-built flat sensor map.
- No optimizer fixture pin: the change touches sensor-key resolution only, not the DP, intent classification, or control/rate mapping, so no golden or `expected_results` moves.

## Documentation

`docs/SOFTWARE_DESIGN.md:433` explicitly claimed the pairing was arranged by `discover_sensors_from_registry`; updated to describe the read-time derivation and the flat-shape exception. `docs/agents/bess-knowledge.md` mentions nothing this diff touches.

## Review findings

`/code-review high` returned 3 low findings, no CONFIRMED blockers. Two were about this diff's own accuracy and are fixed here (the "every read" wording overstated the legacy-flat branch; the changelog omitted Solis). The third is pre-existing — `sample_live_power()` gates on `power_sensors`, the InfluxDB list that deliberately *excludes* shared signed entities, even though its own loop uses the sign-splitting getters — so it is parked in `TODO.md` rather than widening this PR.

Closes #604


bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

PR #610, opened 2026-08-16
chore: recover the Python 3.13 add-on base alignment (#547)

bess-agent:
## Summary

Recovers PR #547's Python 3.13 base-image alignment, which was merged but never reached `main`, and finishes the job by aligning the four places still left on 3.12.

## Why this is needed

**#547 was merged into a branch that had already been squash-merged.**

- #546 (`fix/numpy-pin-py311`) → `main`, squash-merged `2026-08-11T22:13:09Z`
- #547 (`chore/ha-base-python-313`) → `fix/numpy-pin-py311`, merged `2026-08-11T22:14:04Z` — **55 seconds later**

The squash captured only #546's own content. #547 landed in a branch that was already dead, and nothing re-PR'd it. `git branch -a --contains 7184a5a1` returns exactly one ref: `origin/fix/numpy-pin-py311`.

**The consequence is live in what ships today.** `main`'s `build.json` and `release-addon.yml` still name `ghcr.io/home-assistant/*-base:3.19` — an Alpine base past EOL (Nov 2025) whose `apk add python3` yields **Python 3.11**. So:

| where | Python (before this PR) |
|---|---|
| **published add-on** | **3.11** |
| CI test jobs | 3.13 |
| dev / E2E container | 3.13 |
| `prod-test` compose | 3.13 (`python:3.13-alpine` — an image nobody ships) |
| **local dev** (`.python-version`) | **3.12.13** |
| Stage 3/5 bot workflows | 3.12 |
| mock-HA container | 3.12 |

Nothing tests the interpreter that ships. `backend/requirements.txt` still says so in tracked prose: *"the HA add-on base ships Python 3.11, and numpy >=2.5 requires >=3.12 — pinning 2.5.2 broke the supervisor build outright."* That is the failure this alignment was written to prevent, still in force.

## What's in here

**Cherry-picked `7184a5a1` unchanged.** Five of its six files were byte-identical to their current state on `main`, so it applied cleanly; `.github/workflows/ci.yml` auto-merged against the unrelated #549 wizard E2E steps added since. It:

- moves all four places the base is named onto `ghcr.io/home-assistant/*-base-python:3.13-alpine3.22` (`build.json`, `release-addon.yml`, `docker-compose.prod-test.yml`, the `Dockerfile` ARG default) — they had drifted apart;
- drops `python3`/`py3-pip`/`python3-dev` from both apk layers. `base-python` ships a source-built interpreter at `/usr/local/bin`; apk would install Alpine's own alongside it at `/usr/bin`, leaving `python3 -m venv` dependent on PATH order. `gcc`/`musl-dev` stay (watchdog builds from source);
- builds `backend/Dockerfile` in CI the way the supervisor actually does. There are two Dockerfiles and CI only built one — `package-addon.sh` ships the *other*, which had no coverage at all, and that is precisely how a 3.11 base plus a numpy pin needing ≥3.12 reached a real install;
- drops the dead `armhf`/`armv7`/`i386` entries and fixes the `io.hass.arch` labels (5 vs 3 vs 2 across three files);
- builds the frontend stage on `$BUILDPLATFORM` instead of a hardcoded `linux/amd64`, which forced QEMU on an arm64 dev machine.

#547 verified all of that against the real aarch64 base: the venv resolves to `/app/venv/bin/python3` (3.13.14) and numpy/pandas/fastapi/anthropic import from musllinux wheels.

**On top of the cherry-pick — the stragglers the alignment left behind:**

- `.python-version` 3.12.13 → **3.13.14**.
- `issue-fix.yml`, `issue-integrate.yml` 3.12 → **3.13**. Stage 3 runs the test suite before opening its PR, so on a different interpreter it green-lights changes `ci.yml` then rejects — the same trap, inside the automation.
- `scripts/mock_ha/Dockerfile` 3.12 → 3.13. It never imports the optimizer, so this is consistency only.
- `backend/requirements.txt`: the numpy comment asserted a Python 3.11 base as the *reason* for the 2.4.6 pin. That stops being true here, so the comment now records it as a plain version pin.

## What is deliberately NOT in here

**numpy is not bumped.** The pin has a second purpose — stopping reported costs moving silently, paired with the goldens' 1e-9 tolerance. Removing the interpreter constraint makes 2.5.x *installable*; whether to take it is a separate decision requiring a measured delta.

**This does not make bit-exact golden pinning safe.** Production is musl/Alpine, CI is glibc/Ubuntu — different libm, so transcendental results can still differ in the last bits at the same Python version. Aligning versions removes the local-vs-CI blind spot; it does not remove the failure class. The plan-selection fragility that exposed this is fixed separately in #608 (#606), and that fix is what actually makes the pins robust.

## Test plan

- [ ] CI: `Docker build & boot` builds `./Dockerfile` on the real HA base, and the new step builds `backend/Dockerfile` the supervisor's way and asserts the interpreter and numpy version inside the resulting image — this PR's actual verification, and it could not run before because that path had no coverage.

## Note for the maintainer

`.python-version` → 3.13.14 needs `pyenv install 3.13.14` locally. Nothing breaks the moment this merges — the existing `.venv` is 3.12 and is shared by symlink across worktrees — but the venv wants recreating to actually develop on 3.13.

Refs #547, #546.


bess-agent:
@claude-bot review

PR #612, opened 2026-08-16
feat: Product Owner agent — backlog digest, role identities, skill and agent

johanzander:
Implements Tasks 1–4 of `docs/superpowers/plans/2026-08-16-product-owner-agent.md`, per the design in `docs/superpowers/specs/2026-08-15-backlogger-agent-design.md`.

## What this adds

A Product Owner agent that owns the backlog — intake, refinement to a Definition of Ready, prioritisation, and dispatching implementation sessions. It never implements and never creates worktrees; `implement-issue` is untouched.

| Piece | File |
|---|---|
| Evidence gatherer | `scripts/backlog-digest.sh` — joins issues, PRs, worktrees, sessions and board state into one JSON document, so no model reads 37 issue bodies to decide what is next |
| Role identity | `scripts/gh-agent.sh --as po\|dev` |
| Procedure | `.claude/skills/backlog/SKILL.md` |
| Agent | `.claude/agents/product-owner.md` |

All durable state lives on GitHub. No local file mirrors board or issue state.

## Not in this PR

Tasks 5–8 are deferred: they need a GitHub account, a PAT and Actions secrets that only the maintainer can create, plus approval of the roadmap themes. **Merging this does not yet produce a working agent** — a pass fails loudly at the first board access until `PROJECT_NUMBER` and a board exist.

## Maintainer steps required

- [ ] `gh auth refresh -s project`
- [ ] Create `bess-product-owner` (PAT needs `repo` **and** `project` scope), add `BESS_PO_TOKEN` to `.env` and to Actions secrets
- [ ] Later, when renaming `bess-agent` → `bess-developer`: change the login in `pr-review.yml`'s gate, the `dev` token variable in `scripts/gh-agent.sh`, and the `bess-agent` references in `docs/agents/workflow.md` — **all in one commit**, never separately

Nothing else is required to merge. The `dev` role keeps using the existing `BESS_AGENT_TOKEN`, verified live (`scripts/gh-agent.sh --as dev api user` → `bess-agent`).

## Security note

An earlier revision pre-authorised `bess-developer` in `pr-review.yml`'s actor gate as a migration window. The final review measured that the username is **unregistered** and this repo is **public** — anyone could have claimed it and triggered a bypass-permissions workflow holding a write-scoped token by commenting `@claude-bot`. Reverted in `cb663aa6`; the gate accepts only the owner and `bess-agent` until the account actually exists.

[...]

PR #614, opened 2026-08-16
fix: enforce mypy on changed files, and make the PR review bot always submit a verdict

johanzander:
## Summary
Three gaps found while landing #612. None were caused by that PR — it just surfaced them.

**Scope note:** this PR previously also carried an undescribed backlog/board bundle. Per the review and the maintainer's direction on this thread, that half has been split out to #656; nothing was lost (its full state is preserved at `eb4d00e9`). This PR is now exactly the three changes it describes, plus the test and severity fix the review asked for.

## 1. mypy was required but enforced nowhere

`docs/agents/rules.md` says code must pass mypy with zero errors. Nothing ran it: not `quality-check.sh`, not CI, and it was not even in `requirements-dev.txt`. A violation reached `main` in #612 and was only caught because the Stage 4 review bot happens to run mypy independently.

Turning it on repo-wide is not viable. Measured:

| Scope | Errors | Files |
|---|---|---|
| `core` + `backend` + `scripts` | 2914 | 191 |
| Production only (excluding tests) | 417 | 39 |

So the gate runs mypy against **files changed vs `origin/main`**. Code you touch must type-check; the legacy backlog burns down as files get edited. Untracked files are included deliberately — a new file is untracked until its first commit, and this gate runs *before* that commit.

`mypy` added to `requirements-dev.txt`.

### And on CI, which is the half that actually gates merges

Review round 3 found the first version of this PR only half-delivered its own opening claim. The gate went into `quality-check.sh` — which runs locally and in the Stage 3 flow — while `ci.yml`'s `Code quality` job still installed only `black ruff`. That job aggregates into the **Merge gate**, the single required status check on `main` (`docs/agents/workflow.md`), so the exact scenario described above was still reachable through any ordinary PR.

`ci.yml` now runs the same scoped check, reusing the same policy rather than inventing a second one: changed against the merge-base with `main`, deletions dropped, repo-wide left alone. The job checks out full history because a shallow clone has no merge-base to compute.

Verified from the run log rather than the check colour — "Code quality pass" was the reviewer's own evidence that a passing job proves nothing here:

```
Checking 1 file(s):
  backend/tests/test_quality_check_mypy_gate.py
Success: no issues found in 1 source file
```

### The severity bug the review found — fixed

When `origin/main` could not be resolved, the check degraded to a `WARNING`, and warnings exit 0. That printed `Errors: 0` for a run that type-checked nothing — the exact anti-pattern the comment above the missing-tool branch already calls out. It is now an `ERROR`, with the reasoning recorded next to it.

### And the test coverage the review found missing — added

`backend/tests/test_quality_check_mypy_gate.py` drives the real script in a throwaway directory with `git`, `black`, `ruff`, `mypy` and `pytest` as PATH shims (same approach as `test_backlog_digest.py`). Only the `git` shim's `merge-base` arm differs between the two runs.

## 2. The review bot went silent when it found nothing

`pr-review.yml` told the bot to end with a summary review but didn't make it mandatory, so a clean run posted nothing at all.

That's worse than cosmetic: **GitHub blocks a merge on the last explicit verdict.** A prior `REQUEST_CHANGES` therefore survives every subsequent clean run, and no amount of re-reviewing can clear it — #612 had to be dismissed by hand to merge. Now mandatory in every case, with the reason stated in the prompt so it doesn't get "simplified" away later. `request-pr-review.sh`'s header updated: a timeout is once again a real signal rather than the usual clean case.

## 3. `gh` fails open to a scopeless token

When a keychain read fails, `gh` doesn't error — it silently falls back to a token with fewer scopes. The symptom is `Forbidden` on an operation you *do* have rights to, which reads as a permissions problem and isn't. Cost a wrong diagnosis during #612: the same `gh api` call returned `Forbidden` in-sandbox and succeeded outside it, same account.

Recorded with how to tell the two apart (`gh auth status` — `✓ ... (keyring)` vs `Failed to log in ... (keyring)`). Since main moved that whole section out of `CLAUDE.md` (#651) while this PR sat, the note now lands in `docs/agents/local-agent-environment.md` alongside the sandbox findings it belongs with.

## Test plan
- [x] `./scripts/quality-check.sh` passes locally: 0 errors, 0 warnings
- [x] The mypy gate exercised in all three states — catches a real error, passes when clean, passes when no Python changed
- [x] New test run green, and run red under mutation (below)

Four mutations, each reddening its own test and only its own:

| Mutation | Result |
|---|---|
| mypy failure counted as a warning | `test_type_error_in_a_changed_file_fails_the_gate` FAILED |
| changed-file list never populated | 2 FAILED, incl. the vacuity guard |
| unresolvable ref back to a warning | `test_unresolvable_origin_main_fails_the_gate` FAILED |
| file list back to a space-joined string | `test_a_changed_path_with_a_space_is_still_checked` FAILED |

Restored after each: tree clean, 4 passed.

The second one matters most, because it is the defect review round 2 caught. The first version of this test shimmed `mypy` to `exit 0` and left the git shim's file-list arm empty, so the gate took its "no changed Python files" path and mypy was never invoked — the control asserting "the gate actually checks types" was two log-string assertions true either way. Now mypy is the real binary and the file list is non-empty, and that mutation reddens two tests instead of passing silently.

One further correction worth recording: the rewritten test passed locally and failed CI, because it reached mypy through `.venv/bin/mypy`, which CI has no such layout for. mypy then failed identically in every run, both deltas collapsed to zero, and the control never saw its string. It now goes through `<the interpreter running pytest> -m mypy`, which holds under both layouts.

## Outcome-level coverage
The outcome asserted is the gate's exit behaviour — the error count and a non-zero exit — not the text it prints. That is the whole behaviour at issue: warnings exit 0, so severity *is* the outcome here. No optimizer fixture or golden applies; this diff touches no DP, intent or control/rate path.

## Documentation check
Grepped `docs/agents/bess-knowledge.md` and `docs/SOFTWARE_DESIGN.md` for everything this diff touches (`quality-check.sh`, mypy, `pr-review.yml`, the review verdict flow): neither file mentions any of it — they document optimizer behaviour and system design. `docs/agents/local-agent-environment.md` is the file that does cover this ground, and it is updated here.

Local verification only — no app behaviour changes, so there is no mock-HA scenario to observe; the gate itself was run directly, which is the equivalent.

No `CHANGELOG.md` entry: all three changes are developer tooling with zero user-visible effect on the add-on.


bess-agent:
@claude-bot review

johanzander:
## Session died without starting — picking this up needs a fresh run

A background session (`pr614-split`) was dispatched to act on the `CHANGES_REQUESTED` review and produced **nothing**: `HEAD` still `eb4d00e9`, 5 commits ahead, clean tree, no split branch, no tracked file written in 2.5 hours. It has been stopped. Nothing was lost; nothing was done.

Recording the work here so the next attempt does not have to reconstruct it from a dead session's transcript.

### Decision taken: split this PR in two

Per the review's primary blocker (undescribed, orthogonal board changes making up over half the diff):

**Stays in #614** — the changes this PR actually describes:
- `scripts/quality-check.sh` mypy enforcement
- `.github/workflows/pr-review.yml` mandatory-verdict instruction
- `CLAUDE.md` `gh` scopeless-token note

**Moves to a new branch off `origin/main` with its own draft PR:**
- `.claude/skills/backlog/SKILL.md`
- both files under `.claude/agent-memory/product-owner/`
- `scripts/backlog-board-init.sh`
- `scripts/backlog-digest.sh` column renames + matching `backend/tests/test_backlog_digest.py` updates

### Two defects to fix while splitting

1. **Stale `P0`** — `SKILL.md` and `backlog-board-init.sh` both encode a `P0` Priority option that this PR's own new memory file (`project_board_field_ids.md`) documents as nonexistent on the live board. Confirmed against the board: options are `P1, P2, P3, P4`. Drop the `P0` references; the memory file is correct. Fix this in the **new board PR**.
2. **`quality-check.sh` mypy severity** — when `origin/main` cannot be resolved the check degrades to a `WARNING`, and warnings exit 0, so it prints `Errors: 0` for a run that type-checked nothing. Make it an `ERROR`. Fix this in **#614**.

### Note for whoever picks this up

The board changes overlap with pending work: the digest currently derives `Ready` / `In progress` while the live board's `Status` options are `Ready for Dev` / `In Progress` / `Done`. Separately, a fix is queued for the digest's column precedence — `analyzed` is checked *before* `awaiting`, so an analysed item that is waiting on a decision still reports `Ready` (this is how #96 was dispatched into a dead end). Coordinate those in the new board PR rather than in #614.

Both branches must pass `./scripts/quality-check.sh` and `.venv/bin/pytest -m 'not slow'` before pushing. Neither gets `Closes/Fixes`.

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

PR #615, opened 2026-08-16
fix: round the charge command up so it delivers the planned charge (Phase 4c, narrow)

johanzander:
## Summary

- The charge rate is written as a whole percent scaled from the plan's own action, rounded to **nearest** — so a planned power between two lattice steps was written as the step *below* and the battery charged less than the plan counted on.
- Measured over the corpus: **4 of 493** charging periods short, worst **−0.0288 kWh** (`synthetic_extreme_volatility:13` — plan 0.4488 kWh, command 7% of 6 kW = 0.4200 kWh). All `GRID_CHARGING`.
- Rounds **up** now. **Corpus cost +0.00000 SEK, 0 of 38 fixtures moved** — this changes what is *written*, not what is *planned*, so there is no golden or baseline re-pin anywhere in this PR.

## 4c's documented premise was false — corrected here

The plan and the design doc both said the executor is configured to deliver less charge than the planner assumes, citing `charging_power_rate` (default 40%) as the value written to the inverter. **It is not written to the inverter at all.**

`battery_system_manager` writes `get_period_settings(period)["charge_rate"]` — the *intent-derived* rate. `INTENT_TO_CONTROL` gives a flat `charge_rate: 100` for `SOLAR_STORAGE`/`IDLE`, and `GRID_CHARGING` is scaled from the plan's own action by `_compute_charge_rate`. `charging_power_rate` reaches hardware only as `power_monitor.py:67`'s *initial* target, `power_monitoring_enabled` defaults to **False**, and even when enabled that target is overwritten every period.

So the savings 4c was scoped around do not exist. Both documents now carry the correction inline, marked so nobody re-derives the old premise from them. The plan's *first* bullet stands and is recorded as still-open (below).

## Fix

`command_index(..., rate_is_ceiling=True)` — the same conversion 4b introduced, renamed from `discharge_command_index` since both paths now share it.

The reason mirrors the discharge side but rests on **different physics**, and that is worth keeping straight:

| | what binds above the command | so round |
|---|---|---|
| discharge ceiling (4b) | actual house load | **up** |
| charge, room-limited (4c) | the battery's own remaining room — inverter stops when full | **up** |
| charge, import-cap-limited (#429) | **nothing** | plan floored to the lattice instead |

Where `import_cap_kwh` limited the plan, rounding up would draw more from the grid than the house fuse allows. There the DP floors the plan onto the lattice (`execution_model.lattice_grid_charge`) — safe because under-drawing never violates a fuse. The safe direction genuinely differs per binding constraint; they must not be unified.

## Rejected, with measurements rather than argument

- **Floor every grid charge onto the lattice** regardless of what bound it: **+1.33 SEK**, 11 fixtures moved. Charging less than the battery could compounds into lost arbitrage — the same shape as D3's +3.67 SEK for #352.
- **Floor all charging, solar included:** 311 further periods and **3.70 kWh** of storable solar discarded, for no fidelity gain — `SOLAR_STORAGE` is commanded at a flat 100% and never rounds.

## Left open, deliberately and on the record

- **`_period_flows` derives charge throughput from `max_charge_power_kw`, not the commanded rate.** The DP and the inverter simulator share that record under P4, so the simulator reproduces the plan by construction and **cannot disagree with it about charge**. That is why the corpus R==P harness reads `0.00000` while the written command is genuinely short — any charge-side R==P result is weaker evidence than the discharge-side equivalent. Closing this is what "4c-full" would mean, and it needs its own measurement first.
- One `SOLAR_EXPORT` period plans a 0.0089 kWh charge against a deliberate 0% command (#313 blocks passive charging) — a sub-noise-floor plan/command inconsistency of a different kind.

## Test plan

- [x] `./scripts/quality-check.sh` green (re-run after merging `origin/main`)
- [x] `pytest -m slow` green — 547 passed, 7 skipped, **no re-pinning required**
- [x] Corpus re-measured: charging periods with `deliverable < planned` go **4 → 0**

- **Reverted:** `rate_is_ceiling=True` → `False` in `_compute_charge_rate`.
  **Result:** `test_grid_charge_plan_is_deliverable_by_the_written_command` FAILED — *"period 13: the controller writes 7% but the simulator models 8% — the two charge mappings have drifted"*. `test_period_groups.py::test_grid_charging_uses_action_derived_rate` also FAILED (`assert 3 == 4`).
  **Restored:** both green.
- Worth recording: an earlier version of that test built its commands only via the simulator's mirror and **stayed green under this mutation**, because I had mutated the controller alone. It now asserts through the production path (`get_period_settings`) *and* cross-checks the simulator against it, so a drift between the two sites fails loudly — which is the property that actually matters here.

## Outcome-level coverage

- `test_grid_charge_plan_is_deliverable_by_the_written_command` — every `GRID_CHARGING` period on a real fixture: the controller's written rate must be able to deliver the planned charge, and the simulator must agree with the controller.
- `test_period_groups.py::test_grid_charging_uses_action_derived_rate` — the rounding direction pinned directly on the controller.
- This one asserts against the *command* rather than the simulator's `battery_charged`, deliberately and with the reason stated in the test: the simulator is blind to this quantity (see "Left open"), so a command-level assertion is the only faithful one until that changes. `rules.md` permits exactly this where no execution model exists.

Refs the Phase 4c entry of `docs/superpowers/plans/2026-08-09-optimizer-target-architecture.md`.

https://claude.ai/code/session_01GtTbxMVD8mqib3Rh1Sc34Q

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
+- **Grid charging now reaches the planned amount instead of stopping just short** — the charge rate is written as a whole percent, and rounding it down meant the battery charged slightly less than the plan counted on.
 - **The battery now covers house load exactly instead of exporting a few Wh and committing the inverter** — a period whose load fell between two discharge steps was planned as a small export, which forces `grid_first` at a fixed rate and imports any load spike. ([#352](https://github.com/johanzander/bess-manager/issues/352))
 - **Huawei, native SolaX and Solis installs no longer report a discharge as negative charging, or an export as negative import** — the signed battery/grid sensor split now works on settings saved before the sensor pairing existed, with no wizard re-run. ([#604](https://github.com/johanzander/bess-manager/issues/604))
 - **The battery schedule no longer varies between otherwise identical runs** — when two plans were worth exactly the same, the optimizer picked between them on floating-point noise, so the same day could plan differently on different machines. ([#606](https://github.com/johanzander/bess-manager/issues/606))
--- a/core/bess/action_selector.py
+++ b/core/bess/action_selector.py
@@ -176,7 +176,7 @@ def _residual_cover_p(
       nearest and a ceiling rounded *down* under-delivers the plan (repro:
       residual 0.12 kW -> 1% of 10 kW = 0.10 kW ceiling -> realized 0.10 for
       a planned 0.12). The write path now rounds ceilings up
-      (`execution_model.discharge_command_index`), so the whole round-DOWN
+      (`execution_model.command_index`), so the whole round-DOWN
       band is executable and the gate reduces to "is the ceiling
       commandable at all", i.e. at or below 100%. Rejecting 39% of the
       corpus's deficits was never the intent; it was the conversion's
--- a/core/bess/dp_battery_algorithm.py
+++ b/core/bess/dp_battery_algorithm.py
@@ -65,7 +65,11 @@ from core.bess.dp_constants import (
     POWER_STEP_KW,
     SOE_STEP_KWH,
 )
-from core.bess.execution_model import DEFAULT_CAPABILITIES, PlatformCapabilities
+from core.bess.execution_model import (
+    DEFAULT_CAPABILITIES,
+    PlatformCapabilities,
+    lattice_grid_charge,
+)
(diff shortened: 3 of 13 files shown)


bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

PR #617, opened 2026-08-16
chore: name the gh release verbs so reading releases stays unattended

johanzander:
## Problem

`.claude/settings.json` guarded releases with a blanket `Bash(gh release*)` ask, so a plain read prompted:

```
gh release list -L 5 -R johanzander/bess-manager-beta
Ask rule Bash(gh release*) overrides auto mode for this command.
```

That fails the standard CLAUDE.md states for the ask list — *the effect escapes the repo **and** git cannot undo it*. `gh release list` / `gh release view` fail the first half outright: they read GitHub and change nothing on it. Worse, the prompt fired on the exact command the release rule mandates ("always check the current published version before tagging") and on every beta-vs-stable version comparison.

## Fix

Name the verbs, exactly as the `git stash` twins already do — a blanket rule there hard-blocked `git stash list` for the same reason.

| Verb | Behaviour |
|---|---|
| `create`, `edit`, `delete`, `delete-asset`, `upload` | still asks — publishes/destroys on GitHub, unrecoverable |
| `list`, `view`, `download` | unattended — read-only, or writes only into the sandboxed repo |

`scripts/quality-check.sh` pins **both** directions by command string, matching how the file already treats push/stash: the mutating spellings go in `MUST_BE_GUARDED`, the read-only ones in `MUST_NOT_BE_GUARDED`. So the split cannot silently collapse back into a blanket rule, nor into a hole.

## Verification

The permission gate runs green: `✅ Permission surface intact (76 command shapes checked, 20 require deny, 18 must stay unattended)` — up from 68 shapes.

No behaviour change outside the permission surface; no Python touched.

https://claude.ai/code/session_01ApGTeBbaAZgigkDdAPqWyp


bess-agent:
@claude-bot review

Issue #593, opened 2026-08-15
Growatt VPP order of register values


ridax67:
I have been noticing spurious energy spikes when Bess is switching modes.  Bess is setting remote before power and is omitting power alltogether when load first is set.

I see that there is a risk that the inverter goes into a mode with a old power value and execute it before it gets the new power value.  This COULD explain the energy spikes I noticed, even if my ha logging can't find this.

Would it be possible to send power first and remote after and to send a zero power for load first?  I am not claiming it will fix this, but it might be the explanation and I feel this order is safer.

johanzander:
@claude-bot analyze

ridax67:
Optimun order for switch should be:

 Switch remote disable -> enable
Power first, remote last

Switching remote enable -> disable
Remote first, power 0 last

What I believe is happening now is Bess is ordering a battery dump and sends remote enabled and power -100.

After dump Bess switch to load first and sends remote disabled (power -100 is left untouched but inactive during this mode).

Next time when Bess switch to Solar export Bess wants to put inverter in grid 0.  It is now the error occurs.  Bess sends remote enabled, and inverter goes into Grid first -100 (old power value) and dumps battery.  Bess continues and sends power 0, inverter cut the battery power now, but not before a -10kW pulse has been sent out.

So sending remote signal first when disabling it and send power signal first when enabling, should be the safe order.

It is the -10kW (and sometimes +5kW charging pulse) that I have noticed sometimes, that have made me look for an explanation for them

johanzander:
@ridax67 — good catch, you're right on both counts.

Confirmed: BESS enables remote control first and sends power second, and on load-first it sends no power value at all.

That matters on Growatt because there's no separate trigger register — enabling remote control *is* the commit. So in the gap between the two writes the inverter is armed and executing the previous period's power value. And since we never clear power on release, it can hold a stale value for hours: battery export at −100% → a long load-first stretch → next active period arms at −100% full-rated export until the new value lands. Mirror image after a grid-charge period. That's your spike shape.

The window can also stretch — the power write retries with backoff, and if it fails outright the inverter stays armed on the old value until BESS retries the whole period minutes later. It's invisible in the logs because BESS logs its intent once, before either write goes out, which is why your HA logging couldn't find it.

Adopting your fix, with one adjustment:

- **Into an active mode:** power, then fallback timer, then remote control last — exactly as you suggest. The timer moves too, so a re-enable after a long gap doesn't arm against a stale dead-man's timer.
[...]

PR #618, opened 2026-08-16
fix: write Growatt VPP power before arming remote control (#593)

johanzander:
## Summary
- Growatt VPP now writes the power target *before* arming remote control, and clears it on release, so a mode switch can no longer briefly execute the previous period's command.

## Root cause
Growatt VPP has no separate trigger entity — writing `growatt_vpp_remote_control` (30407) **is** the commit, so the inverter immediately executes whatever 30409 (power) and 30408 (timer) already hold. `set_growatt_vpp_period` wrote remote control first and power second, and returned early on release without writing power at all:

```python
self._service_call_with_retry("select", "select_option", ...)   # 30407 -- the commit
if not remote_control_enabled:
    return                                                       # 30409 never cleared
power_entity = self._get_entity_for_service("growatt_vpp_power") # 30409, too late
```

`LOAD_SUPPORT` and `SOLAR_STORAGE` both map to `remote_control_enabled=False`, so 30409 retained the last *active* period's value — across a multi-hour load-first stretch. Every Disabled→Enabled transition then armed against that stale value: `-100` after `BATTERY_EXPORT` (full-rated export) or `+100` after `GRID_CHARGING` (full-rated charge). That is the spike reported in #593.

It is invisible in logs because BESS emits one line *before both* writes, describing the intended pair — which is why the reporter's HA logging couldn't find it.

## Fix
- **Activation:** power (30409) → timer (30408) → arm (30407) **last**. The timer moves for the same reason as power: written after the commit, a re-enable after a long release armed against a lapsed dead-man's timer.
- **Release:** disable (30407) **first**, then write power 0. The issue proposed zeroing first; that isn't inert — a 0 written while remote control is still enabled selects the documented `grid_first` hold, changing release behaviour on every `LOAD_SUPPORT` entry. This order clears the latch for the next activation while leaving release identical to before.
- The power entity is resolved **per branch**, not once up front: on activation before any write (a mis-provisioned install fails with the inverter untouched); on release only *after* the disable has landed (returning to `load_first` is the safety-critical half; clearing the latch is a courtesy to the next activation).
- `_last_written_vpp_power` is now `0` on release rather than `None`, since the fix makes 0 the hardware truth. Untested by design — its only consumer reads it exclusively when remote control is enabled, so it is unobservable on the release path.

Intent→VPP mapping, the DP, and intent classification are untouched; no plan economics move.

## Test plan
- [x] `pytest -m slow`: 546 passed, 7 skipped
- [x] **Observed end-to-end** on the real backend against mock-HA (`docker-compose.ci.yml`, scenario `ci-growatt-vpp`, VPP settings fixture, clock pinned with libfaketime), reading mock-HA's ordered `/mock/service_log`:

  Activation — `IDLE` period at 08:30:
  ```
  number.set_value      number.growatt_min_vpp_power             1
  number.set_value      number.growatt_min_vpp_time              20
  select.select_option  select.growatt_min_vpp_remote_control    Enabled
  ```
  Release — `LOAD_SUPPORT` period at 10:05:
  ```
  select.select_option  select.growatt_min_vpp_remote_control    Disabled
  number.set_value      number.growatt_min_vpp_power             0
  ```

- Reverted: re-inserted the original arm-first ordering (`_arm(option)` before the release branch, i.e. 30407 written first on both paths).
- Result: **5 of 6 tests FAILED** — both ordering tests, both missing-entity tests, and the power-write-failure test.
- Restored: tree clean, 6 passed.

The sixth (`test_release_does_not_rewrite_the_fallback_timer`) passes before and after — it is a guard against a future regression, not evidence for this fix, and is labelled as such.

Four of the six were also written RED-first and watched fail before the fix existed. One of those first drafts was itself wrong and got fixed: `test_power_write_failure_leaves_remote_control_unarmed` originally failed *every* write, which would not have isolated a power-write failure; it now fails only the power write.

## Outcome-level coverage
**None, and deliberately so** — these assert on service calls, not outcomes. No execution model covers write *order*: `simulation/vpp_simulator.py` derives one `VppCommand(power_pct, remote_control_enabled)` per period and models the resulting energy, so by construction it only ever sees a fully-committed pair and cannot represent an intermediate state such as "armed, power not yet written". The service-call layer is the only place this defect is observable. The module docstring records that reasoning so the next reader doesn't mistake a mapping check for behavioural evidence.

Per `docs/agents/testing.md` this is not a DP/intent/control-mapping change — no plan economics move — so no `R == P` scenario, no new fixture, and no golden/VPP-baseline re-pin is required.

## Documentation
`docs/INVERTER_PLATFORMS.md` gains a "Register write ordering" section — it documented the intent→VPP mapping but never the ordering, which this fix makes load-bearing. `docs/agents/bess-knowledge.md` and `docs/SOFTWARE_DESIGN.md` were both checked; neither describes the VPP write sequence, so neither needed changes.

## Caveat
Writing 30409 while 30407 is Disabled is documented as inert but is **not real-hardware-validated** — same experimental status as #355/#466. Confirmation from a real debug export around a mode switch is still owed, and would also settle whether this ordering defect actually explains the reporter's spikes. The defect stands on the code either way.

Also note the changed failure mode: a failed power/timer write now leaves remote control untouched and the period degrades to `load_first` self-use, instead of executing a stale command.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
+- **Growatt VPP no longer briefly executes the previous period's power command when switching modes** — enabling remote control commits immediately, so the power target is now written before it, and cleared on release. ([#593](https://github.com/johanzander/bess-manager/issues/593))
 - **The battery now covers house load exactly instead of exporting a few Wh and committing the inverter** — a period whose load fell between two discharge steps was planned as a small export, which forces `grid_first` at a fixed rate and imports any load spike. ([#352](https://github.com/johanzander/bess-manager/issues/352))
 - **Huawei, native SolaX and Solis installs no longer report a discharge as negative charging, or an export as negative import** — the signed battery/grid sensor split now works on settings saved before the sensor pairing existed, with no wizard re-run. ([#604](https://github.com/johanzander/bess-manager/issues/604))
 - **The battery schedule no longer varies between otherwise identical runs** — when two plans were worth exactly the same, the optimizer picked between them on floating-point noise, so the same day could plan differently on different machines. ([#606](https://github.com/johanzander/bess-manager/issues/606))
--- a/core/bess/ha_api_controller.py
+++ b/core/bess/ha_api_controller.py
@@ -2361,20 +2361,32 @@ class HomeAssistantAPIController:
     def set_growatt_vpp_period(
         self, remote_control_enabled: bool, power_pct: int, fallback_minutes: int
     ) -> None:
-        """Write one period's VPP command: remote control, power, and fallback timer.
+        """Write one period's VPP command: power, fallback timer, remote control.
+
+        Register 30407 (remote control) is the commit: Growatt VPP has no
+        separate trigger entity, so enabling it executes whatever 30409/30408
+        currently hold. Both are therefore written *before* arming, and the
+        power latch is cleared on release — otherwise the next activation
+        arms against the previous active period's value, which can be ±100%
+        (issue #593).
 
         ``vpp_time`` is rewritten every period the command is active, resetting
         the inverter's own fallback timer (register 30408) — if BESS stops
         writing (crash, restart), the inverter reverts to ``load_first`` on its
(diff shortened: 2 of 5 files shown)


bess-agent:
@claude-bot review

Issue #592, opened 2026-08-15
VPP idle mode


ridax67:
I know this has been discussed in another PR but can't find which.

There was a long period of idle during night but battery was not empty, because there as a plan to cover more expensive morning hours with the remaining energy, or there was much emergy left for morning dump.

We found that battery first 1 is the state that best holds battery level static  which it is.

But it has the drawback that for those cases that idle is used because there is almost no energy left (soc min) the battery and bms is not going to sleep.

This could be fixed by either using load first in those situations instead (idle at soc min) or use power 0 when soc is at soc min.  Both these modes will let the bms sleep when long periods of idle with no energy.

I have tested to forced power to 0 when soc is at min, and I think it is working nicely so recommend this, although not really incorrect as it is.

johanzander:
Thanks — you've found a real gap, and your description matches the code exactly. The PR you were thinking of is #466 (see also #488 / #517).

## What happens today

Whenever the plan says IDLE, we put the inverter in **battery first** — an actively-held command meaning *don't touch the battery, serve the house from grid/solar instead*.

That is correct when the battery is deliberately holding energy back for the morning peak. It is pointless when the battery is empty: there is nothing left to protect, but we still keep the inverter under active remote control all night, so nothing gets to sleep. Worse, because remote control stays enabled we re-send that command every 15 minutes for the entire idle stretch, refreshing the inverter's fallback timer each time — so it is never released.

## Your two options differ in one way that matters: surplus solar

There are three states available while idling:

| Mode | Battery | Surplus solar |
|---|---|---|
| **Battery first** (today) | untouched | still trickles into the battery |
| **Grid first** (your "power 0") | held against charging | **bypasses the battery, goes to grid** |
| **Load first** (your other option) | inverter's own self-use | still absorbed into the battery |

The optimizer's internal model of IDLE explicitly assumes surplus solar *does* get absorbed. So forcing power to 0 quietly breaks that assumption: any idle-at-empty period with sun would export solar the plan expected to bank.

You wouldn't have seen this — you tested overnight, where there is no sun and the two modes are indistinguishable.

I measured how common idle-at-empty actually is across our 37-day test corpus: **it occurs on 24 of the 37 days**, and one winter day is idle-at-empty for all 24 hours. So it is not a corner case, and several of those days do have sun.

## Recommendation: load first (release control) at minimum SoC

[...]

PR #619, opened 2026-08-17
fix: release VPP control when IDLE at the reserve floor (#592)

johanzander:
## Summary

- Growatt VPP: when the plan says IDLE **and** the battery is at its configured minimum SoC, release remote control instead of holding `battery_first`, so the inverter and its BMS can sleep.
- Above the floor, nothing changes — the #466 hold is preserved.

## Root cause

`_intent_to_vpp` mapped every IDLE period to `vpp_power=+1`, remote control **enabled**, with no state-of-charge input at all. Because remote control stays enabled, `_apply_period_vpp`'s `needs_write` is unconditionally true, so the command is re-asserted every 15-minute period to refresh the inverter's fallback timer (#404). Through a long overnight idle the inverter is therefore never handed back, and the BMS never idles down.

The `battery_first` hold is correct where IDLE is *holding energy back* for a later peak (#466): IDLE's own DP cost model (`_idle_battery_flows`) never credits battery discharge for load, so self-consumption must come from grid/solar. At the reserve floor there is no stored energy left to protect — the hold buys nothing and costs the inverter its sleep.

## Fix

`IDLE + at_reserve_floor` → `(0, False)`: released to the inverter's own `load_first` self-use.

`at_reserve_floor` is derived live in `BatterySystemManager._at_reserve_floor()` and threaded through `apply_period` exactly as `block_passive_charging` (#355) and `strategic_intent` (#413) already are. It is read **fresh at every write** — including the retry path (which fires 3–8 minutes later) and the discharge-inhibit path, both of which would otherwise default it to `False` and silently re-assert the hold.

**Why release rather than `vpp_power=0` with remote control still enabled.** The reporter tested the latter and recommended it, but that is `grid_first`, and it is *not* flow-neutral: it holds the battery against charging, so passive solar surplus is bypassed to the grid. IDLE's DP cost model does credit that absorption. Releasing to `load_first` keeps it, matching today's hold exactly. The reporter's overnight test could not distinguish the two — there is no sun at 03:00. Their other listed suggestion (`load_first`) is what this implements.

The SoE conversion deliberately mirrors `min_soe_kwh`'s own arithmetic (`capacity * pct / 100`), because the case that decides this branch is *exact equality* — a battery parked on its floor overnight.

## Test plan

- [x] `pytest -m slow` passes (554 passed, 8 skipped). The VPP regression baseline **did** move, and the movement is itself the evidence — see "Corpus" below. The earlier revision of this PR claimed the baseline was unchanged and read that as flow-neutrality; that claim was vacuous and has been withdrawn.
- [x] **Observed on the real stack** (`docker-compose.ci.yml`, `ci-growatt-vpp` scenario, `libfaketime` at 2025-01-15 20:00, battery sensor driven to the 10% floor). The DP independently planned `Period 80 (20:00): Intent=IDLE` — the reported condition, produced by the real optimizer. Reading the actual inverter registers, with the fix reverted and restored:

  | | `vpp_remote_control` | Meaning |
  |---|---|---|
  | Fix reverted | `Enabled` (power `1`) | inverter commanded, BMS stays awake |
  | Fix restored | `Disabled` | inverter released, BMS can sleep |

  (`vpp_power` still reading `1` afterwards is expected: `set_growatt_vpp_period` documents `power_pct` as ignored when `remote_control_enabled` is False.)

- **Reverted:** `return (0, False) if at_reserve_floor else (1, True)` → `return 1, True`
- **Result:** 3 tests FAILED — `test_idle_at_reserve_floor_releases_remote_control`, `test_idle_at_the_floor_releases_the_inverter`, `test_released_control_stops_re_asserting_every_period`. The production-path test showed the bug directly: four identical `power_pct: 1, remote_control_enabled: True` writes across four consecutive periods, which *is* "the inverter is never released".
- **Restored:** tree clean, 60/60 green.
- The same mutation was also run end-to-end against the live stack (table above), not only against the unit suite.

## Outcome-level coverage

- **Production write path** — `test_vpp_idle_at_reserve_floor.py` drives `BatterySystemManager._apply_period_schedule` and asserts the command that actually lands on the inverter, not `_intent_to_vpp` with hand-built arguments. This is what stops the new branch being unreachable in production while a unit test still passes.
- **Execution model** — `test_vpp_simulator_branches.py::TestIdleAtReserveFloor` proves the released command produces the *same* battery power as today's hold at the floor, under both solar surplus and load deficit, plus a guard rail showing it *does* discharge one kWh above the floor (i.e. why the gate is needed).
- **Corpus** — the VPP baseline, re-derived. The previous claim ("unchanged across 37 fixtures") could not have held: `derive_vpp_commands` never received `at_reserve_floor`, so the release branch was unreachable from the capture harness and the pin could not move whatever the fixtures contained. With the flag threaded through, the corpus exercises it, and the pin moves in exactly the shape the fix predicts:

  | Half | Entries | Periods changed | Δ realized cost | Δ SoE |
  |---|---|---|---|---|
  | v10.0.2 (historical plan, today's model) | 23 | 234 | 0.000000000000 | 0.000000000000 kWh |
  | Current (today's plan) | 27 | 265 | 0.000000000000 | 0.000000000000 kWh |
  | **Total** | **50** | **499** | **0** | **0** |

  499 periods change their command from `[1, True]` to `[0, False]`; not one moves a joule or an öre. That is the corpus-level flow-neutrality evidence the withdrawn claim asserted without support.

  **Re-pin method.** Both pins moved, `test_vpp_execution_of_the_baseline_plan_is_unchanged` included — expected, since this PR edits the execution model that half replays the historical plan through. `--repin-current` skips that half by design, so the historical plans were extracted from the baseline's own `plan` fields and replayed via `--from-plans`, then `--add-new` restored the three post-tag `plan: null` fixtures. No tag checkout. Verified after writing: fixture set identical, **zero plans altered on either half**, only the 50 command arrays re-pinned — so the drift signal `test_drift_from_the_released_version_is_recorded` depends on is intact.

`run_scenario_realized` / R == P does not apply: that harness is Growatt **TOU**-only. `vpp_simulator` is VPP's execution model, and it is covered above.

## Scope assessment

**Local** — stays inside the existing contract. `_intent_to_vpp` is already a pure `(inputs) → (power_pct, remote_control_enabled)` lookup fed by flags computed once upstream in BSM; this adds a fourth flag on the identical, twice-established path. Only one subclass overrides `apply_period`. TOU platforms never reach this code path and ignore the flag.

**Workaround check:** the diff adds nothing whose job is to route around an ordering, timing, or dependency problem. `at_reserve_floor` is a genuine new *input* the mapping needs and does not currently receive — same category as `block_passive_charging`. No fallback, no second construction site, no extra trigger.

## Review round 2 — what changed

Both Stage 4 reviews (09:40 and 10:55) requested changes. All three findings are addressed:

1. **`at_reserve_floor` was dead at 2 of 3 call sites** (09:40). The `= False` default let both non-production callers keep passing while never supplying the real value.
   - `_vpp_display_state` now receives it via `_planned_at_reserve_floor()`, which derives the flag from the *plan's* SoE trajectory (the display counterpart to the write path's live SoC). Without this the schedule API reported the old hold for periods production releases, breaking `_mode_display_fields`'s own no-fabrication contract.
   - `derive_vpp_commands` now threads a per-period flag from the simulated SoE trajectory, which is what makes the corpus evidence above mean anything.
2. **`_at_reserve_floor()` could raise on an unreadable sensor** (10:55, plus the inline comment on `battery_system_manager.py:2816`). `get_battery_soc()` is `float | None`, and the helper multiplied it raw — an uncaught `TypeError` on every period write, on every platform, from two apscheduler jobs with no exception handling. It now reuses `_get_current_battery_soc()`'s `soc is not None and 0 <= soc <= 100` validation and **holds** on an invalid reading, logged. Holding is chosen deliberately: it is the safe direction and is exactly the pre-#592 behaviour, since releasing is what could let self-use draw the battery down. Explicit and logged, not a silent fallback.
   - Writing that test RED surfaced a second bug the review did not mention: a **negative** SoC reading was *releasing*, because a negative SoE is trivially `<= min_soe_kwh`. The range check closes it.

## Known limitation — please read before merging

**How far the battery can actually fall under released self-use is the inverter's own `discharge_stop_soc`, not BESS's `min_soc` — and in VPP mode BESS never writes that register** (`initialize_hardware` returns before `sync_soc_limits`, per #309). If the inverter's own floor sits below the configured `min_soc`, released self-use can draw the gap between them.

This is **not new to this PR**: `LOAD_SUPPORT` (#413) and `SOLAR_STORAGE` already release control the same way at *any* SoC, so this extends an existing exposure to IDLE-at-floor rather than creating one, bounded by the gap between the two floors (zero on a correctly configured inverter). `vpp_simulator` models the release as a hold at `min_soe_kwh`, i.e. it assumes the two floors agree.

Ships experimental pending hardware confirmation from the reporter that (a) the BMS does sleep, and (b) the battery does not discharge below the configured minimum.

## Documentation

- `docs/INVERTER_PLATFORMS.md` — new "IDLE at the reserve floor" subsection under IDLE semantics, including the caveat above.
- `docs/agents/bess-knowledge.md` — the IDLE/VPP mapping paragraph stated the hold unconditionally and would otherwise have gone stale.
- `CHANGELOG.md` — one line under `## [Unreleased] → Fixed`.

Refs #592 — deliberately not `Closes`, since the reporter still needs to confirm on real hardware.


bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

PR #620, opened 2026-08-17
fix: guard git push by shape so feature-branch pushes run unattended

johanzander:
## Problem

The blanket `Bash(git push *)` ask rule stalled every autonomous run — `implement-issue` Step 9 and every `sweep-prs` push. In this session it blocked a routine sweep three times before the work could continue.

`scripts/quality-check.sh` justified the blanket rule like this:

> the dangerous shapes put their marker at an arbitrary argument position … which a prefix glob cannot reach

**That claim is false, and the proof is in the same file.** `matches()` is:

```python
re.fullmatch(re.escape(inner).replace(r"\*", ".*"), command)
```

`*` becomes `.*`, which spans spaces — so `Bash(git push *--force*)` matches `git push origin main --force` perfectly well. The greedy-globs note ~140 lines further down says exactly this. The two comments contradicted each other, and the blanket rule was built on the wrong one.

Why the earlier attempt really leaked: it was **prefix-anchored** (`Bash(git push --force*)`), which genuinely cannot reach argument position 3. The `*marker*` spelling can. That distinction is the whole fix.

## Change

Push is guarded per *shape*: force in any position, refspec `+`, ref deletion, `--mirror`/`--prune`/`--tags`, release tags, and anything naming `main`/`master`/`beta` in either the `origin main` or `HEAD:main` spelling. **A push naming a feature branch runs unattended.**

Branch protection already refuses the case the blanket prompt was standing in for, so it cost an autonomous run and bought nothing.

## Verification

Every push spelling the gate already pinned, plus real branch names from this repo:

| still asks | now unattended |
|---|---|
| `git push origin main --force` / `--force-with-lease` / `-f origin main` | `git push origin HEAD:fix/issue-592-vpp-idle-at-floor` |
| `git push origin +beta-release-9.9` | `git push origin feat/phase4c-charge-commands` |
| `git push origin --delete release-9.9`, `:main` | `git push -u origin fix/issue-604-signed-pair-aliases` |
| `git push origin v9.9.0`, `--tags`, `--mirror` | `git push origin worktree-po-followups` |
| `git push origin HEAD:main`, `HEAD:refs/heads/main`, `master` | `git push origin maintenance-cleanup` ← the `main`-substring trap |
| `git push beta main`, `HEAD:beta`, `beta-release-*-tmp` | `git -C .claude/worktrees/x push origin HEAD:fix/…` |
| bare `git push`, and all of the above via `git -C …` / `git -c …` | |

**0 holes, 0 false prompts.** `./scripts/quality-check.sh` green: *"Permission surface intact (85 command shapes checked, 20 require deny, 20 must stay unattended)"*, `Errors: 0`, full suite 1992 passed / 125 frontend.

The push that created this PR ran without a prompt — the change is demonstrated on itself.

## Gate changes

- `MUST_BE_GUARDED` gains the colon-refspec forms a `* main` pattern cannot see (`HEAD:main`, `HEAD:refs/heads/main`, `:main`, `HEAD:beta`), plus `--force-with-lease`, `-f`, `--mirror`, `--tags`.
- `MUST_NOT_BE_GUARDED` gains the feature-branch pushes and `maintenance-cleanup`. That last one pins the substring trap that would reappear if the protected-ref rules were loosened to `* main*`.
- `REQUIRED["ask"]` no longer names the push rules. 52 pattern names would recreate the presence-check failure mode this gate exists to catch; the command strings hold it instead.

## Deliberately unchanged

**`gh api` keeps its blanket rule.** It is not separable this way: any `-f`, `-F` or `-X` turns a read into a mutation, so no lexical marker isolates the safe subset. If it stalls runs, the fix is an allow-list of read-only paths, not a shape guard.

**`--force-with-lease` still asks.** A lease protects against a concurrent writer, not against wrong local history.

## Docs

`CLAUDE.md`'s Permissions section asserted the opposite in three places ("Every `git push` asks", "guarded bluntly, and that is deliberate", and the `--force-with-lease` rationale). All three are corrected, including the false "prefix globbing cannot reach it" premise.

## Follow-ups not in this PR

- `scripts/quality-check.sh:120` — the mypy check degrades to a `WARNING` when `origin/main` can't be resolved, and warnings exit 0, so it prints `Errors: 0` for a run that type-checked nothing. Found by the #614 review; being fixed there.
- The #617 review noted `gh release download` is claimed unattended in docs but has no `MUST_NOT_BE_GUARDED` string pinning it. Behaviour is correct, verification is missing.


bess-agent:
@claude-bot review

johanzander:
## Prerequisite is now in place — this PR should shrink

`enforce_admins` was **false** on both remotes, which invalidated the premise this PR's protected-ref patterns were being judged against. It is now `true`:

```
origin/main enforce_admins=true
beta/main   enforce_admins=true
```

So `main` is genuinely unpushable — including by the owner token every agent uses. That was the missing piece.

### Why that matters for this PR

Review found **two more holes** (`git push origin refs/heads/main`, `git push --all`), both verified against a real repo. I then found **two more** the review didn't (`git push origin refs/tags/v1.2.3` source-side, and `git push origin HEAD` — which pushes `main` when you are on `main`). That is the third consecutive round of the same enumeration failure, and the review's second checklist point is the real finding:

> it only proves "the strings we thought of are covered" — it does not prove "no valid git-push spelling reaches a protected branch"

Correct. Prefix globs cannot express that property, so enumerating protected-ref spellings will keep leaking.

### Agreed direction

**Delete the protected-ref half of the enumeration** and keep only the shapes that are lexically unambiguous *and* not otherwise prevented:

- keep: `--force` / `-f` in any position, `--force-with-lease`, `+refspec`, `--delete`, `--mirror`, `--prune`, `--tags`, `--all`, `refs/tags/` (both source and destination side), and any push to the `beta` remote
- drop: every `* main` / `*:main` / `*refs/heads/main` pattern and the `master` and `beta`-ref equivalents — GitHub now rejects those server-side, which is a *provable* guarantee rather than an enumerated one

Net effect: a much smaller pattern set, and the half that has failed three times stops existing.

### Still to do here

1. Remove the protected-ref patterns from `.claude/settings.json` and the matching `MUST_BE_GUARDED` strings.
2. Add `--all` and source-side `refs/tags/` to both the `ask` list and `MUST_BE_GUARDED` — these are genuine gaps that branch protection does **not** cover (`--all` pushes every branch; a tag push is not a branch push).
3. Re-point `CLAUDE.md`'s Permissions section at the branch-protection argument instead of the enumeration argument, and record `enforce_admins: true` as load-bearing — if anyone turns it off, this PR's reasoning collapses and the patterns must come back.
4. Re-run `./scripts/quality-check.sh` and re-request review.

The `gh release download` pin and the merge of `main` (#617 conflict) are already in.

bess-agent:
@claude-bot review

bess-agent:
## Status: parked pending a direction decision

**What is on the branch now** (`e9943be6`, CI green, gate at 103 shapes):

The two holes Stage 4 found are closed, plus one it did not:

| Spelling | Before | After |
|---|---|---|
| `git push origin refs/heads/main` (+ master, beta) | hole | guarded |
| `git push --all` / `--all origin` | hole | guarded |
| `git push --branches origin` | hole | guarded |
| `git -C x push origin refs/heads/main` | hole | guarded |

`--branches` is a documented synonym for `--all` (`git push --help`: "--all, --branches"), added in the same pass rather than left to become a fourth leak. All eight new pins were verified as holes against the pre-fix settings and guarded after; `git push origin my-feature`, `git push -u origin fix/issue-620` and `git push origin fix-all-the-things` all stay unattended, the last confirming `*--all*` does not over-match a branch name containing "all".

**Why it is parked.** The comment above proposes the opposite direction for the protected-ref half — delete it entirely, on the grounds that `enforce_admins: true` makes GitHub reject those pushes server-side, which is a provable guarantee rather than an enumerated one. That argument is sound, and the commits currently on this branch go the other way (they *extend* the enumeration). Rather than churn the branch twice, the direction is being settled first.

Verified while assessing it: `enforce_admins` is `true` on both `origin/main` and `beta/main`.

**Open questions the shrink proposal does not yet cover:**

1. **The premise is unchecked.** `enforce_admins` is a GitHub setting, not repo state — untracked, and silently flippable. If it is turned off, the protection evaporates with no signal. Worth asserting it in `scripts/quality-check.sh` so the gate goes red instead, which would turn a load-bearing assumption into a checked invariant.
2. **`beta-release-*` branches on origin are not protected** — branch protection covers `main` only, so dropping the `* beta*` pattern opens a gap that the server side does not backfill.
3. **Tag pushes are not branch pushes.** `git push origin v9.9.0` (no `refs/` prefix) is caught by neither `refs/tags/` patterns nor branch protection.

**#628** tracks the durable fix for the underlying class problem: normalise a refspec's effective destination ref before matching, instead of pattern-matching the raw command string. Three consecutive incomplete rounds is the evidence that enumeration will not converge on its own.

bess-agent:
@claude-bot review

PR #622, opened 2026-08-17
fix: make the review loop actually finish — terminal verdicts, and resume for dead sessions

johanzander:
Two changes, one theme: **the review loop never reached its own endpoint.** Both are described in full below — they touch the same subsystem, and neither is a quiet passenger.

---

# 1. Don't mistake the bot's placeholder for a verdict

`scripts/request-pr-review.sh` took the **last** review newer than its trigger and called it the verdict, regardless of state. But the review bot posts twice:

1. a `COMMENTED` review whose body is literally *"Inline notes below; summary review to follow."*
2. the real `APPROVED` / `CHANGES_REQUESTED` summary, seconds later

Measured on PR #617: **placeholder at `06:57:13Z`, `APPROVED` at `06:58:03Z` — 50 seconds apart.**

Any poll landing in that window returned `COMMENTED`. `implement-issue` Step 11 then saw a non-`APPROVED` verdict and skipped `gh pr ready`, leaving an approved PR as a draft with nothing to do but the merge. `CLAUDE.md` deliberately leaves `gh pr ready` unattended *specifically* so Step 11 can flip it — the loop just never got a verdict it recognised.

**PR #615 proves it**: `CHANGES_REQUESTED` → fixed → `APPROVED` at 21:13, still a draft the next morning.

## Fix + verification

Filter on state **before** taking `last`. Verified against #617's real review history by simulating a poll at `06:57:30Z`, when the placeholder was newest:

| | result |
|---|---|
| **old** | `returns COMMENTED -> Step 11 sees non-APPROVED, skips gh pr ready` |
| **new** | returns empty → keeps waiting → next poll picks up `APPROVED` at `06:58:03Z` |

Both agree once the summary has landed, which is why this never surfaced in normal use — the race only bites when a poll falls in the gap.

## Timeout diagnostics

The timeout path printed one message for two faults needing opposite responses: **a review that started and never summarised** vs **a trigger that never reached the workflow**. It now reports which. **PR #619 is currently the second kind** — two consecutive requests produced no review of any state while #614/#615/#617 all reviewed fine in the same window — and that was invisible before.

---

# 2. Resume an issue whose session died (Step 0)

Sessions die mid-issue routinely, and **nothing picked them up.** A fleet audit found 34 worktrees whose sessions had exited: 8 with real unpushed commits and no PR (one with **32 commits**), plus three PRs green-or-reviewed with no owner. `sweep-prs` refuses the job by design, so the work simply stopped.

## Why here and not a new skill

The loop that acts on review feedback **is Step 11, and already lives here.** A separate skill would duplicate it, and duplicating a review loop is how one of them goes stale.

## How it works

Step 0 keys off state observable from *outside* the dead session and re-enters at the earliest incomplete step:

| Evidence | Dead session got at least to |
|---|---|
| branch or worktree exists | Step 4 |
| commits ahead of `origin/main`, RED test in diff | Step 5–7 |
| open PR with `## Scope assessment` | Step 9 |
| `gh pr checks` green | Step 10 |
| terminal review verdict | Step 11, mid-loop |

**The one thing that dies with the session is Step 2's diagnosis**, which Step 11 depends on holding. It's recoverable *only because this skill already forces it to be written down*: the Stage 2 `analyze` comment, plus the PR body's `## Scope assessment` and `## Test plan`. That's the actual argument for extending this skill — the rehydration sources exist because it mandates them.

When those don't reconstruct a coherent approach, **Step 0 stops** rather than re-diagnosing on top of commits encoding decisions it cannot see.

## Hard rules, each from an observed failure

- **Never** run Step 4's fresh-from-`origin/main` worktree when a branch for the issue already has commits — that deletes them
- **Never** reset or force-push a resumed branch; its commits are the only copy
- Check for a live session unscoped **and unsandboxed**: a sandboxed `claude agents --json` returned **1** session where the truth was **17**, because `~/.claude/jobs` is sandbox-denied and every other session read as dead
- Uncommitted tracked changes are unfinished work, not debris — WIP-commit first
- **If the same issue has died twice, say so and stop.** A second silent relaunch is how a real blocker gets mistaken for bad luck

CI mode gets a Step 0 row: Stage 3 is re-triggered by hand, so a second `@claude-bot fix` on an issue that already has a `has-fix-pr` PR is a **resume, not a restart**, and must not open a second PR.

---

## Verification

`./scripts/quality-check.sh` green on both commits — `Errors: 0`, `Warnings: 0`, permission surface intact. `bash -n` clean.

## Scope note

These are two commits on one branch rather than two PRs. They share a subsystem and a root theme, and both are fully described above — unlike #614, where half the diff was undescribed and orthogonal. Splitting them would need a cross-worktree patch the worktree-isolation guard blocks. Say the word if you'd still rather have them separate.

### DIFF
--- a/.claude/skills/implement-issue/SKILL.md
+++ b/.claude/skills/implement-issue/SKILL.md
@@ -28,6 +28,19 @@ the `bess-analyst` sub-agent.
   inverter/price-provider platforms) — that skill owns experimental→stable
   graduation across multiple beta cycles. Use `implement-issue` for
   single-PR bug fixes and small enhancements.
+- **Also for picking an issue back up after a session died mid-flight.** Same
+  invocation, `/implement-issue <n>`; Step 0 detects the prior work and
+  re-enters at the right step. This is not a separate skill because the loop
+  that acts on review feedback is Step 11 and lives here — a second skill
+  would duplicate it, and duplicating a review loop is how one of them goes
+  stale.
+
+**Sessions die mid-issue routinely, and nothing used to pick them up.** A
+fleet audit found 34 worktrees whose sessions had exited: 8 with real
+unpushed commits and no PR, and three PRs sitting green-or-reviewed with
+nobody left to finish them. #615 was `APPROVED` and still a draft the next
+morning; #614 had `CHANGES_REQUESTED` with no owner to act on it. That is the
+gap Step 0 closes.
 
 ## CI mode (GitHub Actions)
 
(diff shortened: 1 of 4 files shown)


bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

Issue #621, opened 2026-08-17
Setup wizard disables the Huawei platform when auto-detection fails, blocking manual selection for custom/EMMA integrations


bess-product-owner:
Reported by **@valexi7** in [a comment on beta PR #102 (`release: v10.1.0b10`)](https://github.com/johanzander/bess-manager-beta/pull/102), filed here because triage and the board live in this repo.

## Symptom

The setup wizard disables the Huawei tab whenever discovery does not recognise the integration. For a user running **Huawei LUNA2000/EMMA via the Huawei Modbus TLS Server / EMMA Management integration** — rather than the stock `huawei_solar` integration — auto-detection fails, the platform cannot be selected manually, and **setup cannot be completed at all**.

> with my custom/EMMA integration the wizard still disables Huawei when auto-detection fails, so I can't actually complete setup and reach that logic
>
> the selection gate is what currently prevents me from using the setup at all

Screenshot in the original comment: https://github.com/user-attachments/assets/44077d66-155d-4d1f-9233-517d28303f82

## Expected vs actual

- **Expected:** discovery failure narrows the *defaults*, not the *choices*. A user who knows their platform can select Huawei manually and proceed with explicit sensor configuration.
- **Actual:** `huaweiFound === false` disables the Huawei tab outright, so a non-stock Huawei integration has no path through the wizard.

## Why this is the blocker, not the aliasing work

This matters for sequencing. The signed-sensor aliasing fix (this repo's #604 / beta #102) only takes effect **once a platform configuration exists**. The wizard gate runs *before* that, so a user on a custom integration never reaches the aliasing logic. Fixing aliasing alone leaves them stuck.

`@valexi7`'s own [beta PR #100](https://github.com/johanzander/bess-manager-beta/pull/100) contains a fix for the selection gate plus two other Huawei items (EMMA `Time Of Use` working-mode equivalence, and discharge-stop SOC mapping to `storage_discharging_cutoff_capacity`). That PR is **`CONFLICTING` and has not moved since 2026-08-14**.

## Notes for triage

- **Definition of Ready gaps:** no debug log or bundle is attached yet, and there is no reproduction beyond the screenshot. Both are needed before dispatch — the reporter should be asked for a debug export.
- Beta PR #100 is against the beta repo. Per this project's rules, the fix belongs on `origin/main` and reaches beta by fast-forward; #100's content needs porting rather than merging there.
[...]

Issue #542, opened 2026-08-11
Battery Discharge Power sensor missing from Solax Modbus


dontommazo:
My Solax Modbus integration doesn't present the Battery Charge Power and Battery Discharge Power as two desparate sensors, but it is one sensor solax_inverter_battery_power_charge, that goes positive for charging and negative for discharging. 

Do I need to create two custom sensors that separates battery charging power and battery discharging power for BESS Manager? It is not clearly described in the installation guideline.

To simplify installation and configuration, it would be nice if the BESS auto-configuration could work with the standard battery power sensor that is provided by the Solax modbus integration.

johanzander:
@claude-bot analyze

johanzander:
@claude-bot analyze

dontommazo:
Amazing work you do with BESS Manager, Johan! I have now successfully run it in Demo mode for a few days. Because of my battery is so small, only 6kWh, I won't get much benefit from BESS Manager until I upgrade the capacity. 

I have some suggestions. Not sure if this is the right place to communicate them, but here they come:

Because my consumption pattern is a bit random and unpredictable, BESS Managers open-loop power control will not be ideal. It will result in missing load peaks that may occur here and there. Instead of using direct VPP power commands, it would be desirable if BESS Manager could use the "Self Use" mode that are available in the Solax Modbus integration. In that way, the precise and fast control of the inverter DSP can be used to avoid unnecessary import and export. 
To avoid charging up the battery in the expensive morning hours, there is a register available to set a dynamic maximum SOC limit. 

`remotecontrol_target_soc_8_9`

PR #623, opened 2026-08-17
fix: make backlog grooming reflect what actually blocks an issue

johanzander:
## Problem

The digest misclassified enough of the board that grooming couldn't be trusted — and every misclassification pushed the same way: **work looked more ready than it was.**

Measured against the live board (Project #1):

| column | before | after |
|---|---|---|
| Analysis | 13 | **17** |
| Backlog | 13 | 15 |
| **Ready for Dev** | **1** | **0** |
| **In Progress** | **5** | **1** |
| In Review | 4 | 4 |

`Ready for Dev: 0` is the real finding. Nothing in the backlog is dispatchable — everything analysed is blocked or awaiting something. The single item that looked ready was a false positive, which is exactly why dispatching kept failing.

## Five defects, each traced to a real item

**1. `analyzed` was tested before any wait.** So an analysed-but-blocked item reported `Ready`. #96 was labelled `analyzed`, prioritised `P2`, carried no blocking label — and still wasn't implementable, because *how* to build it was undecided. It read as dispatchable, a session was dispatched at it, and that session deadlocked on three design questions with nobody there to answer. **Waits now outrank `analyzed`.**

**2. The board's `Awaiting` field was never read.** 17 items have it set (`discussion` ×11, `reporter` ×6); the digest derived its own value from labels instead. So grooming that had *already been done* had no effect on anything. The field is now authoritative, with `awaiting_source` and `awaiting_suggested` exposed so an unset field gets reconciled rather than silently invented.

**3. `awaiting: discussion` fired on any human comment** — which is not a blocker. Thanks, a "me too", a follow-up question all pushed an item into Analysis. Only a recorded wait or a blocking label does that now.

**4. A worktree left on disk pinned its issue to `In Progress` forever.** #593, #571, #542 and #466 all reported `In Progress` while their PRs (#618, #579, #591, #517) had already merged. Staleness is now decided by comparing the worktree's **own branch** against merged PRs — an exact match, deliberately *not* the fuzzy issue-number match used to associate a worktree with an issue. Those answer different questions and conflating them is what caused defect 6 below.

**5. `blocked` didn't fail Ready.** #571 reported `Ready for Dev` while carrying the `blocked` label. Definition of Ready criterion 5 now holds, including an unresolved `Blocked by #N`.

Plus: **`Ready for Dev` finally requires a `Priority`** — which the design always specified and the code deferred *"until a board exists"*. It exists, and every item carries P1–P4.

## New signal: who spoke last

`last_comment: {author, days, is_reporter, is_bot}`.

Without it the digest could not represent the transition that matters most to grooming — **the reporter answering us**. A comment *count* and a last-activity *date* cannot distinguish "the reporter attached the debug log we asked for" from "we posted a nudge" from "the reporter asked something new". All three just increment a number. #621 crossed the Definition of Ready line and nothing noticed.

## A change I made and then rejected

I initially mapped a merged PR to `Done`. It reclassified **7 open issues as finished** (#118, #120, #403 among them), and it contradicts this project's own rule that beta PRs omit `Closes #N` until graduation — so **an open issue with a merged fix is the normal state**, not a completed one. `merged_pr` is now reported for information, moves no column, and a test pins that.

## Verification

`./scripts/quality-check.sh` green — `Errors: 0`, `Warnings: 0`. **25 tests pass** (was 18), including one per defect above and both halves of the stale-worktree rule, so the fix can't hide live work.

Three test-harness bugs fixed along the way, each of which had been masking real behaviour:

- The `gh` shim matched on `$1 $2`, so both `pr list` calls returned the **open** list — every open PR would have looked merged. It now branches on the full argument string.
- Comment fixtures had no `createdAt`, which real `gh` always sends. Their absence failed `strptime` instead of testing anything.
- `_run` used `check=True`, so a jq failure surfaced as a bare `returned non-zero exit status 5` with the actual error swallowed. It now reports the digest's stderr.

One implementation note: merged PR bodies are reduced to their closing references *before* reaching jq — passing 200 of them through `--argjson` overflows the argument list (`/usr/bin/jq: Argument list too long`).

## Not in this PR

**Nothing runs any of this yet.** The design (`docs/superpowers/specs/2026-08-15-backlogger-agent-design.md:86-91`) puts follow-up on a local `/loop /backlog` "Rhythm" surface and predicted this exact failure — *"If reports start going stale, that is the signal to move Rhythm to an Actions cron"*. That condition has fired. Scheduling it is the remaining half and is deliberately separate, since this PR is what makes the signals worth acting on.

### DIFF
--- a/.claude/skills/backlog/SKILL.md
+++ b/.claude/skills/backlog/SKILL.md
@@ -20,20 +20,28 @@ Open an individual issue only when you are deciding about that issue.
 
 ## Prerequisites
 
-Board reads need `PROJECT_NUMBER` set and the board created (deferred —
-`scripts/backlog-board-init.sh`). Board writes need `BESS_PO_TOKEN` with
-`project` scope (also deferred). Until both exist, a pass fails loudly at the
-first board access in `backlog-digest.sh` — that failure is expected, not a
-bug to route around.
+The board exists: **Project #1, "BESS Manager Backlog"**. `PROJECT_NUMBER=1`
+lives in the repo's `.env`, which `backlog-digest.sh` does **not** source — so
+export it first, or the digest exits with a "board has not been created yet"
+message that is misleading rather than wrong:
 
-**One thing to verify the first time a board exists.** No board has ever
-existed, so the JSON shape `gh project item-list --format json` uses for a
-custom field is unconfirmed; the digest assumes each item carries a top-level
-`.priority`. On the first real run, check that `priority` is populated rather
-than `null` for an item you have set a priority on. If it is `null`, fix the
-jq path in `scripts/backlog-digest.sh` — do not add a fallback that tries
-several shapes. A silently-null `priority` disables ranking axis 2 without
-any error.
+    set -a; . ./.env; set +a; ./scripts/backlog-digest.sh
+
+Board writes need `BESS_PO_TOKEN` with `project` scope. The custom-field JSON
+shape is confirmed: `gh project item-list --format json` puts each single-select
+value at the item's top level, so `.priority` and `.awaiting` both read directly.
+
+Field options, as they actually exist — do not invent values outside these sets:
+
+| Field | Options |
+|---|---|
+| `Status` | `Backlog`, `Analysis`, `Ready for Dev`, `In Progress`, `In Review`, `Done` |
+| `Priority` | `P1`, `P2`, `P3`, `P4` — **there is no `P0`** |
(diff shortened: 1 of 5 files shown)


bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

Issue #624, opened 2026-08-17
Beta 10 never goes through initialisation


ridax67:
[bess-debug-2026-08-17-115354.md](https://github.com/user-attachments/files/31137949/bess-debug-2026-08-17-115354.md)
Stuck on initialisation.

johanzander:
@claude-bot analyze

Issue #578, opened 2026-08-14
feature: comfort / heatwave reserve — suppress Battery Export without raising Min SOC (announced high evening load)


Frank-Leysen:
﻿## Context (live Growatt test, 14 Aug 2026)

Hottest day of the year here (~39 °C). Frank needs **aircon ~21:00–00:00**, but BESS planned **Battery Export ~19:15–20:45** down toward ~44% SOC — leaving too little for cooling.

Raising **Min SOC to 45%** was the wrong tool: it stops *all* on-grid discharge at 45%, so AC would hit the floor and pull from the grid. What we actually need is: **skip / cap Battery Export tonight**, while **Load Support** (house + AC from battery down to normal Min SOC) remains allowed.

Tonight’s workaround: let export run until **~20:15**, then manually force **Load First** + keep Min SOC **12%**.

There is already a mapped `discharge_inhibit` helper (`hoog huisverbruik — geen avondexport`) on this install, but if that blocks *all* discharge it is too coarse for this case.

## Feature ask

A first-class control for **announced high evening load / comfort reserve**, e.g.:

- Toggle or mode: **“No Battery Export for the next N hours / until time T”** (or “heatwave / high home load”)
- Effect: suppress or heavily cap **BATTERY_EXPORT** / Grid First sell
- **Do not** raise Min SOC as the mechanism
- **Do** allow **LOAD_SUPPORT** so the battery still covers house + AC down to the normal Min SOC
- Ideally visible in the schedule UI so the plan recalculates (not only a last-minute inverter override)

Related existing threads: overnight reserve floor (#381), risk-aware overnight buffer (#418), consumption forecast strategy (#570), external series seam (#428).

## Design question (internal vs external load planning)

This sits in the larger product question: **should non-baseline consumption be planned inside BESS, or supplied externally?**

| Approach | Fit for tonight |
|---|---|
| **A. In-BESS announced overlay** | User/HA says “AC ~2 kW from 21:00–00:00” (or heatwave flag). Optimizer treats it as known load → less evening sell, more SOC at 21:00. Same *shape* as EMHASS “deferrable / thermal comfort” *inputs*, without requiring a parallel EMHASS stack. |
| **B. External forecast series (#428)** | EMHASS (or other) publishes a 96-point load series including AC; BESS only optimizes the battery against that series. |
| **C. Manual inhibit (tonight)** | Boolean / override — blunt but shippable as a v1 of the comfort switch. |

[...]

PR #626, opened 2026-08-17
fix: make a backlog pass runnable without a shell preamble

johanzander:
Three fixes to the Product Owner tooling, all found by actually running `/loop 30m /backlog` against the live fleet. Each had the same shape: **the surface asserted a state it had not checked**, and every failure pointed the same way — toward acting on work that was already handled.

---

## 1. A backlog pass could not run at all

`/loop /backlog` died on its first line, every tick:

```
backlog-digest.sh: PROJECT_NUMBER is not set — the backlog board has
not been created yet. Run scripts/backlog-board-init.sh (deferred) to
create it, then set PROJECT_NUMBER.
```

Every clause of that is wrong. The board is **Project #1, populated, 37 cards**. `PROJECT_NUMBER=1` lives in the gitignored `.env`, which nothing exports, and `backlog-board-init.sh` does not exist because the board was created by hand.

The skill documented a workaround — `set -a; . ./.env; set +a; ./scripts/backlog-digest.sh` — but a workaround needing a shell preamble cannot survive an unattended pass, because the caller is a skill, not a shell someone typed into.

**Fix:** the digest sources `.env` itself, located via `git rev-parse --path-format=absolute --git-common-dir`. That is one rule, not a search over candidate paths: `.env` is gitignored, so it exists only in the main checkout and never in a worktree, and the common dir resolves to the main checkout from either. `--path-format=absolute` is load-bearing — git otherwise answers the relative `.git`, whose dirname is `.`, silently resolving against the caller's cwd. The environment still wins over the file, or a pinned `PROJECT_NUMBER=2` would be un-overridable.

### Two gaps this exposed

Both were things the `board` verb needed and had to recover with a second `gh project item-list` call by hand — the one thing the digest exists to stop.

**`board_status`** — where the card sits *now*, next to `column` (where the evidence says it belongs), in the same six Status strings. The skill says "reconcile every card against the derived column, the digest always wins" and the digest carried nothing to reconcile *against*. Seven live mismatches were invisible: #602, #593, #592, #578, #571, and #621/#624 which had no card at all.

**`issue_no_card`** — an open issue with no card. `Ready for Dev` requires a Priority, and Priority is a board field, so an off-board issue can **never** become dispatchable however well analysed, while reading as an ordinary Backlog item.

---

## 2. An unreviewed PR was reported as ready to merge

The rhythm resolved *any* non-draft PR to `awaiting_maintainer`, "nothing left but your merge". It never looked at reviews.

That rests on an assumption which does not hold: that only Step 11 clears the draft flag, and only after an APPROVED verdict. **This PR broke it.** It was flipped out of draft by hand because it *looked* stuck, had zero reviews at the time, and the next pass duly reported it as ready to merge — routing straight around Stage 4, the gate the whole pipeline is built on. A Stage 3 CI-mode PR has the same hole from the other side: it never runs Step 11 at all.

**Fix:** an approving review is the bar, matching Step 11. Two new actions carry what used to collapse into "merge it": `request_review` (Stage 4 never ran) and `rework_review` (changes requested). Both fields were already fetched by `gh pr list` and simply unused.

---

### 3a. Review state — neither signal alone is correct

The first attempt at fix 2 asked "is there an APPROVED review anywhere" *before* consulting `reviewDecision`. **The Stage 4 review caught this and reproduced it**, correctly: GitHub never rewrites an old review when a later round requests changes, so an approved-then-reworked PR keeps its stale APPROVED entry forever and was reported "nothing left but your merge" — the exact failure fix 2 set out to close, reintroduced by the fix for it.

The review proposed keying on `reviewDecision` instead. **That alone is also wrong here, and measurably so:** `reviewDecision` is only populated when the repo *requires* reviews, and this one does not. It reads `CHANGES_REQUESTED` for #619/#620/#614 but `""` for #490, which carries two genuine APPROVED reviews — so keying on it alone reports an approved PR as never reviewed.

Neither signal is sufficient and each fails toward "merge it", so **the order is the whole content of the rule**: trust `reviewDecision` when set, otherwise fall back to the **last** non-COMMENTED review. Last, not any — same staleness trap. COMMENTED is skipped because the bot posts inline notes as a COMMENTED review before its real verdict (see `request-pr-review.sh`).

### 3b. Session liveness — `claude agents` cannot answer the question

`resume_implementation` keyed off `session == null`, and `session` comes from `claude agents`, which lists **background agents only**. Two independent failures:

1. A session started in the terminal — `claude`, then `/implement-issue <n>` — is a foreground session and never appears at all. That is how #624 was dispatched.
2. Even a background agent carries a generated descriptive name (`"Review PR and create branch for bess-manager"`), not the `issue-<n>` the dispatch convention promises, so the exact-name match misses it too.

**Measured: 41 worktrees on disk, `claude agents --json` returning one entry.** So `session` was null for essentially everything, every worktree read as abandoned, and `resume_implementation` fired on live sessions — advising a second session onto a branch its own detail line calls *the only copy*.

**Fix:** the worktree **lock** is what tracks a live session. Git records it as `locked claude session <name> (pid 97626 start ...)`, written by Claude Code itself. 4 of those 41 were locked — exactly the four live sessions, foreground and background alike. It is local and needs no process list. `session` is kept because a name match is strictly more informative when it does happen; it is just no longer what liveness rests on.

---

## Live effect on the current fleet

| Item | Before | After |
|---|---|---|
| #624 (being worked right now) | `resume_implementation` — "no live session" | silent; worktree is locked |
| #626 (this PR) | `awaiting_maintainer` — "nothing left but your merge" | `rework_review` |
| #490 (genuinely approved, `reviewDecision: ""`) | `awaiting_maintainer` | `awaiting_maintainer` |
| #621, #624 (no card) | invisible | `add_card` |

## Verification

- `./scripts/backlog-digest.sh` and `./scripts/backlog-rhythm.sh` both run end-to-end from a clean shell with nothing exported.
- **2052 passed**, 50 skipped (`pytest -m "not slow"`); `./scripts/quality-check.sh` green.
- 20 new tests, including the reviewer's exact reproduction (`test_a_stale_approval_does_not_survive_a_later_changes_requested`) and the case their proposed fix would have broken (`test_an_approval_is_honoured_when_review_decision_is_empty`).
- The digest's git shim answers `rev-parse` so `.env` discovery is deterministic — letting the real git through would read the maintainer's own gitignored file, which CI does not have. The porcelain fixture uses the real `locked <reason>` shape, not the bare keyword.

## Scope

Four files, all inside the domain these two scripts already own: `scripts/backlog-digest.sh`, `scripts/backlog-rhythm.sh`, their two test files, and `.claude/skills/backlog/SKILL.md` (the stale `.env` preamble and the changed `awaiting_maintainer` semantics). No product code touched, so no CHANGELOG entry — this is agent tooling with no user-visible effect.

## Note for reviewers

The jq program is a single-quoted shell string, so **one apostrophe in a comment ends it early** and jq reports the useless `Top-level program not given`. That cost a debugging round here; there is now a comment saying so.

### DIFF
--- a/.claude/skills/backlog/SKILL.md
+++ b/.claude/skills/backlog/SKILL.md
@@ -204,7 +204,9 @@ Actions, and who does what:
 | Action | Do |
 |---|---|
 | `resume_implementation` | `/implement-issue <n>`. **The action that produces a ready PR** — Step 11 requests the review, acts on the verdict and runs `gh pr ready`. Covers a draft needing a first review, a rework, an approved PR that never got flipped, *and* a worktree whose session died |
-| `awaiting_maintainer` | nothing; report it. Out of draft is the finish line |
+| `awaiting_maintainer` | nothing; report it. Out of draft **and carrying an APPROVED review** is the finish line |
+| `request_review` | out of draft but Stage 4 never ran. `scripts/request-pr-review.sh <n>`. **Never report an unreviewed PR as ready to merge** — the draft flag is not a review, and a maintainer who flips it because a PR looks stuck routes around the one gate the pipeline is built on |
+| `rework_review` | out of draft with changes requested: `/implement-issue <n>` to address them, then a fresh review |
 | `resolve_conflict` | hand to `sweep-prs` |
 | `recheck_ready` | the reporter answered: re-check Definition of Ready, clear `Awaiting` if satisfied |
 | `nudge_reporter` | one nudge, as the PO identity |
--- a/.claude/skills/backlog/SKILL.md
+++ b/.claude/skills/backlog/SKILL.md
@@ -22,11 +22,17 @@ Open an individual issue only when you are deciding about that issue.
 ## Prerequisites
 
 The board exists: **Project #1, "BESS Manager Backlog"**. `PROJECT_NUMBER=1`
-lives in the repo's `.env`, which `backlog-digest.sh` does **not** source — so
-export it first, or the digest exits with a "board has not been created yet"
-message that is misleading rather than wrong:
+and `BESS_PO_TOKEN` live in the repo's `.env`, and `backlog-digest.sh` sources
+it itself — **run it plainly, with no `set -a` preamble**. The file is
+gitignored and so exists only in the main checkout; the script resolves it via
+`git rev-parse --git-common-dir`, which points there from inside a worktree
+too.
 
-    set -a; . ./.env; set +a; ./scripts/backlog-digest.sh
+An earlier version required the caller to export those first, which is why an
+unattended `/loop /backlog` pass died on its first line every time: the thing
+invoking the script is a skill, not a shell someone typed into. If the digest
+now exits complaining about `PROJECT_NUMBER`, the variable is genuinely absent
+from `.env` — do not work around it by exporting one.
 
 Board writes need `BESS_PO_TOKEN` with `project` scope.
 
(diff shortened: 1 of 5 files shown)


bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

PR #631, opened 2026-08-17
fix: make a resumed PR read its comments, not only its reviews

johanzander:
## Summary

- `implement-issue` Step 0 now fetches a resumed PR's **conversation comments** alongside its reviews, and states that a maintainer comment outranks every bot review on the PR.
- Step 11 re-checks comments **each round**, not only at Step 0.
- Two Rationalizations rows and two Red Flags name the failure mode directly.

## Root cause

Reviews and conversation comments are two different feeds, and `gh pr view --json reviews` returns only the first. Step 0's rehydration list read:

> - the diff itself, and any inline review comments

It never mentioned the conversation. So a session following the skill *exactly* still missed the highest-authority input on the PR.

The asymmetry is structural rather than incidental: **a review has to attach to a diff.** "Do this differently", "branch protection changed, so this whole approach should shrink" — none of that can be expressed as a review, so it lands as a comment. The feed the skill ignored is the one carrying direction.

## What it cost

On #620 the maintainer posted **"this PR should shrink"** with a four-step plan: `enforce_admins` is now `true` on both remotes, so GitHub rejects pushes to `main` server-side and the protected-ref *enumeration* should be **deleted** rather than extended.

A later session read the reviews, did not read the comments, and spent a full rework **adding** protected-ref patterns — the exact opposite of a standing instruction sitting in that PR's own thread. Two further holes the maintainer had already found by hand (`git push origin refs/tags/v1.2.3`, `git push origin HEAD`) were named in the same comment and stayed open.

The work was CI-green and internally verified. Correctness checks cannot catch this class of error, because the diff was not wrong — it was pointed the wrong way.

## Fix

| Location | Change |
|---|---|
| Step 0 query | `--json ...,reviews,comments`, with a note not to drop it |
| Step 0 rehydration list | conversation comments as an explicit item, both `gh` commands side by side, and the precedence rule |
| Step 11 verdict handling | re-check comments each round — a direction posted mid-loop is invisible to the verdict feed, and the bot will keep approving a diff the maintainer asked you to redo |
| CI-mode table | same note; it matters *more* there, since Stage 3's re-trigger **is** a comment, so the reason is usually in the same thread |
| Rationalizations / Red Flags | "I read the reviews, so I know what this PR needs" and "the bot approved it, so the direction must be fine" |

## Test plan

- [x] Verified the underlying claim against the live API rather than assuming it: `gh pr view 620 --json reviews` returns only the bot's review; the maintainer's direction comment appears solely under `--json comments`.

Not applicable — this is a prose change to a skill file, with no code path and nothing to redden. Stating that rather than filling the section with a suite result, per the skill's own instruction.

The empirical check that *does* discriminate is the one above: the two feeds were queried separately against #620 and confirmed to return different content, which is the whole premise of the change.

## Outcome-level coverage

None, because this is a skill document. The behaviour it governs is a model's reading order, which no fixture pins. The mitigation is redundancy — the instruction now appears in the Step 0 query, the rehydration list, Step 11, the CI table, two Rationalizations rows and two Red Flags, so a session skimming any one section still meets it.

## Scope

One file, `.claude/skills/implement-issue/SKILL.md`. No CHANGELOG entry: agent tooling, no user-visible effect.

### DIFF
--- a/.claude/skills/implement-issue/SKILL.md
+++ b/.claude/skills/implement-issue/SKILL.md
@@ -58,7 +58,7 @@ runners — only repo-level `.claude/skills/` and `.claude/agents/` exist there.
 
 | Step | CI mode |
 |---|---|
-| 0. Resume check | Applies, and matters more here: Stage 3 is re-triggered by hand, so a second `@claude-bot fix` on an issue that already has a `has-fix-pr` PR is a resume, not a restart. Detect the existing PR and continue it — never open a second PR for one issue. The CI checkout has no worktrees, so branch existence on `origin` is the only signal available. |
+| 0. Resume check | Applies, and matters more here: Stage 3 is re-triggered by hand, so a second `@claude-bot fix` on an issue that already has a `has-fix-pr` PR is a resume, not a restart. Detect the existing PR and continue it — never open a second PR for one issue. The CI checkout has no worktrees, so branch existence on `origin` is the only signal available. Reading the PR's **conversation comments** matters more here too, not less: the re-trigger is itself a comment, so the maintainer has very often said *why* in the same thread. |
 | 2. Diagnose | Stage 2 comment absent → STOP. Post "No deep analysis found. Run `@claude-bot analyze` first" and exit — never self-diagnose in CI; the analyze/fix split *is* the human gate. |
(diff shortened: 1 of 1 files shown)


bess-agent:
@claude-bot review

PR #633, opened 2026-08-17
fix: stop planning a sub-floor solar export the inverter will absorb (#630)

### LINKED ISSUE #630: Plan/execution divergence on sub-grid-step IDLE solar charging (impact negligible — canary, not lost savings)
## Impact: negligible — read this before prioritising

**No user will ever see this in their savings.** Measured at **+0.0016 SEK** on a 96-period day whose optimized result is −47.82 SEK: about **0.003% relative error**, order 0.6 SEK/year if it recurred every day. The dashboard overstates predicted savings by a fraction of an öre.

It is filed because it is a **plan/execution divergence**, not because of its size. The corpus gate that caught it is zero-tolerance for a reason: this same class was 3.57 SEK across 33 fixtures before #497 fixed it, and the gate exists so the *mechanism* cannot hide behind a small *number*. Treat this as a canary and a correctness record, not as lost savings.

**Not measured:** the worst case. This is one observed day. The drift accumulates across IDLE periods with sub-grid-step surpluses and only converts to cost when the battery reaches a bound, so a day with many such surpluses and repeated bound-hits could be larger — but there is no evidence it ever becomes material, and that should be bounded properly (`bess-analyst`) rather than extrapolated from a single fixture before anyone spends time on a fix.

---

Found while implementing #624 (PR #629). Not caused by that change — see the control experiment below. Filed separately because it is a distinct defect with its own cause and its own fix.

## Symptom

The DP plans a cost its own execution does not deliver, breaching the corpus-wide zero-tolerance `R == P` gate in `test_realized_matches_planned_across_all_fixtures`.

## Root cause

Passive solar charging below one SOE grid step is invisible to the DP's snapped trajectory but real to the hardware.

At period 32 of `regression_2026_08_17_624`:

- solar 0.0990 kWh, consumption 0.0956 kWh → surplus **0.0034 kWh**
- `SOE_STEP_KWH` is **0.025** — the surplus is an order of magnitude smaller, so the DP's snapped SOE trajectory cannot represent absorbing it and plans it as export
- the command the DP derives for IDLE is `ControlCommand(battery_mode='load_first', discharge_rate_pct=0, grid_charge=False, charge_rate_pct=100)` — which makes the inverter absorb the surplus into the battery anyway

```
period 32  intent=IDLE
    battery_charged      P=   0.00000  R=   0.00341   <<<
    grid_exported        P=   0.00341  R=   0.00000   <<<
    solar_to_battery     P=   0.00000  R=   0.00341   <<<
```

The battery then runs 0.0033 kWh fuller than planned. That is costless for 32 periods, until period 64, where the battery reaches `max_soe`:

```
period 64  intent=SOLAR_STORAGE
    battery_charged      P=   0.80774  R=   0.80433   <<<
    grid_exported        P=   0.53748  R=   0.54089   <<<
```

Being fuller leaves that much less room for solar, which is exported instead — and the two periods' cost deltas (+0.006321 at 32, −0.004766 at 64) sum to exactly the observed +0.001555.

## Not caused by #624's window bisection

Measured, not assumed. With `detect_tie_windows` patched to return no windows — no PWL solve, no splice, no bisection — the gap is bit-identical:

```
windows ON : R=-47.814616 P=-47.816171 gap=+0.001555
windows OFF: R=-47.800070 P=-47.801625 gap=+0.001555
```

Two reviewers independently attributed the gap to the bisection seam during PR #629's review. The seam's own measured contribution is 0.000000 SEK.

## Current state

Recorded in `KNOWN_PLAN_EXECUTION_GAP_SEK` (`core/bess/tests/integration/test_plan_faithfulness.py`) with this evidence, so the corpus gate stays meaningful for everything else.

That entry is self-invalidating rather than a silencer: both the corpus test and `test_the_bisected_schedule_is_executable_as_planned` assert the gap *equals* +0.0016, so fixing this issue reddens the suite and forces the entry's deletion. It cannot rot unnoticed.

Note that adding it re-opened a table the header comment describes as collapsed by #497 ("any future gap is a real finding, and there is no band for one to hide in"). That was a deliberate, documented concession made in PR #629 — flagged here so the next reader sees it rather than discovering it.

## Suggested direction

Either credit passive charging below one SOE step in the DP's IDLE flow model, or stop snapping the reported trajectory where the physical transition is representable. Deciding between those is an optimizer-architecture call (P4 — the flow record must not be re-derived), not something to pick during a fix.

johanzander:
## Summary

- Withhold the SOLAR_EXPORT-below-max bypass candidate wherever the intent classifier would not actually label the period `SOLAR_EXPORT`, so the DP stops planning an export the inverter absorbs.
- Charge-side twin of the `_residual_cover_p` gate that already closes the same failure on the discharge side (#282/#497), on the same constant.
- Corrects the reported plan only. Realized cost is unchanged to the bit — the plan came to meet reality, reality did not move.

## Root cause

The issue attributes the gap to `SOE_STEP_KWH` grid-snapping making a sub-step absorb unrepresentable. That is not the mechanism: the forward replay carries continuous SoE, so absorbing 0.0034 kWh is perfectly representable, and period 32's SoE is held *bit-exactly*, which snapping would not produce.

What actually happens is that the DP deliberately offers a second IDLE-shaped action — the SOLAR_EXPORT-below-max bypass (#313), `consider(0.0, forced_next_soe=soe)`: hold SoE, export this period's own surplus. Nothing commands that hold directly. It is delivered by `classify_strategic_intent` labelling the period `SOLAR_EXPORT`, which is the intent that writes charge rate 0 and so stops the inverter absorbing the surplus.

That label requires `grid_exported > FLOW_NOISE_FLOOR_KWH` (0.01 kWh). At or below it the period falls through to `IDLE`, whose command is `load_first` at charge rate 100 — which absorbs. So the plan books export revenue on energy the hardware puts in the battery, and the battery runs fuller than planned until it reaches a bound and spills the difference.

At period 32 of `regression_2026_08_17_624` (surplus 0.0034 kWh, export below the floor), the battery then runs 0.0033 kWh fuller than planned for 32 periods until period 64, where it hits `max_soe` and exports that much solar instead. The two periods' cost deltas sum to the observed +0.001555 SEK.

`GRID_FLOW_RESOLUTION_KWH` (0.1) is **not** a second band here, despite being what the discharge-side twin uses: `EnergyData`'s noise fold applies only to `battery_to_grid` when the battery is discharging (`models.py:193`). Solar export is never folded, so the classifier's floor is the only binding threshold.

## Fix

`_solar_export_bypass_is_unexecutable()` in `action_selector.py` — a per-period scalar (with no battery flow, `_ac_flows` reads only solar, load and the AC cap, so the export does not depend on SoE). The candidate is withheld at all three sites that define the action set (P1): the shared selector, and both backward passes.

The grid backward pass's existing *"deliberately NOT masked by `_discharge_is_unexecutable`"* exception does not transfer, and the code says why: that exception rests on the coarse lattice having no exact-cover point near the deficit breakpoint, so an in-band action is the better proxy. The bypass is an exact action, not a lattice approximation, and its executable neighbour (plain IDLE) is in the action set at every state.

Withholding is always safe: plain IDLE is offered unconditionally, and wherever the bypass diverts nothing (no surplus, or a battery already at `max_soe`) the two candidates coincide exactly.

**Honest note on the two backward-pass gates: they are inert on today's corpus.** Reverting them alone leaves every test green, including the new one. They are there because P1 requires one action-space definition — leaving those columns valuing an uncommandable action is the #236 class — not because they were measured to change a decision. Stated here rather than left for the reviewer to discover.

Declined one review suggestion: `_solar_export_bypass_is_unexecutable` re-derives the AC cap even though all three call sites hold it. Accepting it as a parameter is what `_period_flows`' docstring warns against — a cap a caller can get wrong is the reward-vs-flows divergence P4 exists to remove. One derivation per period is not worth reopening that seam.

## Test plan

- [x] `./scripts/quality-check.sh` passes locally (2052 passed / 50 skipped; Black, Ruff, TypeScript, ESLint, permission surface all clean)
- [x] `pytest -m slow` passes (549 passed, 7 skipped)
- [x] Every plan pin unchanged across the 32-fixture corpus: `test_current_plan_is_pinned` (bit-exact actions, intents, SoE trajectory), `test_all_scenarios`, the VPP baselines. **Zero golden churn** — consistent with a corpus scan finding exactly one affected period.
- [x] **The real fixture, observed.** `regression_2026_08_17_624` (the user-derived quarterly day) run through the real optimizer and the real inverter simulator, with PR #629 layered on locally so it can run at all:

  ```
  R=-47.814616  P=-47.814616  gap=+0.000000 SEK      (issue reports +0.001555)

  period 32  intent=IDLE
      soe        19.31579 -> 19.31910     (was: held at 19.31579)
      charged     0.00341                 (was: 0.00000)
      exported    0.00000                 (was: 0.00341)

  periods still planning a sub-floor held-SoE export: []
  ```

  R is unchanged from its pre-fix value — the entire gap was the plan overstating, and it closes exactly, so no second mechanism is hiding underneath. The two PRs also apply cleanly together.
- [x] **App-level E2E** via `docker-compose.ci.yml` + mock-HA (`ci-growatt-vpp`, libfaketime-pinned clock): backend healthy, optimizer ran end-to-end, served a full 96-period schedule at 11.36 SEK predicted savings. **This confirms no regression in the service path; it does not exercise the bug** — that scenario is a winter day with zero solar in all 96 periods. The bug-specific observation is the fixture run above.

Two mutations, both run:

- Reverted: the `select_action` gate only (`if not _solar_export_bypass_is_unexecutable(...)` → `if True:`), leaving both backward-pass gates in place.
- Result: `test_subfloor_solar_export_is_never_planned` **FAILED** — `R=2.404206 P=2.401236 gap=+0.002970`, 3× `PLAN_EXECUTION_TOLERANCE_SEK`. 1 failed, 8 passed in that file.
- Reverted: the two backward-pass gates only, leaving the selector gate.
- Result: **all 134 tests passed** — which is the basis for the "inert on today's corpus" note above, not something discovered afterwards.
- Restored: tree clean, full suites re-run green.

The test was also written RED before the fix existed and observed failing at that same gap.

One bug the mutation work caught: the first version of the `dp_battery_algorithm.py` gate used `continue`, which would have skipped the residual-load-cover candidate later in the same loop iteration — harmless on a surplus period, a real defect on any period with a deficit. Restructured as an `if` block.

## Outcome-level coverage

- `R == P` via `run_scenario_realized` in `test_subfloor_solar_export_is_never_planned` — asserts realized cost reproduces planned cost, plus a secondary mechanism assertion that period 0 plans no sub-floor export so a future regression is diagnosable rather than merely red.
- The scenario is built so the bypass is the DP's *honest* economic choice, not an accident: the 0.0099 kWh surplus sits just under the floor; periods 1–2 fill the battery to `max_soe` regardless, so storing period 0's surplus is worth only the export it displaces (sell 0.2) rather than its evening value; selling at 0.5 therefore beats storing; and a larger discharge-and-refill is unprofitable at `0.5 sell − 0.4 cycle − 0.2 displaced`. `terminal_value_per_kwh` is pinned to 0.0 so the leftover-SoE term cannot quietly change which candidate wins.
- Existing corpus pins (goldens, `test_all_scenarios`, VPP baselines) cover the no-churn half.

## Documentation

`docs/agents/bess-knowledge.md` describes this exact bypass under "Deferring absorption uses the SOLAR_EXPORT-below-max bypass (#313)", and claimed unconditionally that such periods classify as `SOLAR_EXPORT`. That is now conditional, so the section records the executability bound. `docs/SOFTWARE_DESIGN.md`'s classifier table (`grid_exported > 0.01 kWh`) and intent→mode table are both still accurate and unchanged.

## Interaction with PR #629

The fixture that exhibits this bug lives in #629, which is still open, so this branch's regression test is a hand-built scenario on `main` instead — it does not depend on #629 merging.

**Whichever of the two merges second must delete #629's `KNOWN_PLAN_EXECUTION_GAP_SEK` entry for `regression_2026_08_17_624`**, since this fix takes that gap to exactly 0.000000 and the entry would then pin a gap that no longer exists. Verified above by running the two together.

Closes #630


bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

PR #636, opened 2026-08-17
feat: put PRs on the board so a decision about one is recorded, not re-argued

johanzander:
## Problem

The board held **issues only**, so every judgement about a *pull request* had nowhere to live.

These were all real decisions the maintainer had already made:

- *"#167 and #354 are blocked in my book"*
- *"#437 is a draft I intend to work on later but has lower priority, same with #490"*
- *"#620 is parked pending a direction call"*

Nothing recorded any of them, so `backlog-rhythm.sh` re-reported all of them as due on **every 30-minute tick** — and the same conversation happened every 30 minutes. The pass was structurally incapable of remembering an answer.

## Fix

Projects v2 takes PRs as items with the identical field set, so this is **membership, not a parallel mechanism**. A PR card carries the same `Priority` and `Awaiting` an issue card does.

| Decision | Set on the PR card | Effect |
|---|---|---|
| blocked / parked on a call | `Awaiting: discussion` (or `upstream`) | suppressed from actions |
| later, not never | `Priority: P4` | suppressed from actions |
| actively being driven | no `Awaiting`, `P1`–`P3` | reported every tick |

`backlog-digest.sh` emits the cards as `pr_board`; `backlog-rhythm.sh` joins by number and suppresses.

**`content.type` is what separates PR cards from issue ones, and it was confirmed against a real card rather than assumed** — an added PR reports `"type": "PullRequest"` with `number`, `title`, `url` and `repository` alongside it. Numbers are unique across issues and PRs in one repository, so this cannot collide with the existing issue lookup.

**Suppressed PRs are counted and listed, never dropped:**

```
  deferred: 4 (#490 priority P4; #437 priority P4; #354 awaiting discussion; #167 awaiting discussion)
```

Silently vanishing would trade one failure for another — the goal is to stop re-asking about a settled decision, not to lose the item. The reason travels with it so a reader can see *which* decision is doing the suppressing without opening the board.

## Contract change: `mark_ready`

An APPROVED, green, still-draft PR is now its own action, and **it is the one thing no board decision can defer.**

It used to hand back to `implement-issue` like any other unfinished draft, on the principle that this pass must not grow a second review loop. That principle still holds — but it is exactly what left **#629 sitting approved, green and draft**: the remedy on offer was a whole `implement-issue` session, and nobody spends one of those to run a single command. `gh pr ready` is a terminal action, not a loop, so naming it here duplicates nothing.

Two pre-existing tests encoded the old contract and are updated deliberately, with the reasoning in their docstrings rather than silently adjusted.

**`awaiting_maintainer` is deliberately NOT carved out the same way.** An approved PR waiting on a merge is not broken; it is the maintainer's call when to take it, and `P4` is exactly how they say "later". #490 sat approved for a day and was reported every tick as though that were news.

## Live effect

Applied the four real decisions to the board and re-ran:

| | before | after |
|---|---|---|
| actions due | 31 | **27** |
| recurring items for settled decisions | 4 | **1 line** |
| #631 (genuinely merge-ready) | reported | still reported |

## Test plan

- [x] **2065 passed**, 50 skipped (`pytest -m "not slow"`)
- [x] Ran `backlog-rhythm.sh` against the live board before and after setting the four cards — the counts above are measured, not projected.

Ten new tests, each pinning a case that would otherwise regress silently:

- `test_an_approved_draft_is_never_deferred` — the carve-out. Give an approved draft a `P4` card and it **still** reports `mark_ready`; remove the carve-out and it vanishes, which is the #629 failure.
- `test_an_approved_pr_can_be_deferred` — the opposite direction, so the carve-out cannot creep into covering `awaiting_maintainer`.
- `test_a_stale_approval_does_not_earn_mark_ready` — an approved-then-reworked draft keeps its APPROVED entry forever; the last non-COMMENTED verdict decides.
- `test_a_conflicting_approved_draft_is_not_mark_ready` — a conflicted PR gets no CI run, so flipping it ready would hand over something unmergeable.
- `test_a_pr_with_no_card_is_reported_as_before` — deferral is opt-in; absent a card nothing changes.

The behaviour change was also observed live: #629 and #631 were both found approved-and-draft during this work and flipped with `gh pr ready`, which is the state this action exists to catch.

## Outcome-level coverage

None applicable — no optimizer path is touched. This is the PO tooling; the pins above are the coverage.

## Scope

`scripts/backlog-digest.sh`, `scripts/backlog-rhythm.sh`, their two test files, and `.claude/skills/backlog/SKILL.md`. No product code. No CHANGELOG entry: agent tooling, no user-visible effect.

## Note for reviewers

Both scripts pass their jq program as a **single-quoted shell string**, so one apostrophe in a comment ends it early. In `backlog-rhythm.sh` that surfaces as `unexpected EOF while looking for matching )` pointing at the *top* of the block rather than at the comment. It cost a debugging round here; there is now a comment saying so in both files.


bess-agent:
@claude-bot review

Issue #441, opened 2026-07-31
Question: Is it always counting with 15 minutes?


ridax67:
This is a very minor issue, if it is any issue at all.

When it calculates the power to dump from the battery during a period, is it always counting with 15 minutes per period, even when Bess is started and the current period is only part of a period?

I was 5 minutes into a -100 battery dump period when haos wanted to reboot for an update.  When it came back up and Bess started, what should the result be?  I had expected that it just continued with the -100 dump, but instead it finished what was left of the period with a -66 dump.

Now I wonder how it counts with the current start period that is only a part of a period.  My guess is that it calculates the correct amount of energy remaining to dump, but it counts with all 15 minutes to do it, even only part of period is left.

If this is the case, I can live with it.  I will just try to restart Bess during less "intensive" periods.

bess-product-owner:
Picking this up — it went quiet on 3 August with a confirmed diagnosis and an untaken decision, which is the worst place for an issue to sit.

**Where it stands.** @ridax67 found that period duration is hardcoded at `0.25` hours throughout, so the DP and the inverter controller both assume a full 15-minute period even when BESS starts part-way through one. Five minutes into a −100 dump, a HAOS restart meant the remaining 10 minutes were planned as if 15 remained. @johanzander confirmed this in the code in this thread. It applies to first start, not only restart.

So this is **not** a question — it is a confirmed bug, and it is now labelled as one.

**What is actually blocking it** is a design decision nobody has made: which layer should know that a period is short.

1. **Scale the command** — the patch @ridax67 posted above, computing `15 - (now.minute % 15)` at the point of commanding. Small, local, one file.
2. **Make period duration a real variable** — thread `dt` through the DP and the controller. @johanzander described this as "changes here and there and in two files".

The tradeoff is not size, it is whether the *plan* is allowed to stay wrong. Option 1 leaves the optimizer still believing it has a full period while the command says otherwise; this codebase pins plan-versus-realized faithfulness deliberately, so a correction applied only at the command layer would show up as drift in exactly the property those tests exist to hold. Option 2 costs more and touches the optimizer core, but the plan and the hardware agree afterwards.

That call belongs to the maintainer, so it is flagged for a decision rather than picked here. @ridax67 — thank you for both the diagnosis and the patch; the delay is on our side, not a rejection of the approach.

PR #639, opened 2026-08-18
fix: scope the review-run check to its own PR, and survive a transient 503

johanzander:
## Problem

Two failures in `scripts/request-pr-review.sh`, both observed live tonight, both of which cost a paid review round.

### 1. Wrong run — it watched whatever ran most recently, not its own PR

`review_run_state()` took the newest `PR Review` run created after the trigger, **with no filter for which PR it belonged to**.

`pr-review.yml` triggers on `issue_comment` — which fires for comments on **issues**, not only PRs. Those runs are gated out and complete as `skipped` in about ten seconds. So *any* comment posted anywhere in the repo during a review produces a newer run that is `completed` and not `success` — which this function read as **`failed`**.

Measured on **#636**: a routine Product Owner comment on issue **#441** at `21:13:07` made the script abandon the review. That review went on to **APPROVE at 21:15:45**.

```
completed  skipped   Question: Is it always counting with 15 minutes?   ← #441, mistaken for ours
in_progress          feat: put PRs on the board ...                     ← #636, still thinking
```

The caller was told the run was broken while it was still working, and the natural response — re-request — would have spent a second review round on a review already in flight.

**Fix:** `displayTitle` carries the PR title for a run triggered on that PR, so it is the discriminator. The title is read once before the loop and matched through `--arg` rather than string interpolation — a title containing a quote would otherwise break the filter silently.

### 2. A single 503 killed the poll after the trigger had already posted

`set -e` turned one transient failure on the verdict read into a fatal `exit 1`. Because the `@claude-bot review` comment had *already* posted, re-running spent a second round on a live review. GitHub returned 503s for roughly ninety minutes on 2026-08-17 and this fired twice.

**Fix:** both review reads tolerate a failed call and retry on the next poll — which is exactly what `review_run_state` already did via its `unknown` state. A failed read is not a result.

## Evidence the tests discriminate

Reverted the script to `HEAD` and re-ran the three new tests:

```
FAILED  test_a_newer_run_for_a_different_pr_is_not_this_review
FAILED  test_a_transient_api_failure_does_not_kill_the_poll
PASSED  test_this_prs_own_failed_run_is_still_reported
2 failed, 1 passed
```

- The first two fail pre-fix and pass post-fix. The second reproduces the exact production shape — `returncode=1`, `stderr='HTTP 503: no server is currently available'`.
- **The third passes both before and after, deliberately.** It is the guard that the scoping does not swallow a *genuine* failure of this PR's own run — the failure mode the run-state check was added for in the first place (#623 burned 16 minutes waiting on a run that had already died).

Restored, all 12 tests green.

## Test plan

- [x] 12/12 in `backend/tests/test_request_pr_review.py`
- [x] Mutation verified above, not merely asserted

## Outcome-level coverage

Not applicable — no optimizer path touched. The pins above are the coverage, and the harness exercises the real jq filters rather than echoing fixtures back.

## Scope

`scripts/request-pr-review.sh` and its test file. No CHANGELOG entry: agent tooling, no user-visible effect.

### DIFF
--- a/backend/tests/test_request_pr_review.py
+++ b/backend/tests/test_request_pr_review.py
@@ -57,7 +57,16 @@ def env_file(tmp_path: Path) -> Path:
     return p
 
 
-def _gh(bin_dir: Path, reviews: list, run_state: str) -> None:
+PR_TITLE = "fix: the pr under review"
+
+
+def _gh(
+    bin_dir: Path,
+    reviews: list,
+    run_state: str,
+    extra_runs: list | None = None,
+    title: str = PR_TITLE,
+) -> None:
     """A `gh` answering the two queries the script makes.
 
     `reviews` is returned for `pr view --json reviews`; `run_state` drives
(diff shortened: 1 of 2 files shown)


bess-agent:
@claude-bot review

PR #641, opened 2026-08-18
docs: record the release-* ruleset, which now exists

johanzander:
Follow-up to #635. The `Protect stable hotfix branches` ruleset now exists:

```
id=20972632  refs/heads/release-*  non_fast_forward  enforcement=active  bypass_actors=[]
```

#635 shipped with a ⚠️ in CLAUDE.md saying `release-X.Y` was unguarded at both layers, plus a matching comment in `quality-check.sh`. That was accurate when written and is now false. A doc asserting a gap that has been closed misleads exactly as much as one asserting a guard that does not exist — and both failure modes have already happened once each in this area today.

## Changes

- Adds the `release-*` row to the ruleset table (four → five)
- Replaces the ⚠️ with the reason for the one asymmetry in that table

## The asymmetry, since it will look like an oversight later

`release-*` blocks `non_fast_forward` and deliberately **not** `deletion`, unlike every other branch ruleset here. The branch is pushed and tagged by the `release` skill (steps 2–6), so *rewriting* it must fail — but deleting it once the release is out is ordinary cleanup and must not. What makes dropping the deletion guard safe is the tag ruleset one row up: the published tag pins the released commit, so a spent `release-*` branch costs nothing to lose. Before tagging it is recoverable only from a local reflog, so delete it after, not during.

## One wording trap

`--delete release-X.Y` is still permitted, so the summary above the table cannot say the rulesets "refuse every line above" — it refuses every *rewrite*. I wrote the wrong version first and corrected it before pushing; flagging it because the same sentence is the natural thing to write next time the table changes.

## Verification

`scripts/quality-check.sh` — 0 errors, 0 warnings. Docs and comments only; no behaviour change.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -330,30 +330,28 @@ git push origin v9.9.0                # moves a published release tag
 
 The glob was never what made those safe — it was a blunt instrument
 compensating for having no guard at the only layer that can see a *ref update*
-rather than a command string. That layer now exists. Four **GitHub rulesets**,
-all `enforcement=active` with an **empty `bypass_actors` list**, refuse three
-of the four lines above — see the gap called out immediately after the table:
+rather than a command string. That layer now exists. Five **GitHub rulesets**,
+all `enforcement=active` with an **empty `bypass_actors` list**. They refuse
+every *rewrite* above; the one thing they still permit is deleting a spent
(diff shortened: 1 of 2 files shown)


bess-agent:
@claude-bot review

PR #642, opened 2026-08-18
fix: let a sandboxed session read the agent list, which silently truncates

johanzander:
## Problem

`claude agents --json` reads the session list from `~/.claude/jobs`, which is not in the sandbox allow list. **A sandboxed call does not fail — it silently truncates.**

Measured in one session, seconds apart:

```
sandboxed:    1 agent
unsandboxed: 14 agents, 7 of them live sessions sitting in worktrees
```

All seven read as dead.

## Why it matters

`backlog-rhythm.sh` keys `resume_implementation` off *"worktree on disk, no live session"*. With a truncated listing it told the maintainer to re-enter a worktree that a **live session was actively working** — putting a second session on one branch, against commits the advice itself calls *the only copy*.

That happened tonight, on **#619**. `implement-issue` Step 0 reads the same list before touching a resumed branch.

**An under-count rather than an error is the dangerous shape:** nothing about a short list looks wrong, so the wrong answer is acted on with full confidence. It also defeated two layers of documentation — both `backlog` and `implement-issue` already carry a "run this unsandboxed" warning, and it still bit, because the warning only helps whoever remembers to read it.

## Fix

One entry: `~/.claude/jobs` in `sandbox.filesystem.allowWrite`, with the rationale added to CLAUDE.md's per-knob list, where every other non-default entry is already explained.

This makes the warnings unnecessary rather than merely repeated.

## Test plan

- [x] The 1-vs-14 measurement above, taken in this session

## ⚠️ Not verified in this session, and it cannot be

The sandbox **captures its policy once at activation and ignores every later edit** (CLAUDE.md, "The sandbox captures its policy ONCE"). So the numbers above come from the *old* policy, and re-measuring here would prove nothing — it would return the same 1 no matter what this diff says, which is exactly the trap that section documents.

Verifying needs a genuinely fresh session, then:

```bash
bash scripts/verify-sandbox.sh          # via the Bash tool, not a ! prefix
claude agents --json | jq length        # sandboxed; expect >1
```

Flagging this rather than claiming a green I cannot produce.

## Scope

`.claude/settings.json` (one line) and `CLAUDE.md` (the rationale). No CHANGELOG entry: agent tooling, no user-visible effect.

## Context

This is the one piece of #620 that survives. #635 superseded the rest of that PR by guarding pushes server-side with GitHub rulesets, which is a better answer than the pattern enumeration #620 was shrinking — but `~/.claude/jobs` was unrelated to pushes and went with it.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -16,6 +16,7 @@
       "allowWrite": [
         ".",
         "~/GitHub/bess-manager",
+        "~/.claude/jobs",
         "~/Library/Caches/ms-playwright",
         "~/.cache/ms-playwright",
         "~/.npm"
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -617,6 +617,21 @@ redundant:
   **An under-count, not an error, is the dangerous shape**: nothing about a
   short list looks wrong, so the wrong answer is acted on with full confidence.
 
+  **`allowWrite`, not `allowRead`, and that is measured rather than assumed** —
+  the obvious objection is that `claude agents --json` only *reads* the list.
+  Two probes in a sandboxed session settle it:
+
+  ```
+  ls ~/.claude/jobs             ->  16 entries              # read:  ALLOWED
+  touch ~/.claude/jobs/.probe   ->  Operation not permitted # write: DENIED
+  ```
+
+  Reads were never blocked; the read policy denies only `~/.claude/ide`.
+  Enumeration needs to WRITE, and a dropped entry rather than an error is what
+  produces the truncation. Same lesson as the podman entry below, where
+  `filesystem.allowRead` was tried and disproved: read the actual error instead
+  of reasoning about what ought to be blocked.
+
 - **`filesystem.allowWrite`: the two user-level caches `worktree-setup.sh`
   writes** — `~/Library/Caches/ms-playwright` (plus its Linux spelling
   `~/.cache/ms-playwright`) and `~/.npm`. Setup runs `npm install` when a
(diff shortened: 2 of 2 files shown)


bess-agent:
@claude-bot review

bess-agent:
Good question, and the precedent you cite is the right one to hold this against — the podman entry a few lines below is exactly a case of `filesystem.allowRead` being tried and disproved by reasoning about what *ought* to be blocked. So I answered it the same way that entry was answered: with two probes rather than an argument.

In a sandboxed session, against `~/.claude/jobs`:

```
ls ~/.claude/jobs             ->  16 entries               # read:  ALLOWED
touch ~/.claude/jobs/.probe   ->  Operation not permitted  # write: DENIED
```

**Reads were never blocked.** The read policy in this sandbox denies exactly one path, `~/.claude/ide`, and `~/.claude/jobs` is not it — hence 16 entries listed. It is the *write* that is refused, and `~/.claude/jobs` appears in the write policy's deny-within-allow list.

So the enumeration can see the job files and still under-reports, because it needs to write (a lock or heartbeat beside each job) and drops the entry rather than raising when it cannot. That is also the mechanism behind the shape I called dangerous in the description: a write failure surfacing as a shorter list instead of an error.

`allowRead` would have changed nothing here, and adding it would have been the wrong-knob mistake in the other direction.

Pushed as `9923b4bf` — the measurement is now recorded in the CLAUDE.md entry itself, in the same shape as the other knobs in that section, so the next reader gets the answer without re-deriving it.

Your other two points stand and I have not tried to argue them away: `quality-check.sh` genuinely does not assert `sandbox.filesystem`, so its green says nothing about this fix; and the fix itself still cannot be verified in-session because the sandbox captures its policy once at activation. That verification is a fresh session plus `scripts/verify-sandbox.sh`, exactly as the description says.

bess-agent:
@claude-bot review

Issue #643, opened 2026-08-18
Runtime error: Check grid charge state


Doodlehusse:
[bess-debug-2026-08-18-155345.md](https://github.com/user-attachments/files/31182628/bess-debug-2026-08-18-155345.md)

## What happened
Operation: Check grid charge state
Category: sensor_read
Time: 2026-08-18 13:10:56

502 Server Error: Bad Gateway for url: http://supervisor/core/api/states/select.solaxa_inverter_allow_grid_charge

## Debug bundle
Please drag-and-drop **bess-debug-2026-08-18-155345.md** from your Downloads folder into the editor below before submitting.
The file contains your version, settings, recent logs, and the raw HA discovery dump that the maintainers need to reproduce the issue.

johanzander:
@claude-bot analyze

johanzander:
@claude-bot analyze

johanzander:
@claude-bot analyze

bess-product-owner:
**Fix status:** merged in #675, now on `main` and shipping in the current beta build. This issue closes automatically on the next stable release. Labelled `awaiting-release` to track.

bess-product-owner:
This is repo/automation tooling, not BESS Manager add-on behavior, so it never graduates through a product release — its fix merged to main and took effect immediately. The `awaiting-release` label doesn't apply here; removing it and closing as shipped.

PR #649, opened 2026-08-18
Fix #646: make Stage 2 publish by command, and fail the job when it doesn't

### LINKED ISSUE #646: Stage 2 analyze exits success, charges $0.33–1.58, and posts nothing — 2 of 6 runs silently produced no comment or label
## Problem

Stage 2 (`issue-analyze.yml`) **exits `success`, spends money, and delivers
nothing** — no comment on the issue, no label change. The workflow run is green,
`is_error: false`, `permission_denials_count: 0`, so every downstream signal says
the analysis happened. Nothing did.

It is not a hard failure and not intermittent noise: it reproduced on demand.

## Measured

Six real (non-`skipped`) Stage 2 runs exist. Four landed, two did not:

| Issue | Run | Outcome |
|---|---|---|
| #118, #252, #624, #627 | success | comment posted, `analyzed` applied |
| **#593** (2026-08-15 09:22) | success | nothing posted, label unchanged |
| **#643** (2026-08-18 18:57) | success | nothing posted, label unchanged |

Re-firing `@claude-bot analyze` on both reproduced it immediately:

| Run | Issue | Duration | Turns | Cost | Comment posted |
|---|---|---|---|---|---|
| 32173921625 | #643 | 179 s | 33 | **$1.58** | no |
| 32176358237 | #643 | 54 s | 9 | $0.33 | no |
| 32176355892 | #593 | 47 s | 7 | $0.54 | no |

**~$2.45 burned across three runs on two issues, with zero product.**

The turn counts are the tell. `--max-turns` is 100 and was never approached. The
workflow prompt requires reading three docs, delegating to the `bess-analyst`
sub-agent, independently re-reading every cited `file:line`, posting a structured
comment, and editing labels. Seven turns cannot be that. The agent is deciding
it is finished early and returning a final message instead of writing to the
issue.

## Why it went unnoticed for 3 days

Every layer reports healthy:

- The workflow run is **green** — nothing to investigate in the Actions tab.
- The `ready-for-analysis` label stays on, which is indistinguishable from
  "nobody has run analyze yet".
- The backlog pass sees an un-analysed issue and reports `triage_labels` or
  nothing — it has no notion of "analyze was requested and silently failed".

On #593 the maintainer eventually wrote the diagnosis by hand (a 2302-character
comment on 2026-08-16), which is the real cost: the stage was paid for and the
work was done twice.

## What is not the cause

Ruled out by direct inspection, so nobody re-derives them:

- **Not permissions.** `permission_denials_count: 0`, and the job grants
  `issues: write` with a GitHub App token that posts fine on the four runs that
  worked.
- **Not the trigger gate.** All six runs reached the Claude step; the `if` filter
  correctly skips PR comments and non-owner comments (a `skipped` run is the
  normal, healthy shape here).
- **Not a missing sub-agent.** `.claude/agents/bess-analyst.md` is in the
  checkout.
- **Not `--max-turns`.** 7–33 of 100 used.

## Two prompt defects visible on inspection

Neither is proven to be the cause, but both are real and cheap to fix:

1. **`PROCESS` has two steps numbered `3`** — "Delegate the investigation" and
   "After the sub-agent reports back, independently verify". So the posting step
   is labelled `4` while being the fifth instruction, and an agent tracking "am I
   done with step 3" can satisfy the wrong one.
2. **Posting is prose, not a command.** Step 4 describes a comment body; step 5
   is the only literal shell line (`gh issue edit …`). Every step the runs *did*
   execute is one with a literal command in it.

## Acceptance

A Stage 2 run must not be able to exit `success` without having written to the
issue. Whatever the fix, the check belongs *after* the agent, in the workflow —
a run that posted no comment and applied no label should fail the job loudly,
rather than passing and charging for it. `needs-human-review` exists precisely
for the honest-failure case and was never applied either.

## Repro

    gh issue comment 643 --body "@claude-bot analyze"   # as the repo owner
    gh run view <id> --log | grep -E '"(num_turns|total_cost_usd|is_error)"'
    gh issue view 643 --json labels,comments

johanzander:
## Summary

- Stage 2's prompt now **publishes by running `gh issue comment --body-file`** instead of describing a comment it never posts.
- A **post-agent step fails the job** when no comment was posted or neither `analyzed` nor `needs-human-review` was applied — the issue's acceptance criterion.
- `PROCESS` renumbered (it had two steps numbered `3`).
- `quality-check.sh` gains a static gate so neither half can be removed silently.

## Root cause

The prompt never told the agent to **run** anything to publish. Step 4 read:

> `4. Post ONE comment on the issue with this structure:` — followed by a markdown template.

That is a description of a document, not an instruction to execute a command. These workflows run in **agent mode** (`use_sticky_comment: false`, `track_progress: false`, confirmed in the run log), so the action posts nothing on the agent's behalf and creates no tracking comment. An agent that renders the template as its final assistant message has, by its own lights, posted the comment: the run exits `success` and nothing reaches GitHub. Having "finished" at step 4 it never reaches step 5, which is why the label was untouched too.

**The same-version contrast is what identifies it rather than merely fitting it:**

| Workflow | How its prompt says to publish | Result on action `459ad358` / CLI 2.1.234 |
|---|---|---|
| Stage 1 triage | `gh issue comment <n> -b "…"` — literal | posts |
| Stage 4 PR review | `gh pr review` — literal, named 3× | `APPROVED` landed 2026-08-18 06:38 |
| **Stage 2 analyze** | **prose template, no command anywhere** | **silent 3/3 the same day** |

Stage 2 was the only bot workflow whose publish step was prose, and the only one that failed to publish. The action is not broken — Stage 4 wrote to GitHub fine on the identical version hours earlier. Every step the silent runs *did* perform is one carrying a literal command (`gh issue view` ran; the analysis itself got done — #542 burned 17 turns and $0.72 of real investigation before publishing nothing).

**Limits of this diagnosis, stated plainly:** the action sets `show_full_output: false`, so every turn is hidden and only the result JSON survives. I could not observe the agent emitting the template as prose and stopping — this is inference from what the runs did and did not do, not from a transcript. That is precisely why the fix does not rest on the prompt change alone.

## Fix

**1. Publishing is a command.** Both the conclusive and the inconclusive path now write the body to a file and run `gh issue comment --body-file` (a file, not `-b`, because the body carries backticks, quotes and newlines that do not survive shell quoting). The prompt says explicitly that writing the diagnosis as a reply does not publish it.

**2. The workflow verifies it afterwards.** A step keyed `if: always()` re-reads the issue and fails the job if no comment appeared since the agent started, or if neither label was applied. A timestamp step before the agent lets it distinguish a comment the agent posted from the `@claude-bot analyze` trigger comment that is always already present.

The guard **asserts and deliberately does not repair.** Posting the comment on the agent's behalf would be a second publishing path whose only job is to route around the first one failing — it would mask the regression instead of surfacing it (`docs/agents/rules.md`, Debugging Protocol step 8). A loud red run is the product.

Worth being precise about what this buys, since the issue is framed around cost: it does not prevent the spend. A silent run still burns its $0.33–1.58 — the guard makes that visible instead of green. The saving is the *second* cost, the one on #593 where the maintainer wrote the diagnosis by hand three days later.

## Test plan

- [x] `./scripts/quality-check.sh` passes (Errors: 0, Warnings: 0), re-run after merging `origin/main`
- [x] `.venv/bin/pytest -m slow` — 554 passed, 8 skipped, 423s
- [x] YAML parses; job has 5 steps in the right order (`checkout`, `app-token`, `Mark analysis start`, `claude-code-action`, `Verify the analysis reached the issue`)
- [x] **The guard executed for real against the live GitHub API**, all three branches, using the exact script body from the workflow:
  - silent run simulated (`ISSUE=646 SINCE=2030-…`) → `posted: 0`, error annotation, **exit 1**
  - healthy run (`ISSUE=646 SINCE=2020-…`) → `posted: 1`, labels `bug,analyzed`, **exit 0**
  - **`ISSUE=643`, one of the real issues whose Stage 2 run went silent** → comments present but labels still `bug,ready-for-analysis` → label branch fires independently, **exit 1**
- [ ] **Not verified, and cannot be from here:** that the prompt change actually stops the silence. GitHub Actions cannot be run locally, so nothing on my machine tests it. The honest test is a live re-fire of `@claude-bot analyze` on #643 or #593 after this merges — both reproduce on demand, ~$0.50. If the prompt change is wrong, the guard turns that run red instead of green, so we find out immediately rather than in another three days.

- Reverted: `.github/workflows/issue-analyze.yml` to `origin/main` (`git checkout origin/main -- …`), leaving the gate in place
- Result: the gate failed on **both** halves independently —
  - `❌ issue-analyze.yml never names 'gh issue comment'. Its publish step is prose…`
  - `❌ issue-analyze.yml has no post-agent verification step…`
  - `Errors: 1`
- Restored: `git checkout HEAD -- …`; gate green again, `git status` clean

The two assertions were also checked as predicates before the fix — `names_command=False`, `has_always=False`, `has_readback=False` — and after: all three `True`.

Near-miss check on the label matcher: a label named `not-analyzed` does **not** satisfy the `analyzed` requirement (the comma-wrapping is what makes that hold).

## Outcome-level coverage

- `quality-check.sh` → "bot workflow publish contract": pins that both Stage 1 and Stage 2 name their publish command, and that Stage 2 carries an `if: always()` step re-reading `--json comments` / `--json labels`. Checked by shape rather than step name, which renames freely.
- The behavioural outcome — *a silent run goes red* — is covered by the live three-branch execution above rather than by a unit test, because the outcome only exists against the GitHub API. There is no execution model for a workflow run in this repo to assert against.

## Scope assessment

**Workaround check: the diff adds no parameter, flag, default-fallback, second construction site, extra trigger or branch whose job is to route around an ordering/timing/dependency problem.** The guard is an assertion that fails the job — not a retry, and explicitly not a step that publishes on the agent's behalf.

**Category: local.** Both changes stay inside `issue-analyze.yml`'s existing contract (its prompt text, its job steps), plus one assertion in a file that already does exactly this kind of assertion. The other three bot workflows were checked by grep and already name their command, so none of them needed touching.

## Notes for the reviewer

**No `CHANGELOG.md` entry, deliberately.** This is agent-infrastructure with zero user-visible effect — no BESS Manager user is affected — and the curated changelog omits internal changes of that kind. Say the word and I'll add one.

**Documentation check:** `docs/agents/bess-knowledge.md` and `docs/SOFTWARE_DESIGN.md` describe no mechanism this diff touches (`SOFTWARE_DESIGN.md`'s "Stage 2" hits are sensor-discovery stages, unrelated). `docs/agents/workflow.md` **is** updated — its pipeline diagram now records that Stage 2 fails the job when it publishes nothing, so a red Stage 2 run is interpretable.

**Flagged, not fixed:** all five workflows track the floating `anthropics/claude-code-action@v1` tag, so the code running in CI can change with no commit in this repo. It moved across this regression's boundary (`9d7150bc`/2.1.233 → `459ad358`/2.1.234).

To be clear about what that is and is not: it is **not** this bug's cause — Stage 4 published fine on `459ad358` hours after the switch, which is what rules it out. It is a reason future upstream changes arrive unannounced. Pinning each `uses:` to a full SHA would address that class, at the cost of a standing maintenance commitment (a pinned action needs deliberate bumping, so it wants a Dependabot rule alongside). Five files plus a policy is a separate call, not something to fold into a fix PR.

Closes #646

### DIFF
--- a/.github/workflows/issue-analyze.yml
+++ b/.github/workflows/issue-analyze.yml
@@ -35,6 +35,13 @@ jobs:
           app-id: ${{ secrets.CLAUDE_REVIEWER_APP_ID }}
           private-key: ${{ secrets.CLAUDE_REVIEWER_PRIVATE_KEY }}
 
+      # Marks the point the agent started, so the verification step below can
+      # tell a comment IT posted from the `@claude-bot analyze` trigger comment
+      # that started this run (which is always already present).
+      - name: Mark analysis start
+        id: started
+        run: echo "at=$(date -u +%Y-%m-%dT%H:%M:%SZ)" >> "$GITHUB_OUTPUT"
+
       - uses: anthropics/claude-code-action@v1
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
@@ -109,40 +116,65 @@ jobs:
                    the reporter's), file:line references, and whether the
                    debug bundle reveals a different root cause than claimed."
 
-            3. After the sub-agent reports back, **independently verify** by
+            4. After the sub-agent reports back, **independently verify** by
                reading the file:line locations it cited. Do not just trust
                the summary — quote the actual code.
 
-            4. Post ONE comment on the issue with this structure:
+            5. **Publish the analysis by RUNNING these two commands.** Writing
+               the diagnosis as your reply does NOT publish it — you are running
+               headless, nothing is watching your output, and a final message
+               reaches nobody. The issue only changes if `gh` changes it.
 
-                  ## Root cause
-                  <one-paragraph plain-English explanation>
+               Write the body to a file first, then post it. Use a file rather
+               than `-b` because the body contains backticks, quotes and
+               newlines that do not survive shell quoting:
 
-                  ## Evidence
-                  - `path/to/file.py:LINE` — <quoted code excerpt>
-                  - <additional file:line refs as needed>
+                 cat > /tmp/analysis.md <<'ANALYSIS_EOF'
+                 ## Root cause
+                 <one-paragraph plain-English explanation>
 
-                  ## Proposed fix
-                  <high-level approach in 2-4 sentences — NO code yet>
+                 ## Evidence
+                 - `path/to/file.py:LINE` — <quoted code excerpt>
+                 - <additional file:line refs as needed>
 
-                  ## Risks / open questions
-                  <anything you couldn't determine, edge cases, or test gaps>
+                 ## Proposed fix
+                 <high-level approach in 2-4 sentences — NO code yet>
 
-                  ---
-                  Reply `@claude-bot fix` to implement and open a draft PR.
+                 ## Risks / open questions
+                 <anything you couldn't determine, edge cases, or test gaps>
 
-            5. Update labels:
+                 ---
+                 Reply `@claude-bot fix` to implement and open a draft PR.
+                 ANALYSIS_EOF
+
+                 gh issue comment ${{ github.event.issue.number }} --body-file /tmp/analysis.md
+
+               Check that the command succeeded. If it failed, fix the problem
+               and run it again — do not proceed to step 6 with the comment
+               unposted, and do not end the run having only described it.
+
+            6. Update labels:
                  gh issue edit ${{ github.event.issue.number }} --add-label analyzed --remove-label ready-for-analysis
 
             ──────────────────────────────────────────────────────────────────
             IF YOU CAN'T REACH A CONCLUSION
             ──────────────────────────────────────────────────────────────────
-            Post what you found, what you read, and what specific information
-            is still missing. Then:
+            Publish what you found, what you read, and what specific
+            information is still missing — again by RUNNING the commands, not
+            by describing the outcome:
+
+                 cat > /tmp/analysis.md <<'ANALYSIS_EOF'
+                 ## Inconclusive
+                 <what you read, what you ruled out, what is still missing>
+                 ANALYSIS_EOF
+
+                 gh issue comment ${{ github.event.issue.number }} --body-file /tmp/analysis.md
                  gh issue edit ${{ github.event.issue.number }} --add-label needs-human-review
 
             Be honest — "I couldn't find the bug in the current code" is a
             valid outcome. Do not invent a root cause to fill the template.
+            An inconclusive comment that is actually posted is worth far more
+            than a confident one that is not.
 
             ──────────────────────────────────────────────────────────────────
             HARD CONSTRAINTS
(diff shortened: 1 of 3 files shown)


bess-agent:
@claude-bot review

PR #653, opened 2026-08-18
fix: stop the fleet prune from destroying the worktrees it cannot remove

johanzander:
The fleet prune reported "14 merged worktrees kept because each has uncommitted tracked changes — a real backlog of stranded edits." There were no stranded edits. Every one of those worktrees was wreckage the prune itself had created on earlier runs.

## What happens

`git worktree remove` is sandbox-denied. Unlike `git worktree add`, which fails harmlessly, it fails **destructively** — removal deletes the working tree *first* and only then unlinks `.git/worktrees/<name>`, which is the denied write:

```
$ git worktree remove --force .claude/worktrees/backlogger
error: failed to delete ".../worktrees/backlogger": Operation not permitted
error: failed to delete ".git/worktrees/backlogger": Operation not permitted
```

By then ~393 tracked files are gone, and there is no rollback. What remains is a **carcass**: a registered worktree whose `git status` is a few hundred ` D` lines and nothing else.

Both prune loops read that as "uncommitted tracked changes" and correctly refuse to auto-delete it — so **the failure makes the worktree permanently unprunable by itself**. Re-running hits the no-`--force` refusal; `--force` re-hits the denial. `git worktree prune` performs the same unlink, so it cannot clear the wreckage either. There is no in-sandbox path to removing a worktree at all.

## Evidence

13 carcasses accumulated over three sweeps (deletion timestamps cluster at Aug 15 14:30, Aug 16 23:15, Aug 18 22:13) across PRs #568, #596, #597, #600, #601, #603, #609, #612, #617, #629, #633, #634, #641 — all merged.

The missing path set is **byte-identical in every one of them**: `core/` (270), `frontend/` (111), `bess_manager/` (3), plus `pyproject.toml`, `Dockerfile`, `CHANGELOG.md`, `LICENSE`, `.pre-commit-config.yaml`, `docker-compose.ci.yml`, `mock-run.sh`, `package-addon.sh`, `.claude/settings.json`. Same filenames means the same APFS readdir order means the same abort point. Identical damage across many worktrees is the signature of this bug, not a coincidence.

Two branches had commits past their merged PR head; both check out as superseded — #600s `9923be4d` is verbatim in `origin/main:scripts/request-pr-review.sh:140-146`, and #568s extras are mains own squash commit pulled back via a merge. Nothing recoverable was lost.

## The fix

- **Neither skill removes a worktree any more.** `sweep-prs` Step 3 and `implement-issue` Step 4 report `PRUNE` and emit one `!`-prefixed command for the maintainer to run unsandboxed.
- **Both classify a carcass distinctly.** A dirty set that is entirely ` D` is reported as `CARCASS (failed prune)`, not as stranded edits. The predicate anchors on `^ D ` — unstaged deletions only — so a *staged* deletion still reads as intentional work. Verified against 5 cases (pure carcass, edit+deletion, staged deletion, added file, clean).
- **`docs/agents/local-agent-environment.md`** gains the `remove`/`prune` half of the `.git/worktrees` denial, next to the `add` half it already documented.

`implement-issue` Step 11 keeps using `ExitWorktree action=remove`, which is the harness doing the removal and therefore unsandboxed — that path was never affected.

`scripts/verify-sandbox.sh:88` already predicted this in a parenthetical ("a stray directory when `git worktree remove` then fails"). Nothing acted on it.

## Verification

`./scripts/quality-check.sh` — 0 errors, 0 warnings. Docs and skills only; no runtime code, so no CHANGELOG entry.

Cleanup of the 13 existing carcasses is a separate manual step — the command is in the skill and has to be pasted with `!`, since this session cannot run it.

https://claude.ai/code/session_014FQA38o22RkHZJ4zEHDZS3


bess-product-owner:
Confirming the premise with a measurement across the whole fleet — it is stronger than the PR states.

Every worktree currently failing the `status --porcelain -uno` gate is **100% deletions, zero modifications**:

```
backlogger:                          D=393  M=0  other=0
docs-release-ruleset:                D=397  M=0  other=0
drop-permission-hooks:               D=389  M=0  other=0
feat+issue-554-defer-tou-writes:     D=389  M=0  other=0
fix-gh-release-ask:                  D=393  M=0  other=0
fix-issue-569-huawei-solar:          D=389  M=0  other=0
fix-issue-604-signed-pair-aliases:   D=393  M=0  other=0
fix-issue-621-wizard-platform-gate:  D=397  M=0  other=0
fix-issue-624-pwl-window-bisect:     D=397  M=0  other=0
fix-issue-630-subfloor-solar-export: D=397  M=0  other=0
issue-589-divergence-gate:           D=389  M=0  other=0
review-loop:                         D=389  M=0  other=0
skill-ready-on-approval:             D=389  M=0  other=0
```

**13 of 13.** Not "mostly wreckage" — there is not a single modified file
anywhere in the set, which is what a genuine stranded edit would look like. Each
count clusters at 389/393/397, i.e. the tracked file count of `main` at the
moment that worktree was hit.

Two things worth folding in:

1. **It predates today.** I created four of these in a sweep this evening
   (`docs-release-ruleset`, `621`, `624`, `630`, all D=397). The other nine
   were already wreckage before I started, at lower counts — so the prune has
   been quietly doing this across multiple sessions, and each run's damage then
   trips the *next* run's uncommitted gate. That is the self-perpetuating part.

2. **A second, independent failure mode hits the same command**, and the fix
   should cover it: `git worktree remove` also aborts when the worktree holds
   **untracked** files — and it aborts *after* unlinking. The gate uses `-uno`,
   so untracked content passes the check and then blocks the removal it just
   authorised. Separately, `fix-issue-646-stage2-silent` classified as `PRUNE`
   while carrying `lock reason: claude session … pid 90253`; the skill's table
   says locked worktrees are skipped, but the script only matches on session
   `cwd`. I confirmed that pid was dead before unlocking it.

So there are three ways this command destroys a worktree it cannot remove:
sandbox-denied `.git/config`, untracked files, and a stale lock.

Issue #627, opened 2026-08-17
SPH read? (WIT inverter)


tissson:
[bess-debug-2026-08-17-191722.md](https://github.com/user-attachments/files/31146602/bess-debug-2026-08-17-191722.md)

## What happened
It seems Bess Manager have automatically chosen SPH even though I am using a Growatt WIT (low voltage) hybrid inverter. I have added all the necessary sensors manually and it seems to work except for the dashboard triggering errors (growatt server read ac charge times) and (critical system issues deteced, system health check)

Growatt WIT detected as growatt_server_sph

BESS auto-detects my Growatt WIT as MIX/SPA/SPH (DC-coupled).

All sensors are detected correctly:
- SOC
- Charge power
- Discharge power
- Battery capacity
- PV
- Grid

But BESS repeatedly calls:

growatt_server.read_ac_charge_times

which returns:

500 Server Error

Demo Mode works fine but no schedule is created.

Is WIT currently supported or should another platform be selected?

johanzander:
@claude-bot analyze

bess-product-owner:
Thanks for the bundle — it was detailed enough to separate four different things that were all firing at once here. The short version is that **the SPH/WIT detection you spotted is real, but it is not what is stopping your schedule from being created.**

**What is actually blocking the schedule** — these two are configuration, and you can act on them now:

1. **Price fetching is failing** (Nordpool `400`). Your exported config has `nordpool_hacs` as the active provider. Worth re-checking that setting, and whether the HACS Nordpool integration is healthy on your side. Without prices there is nothing to optimise, so no schedule is produced.
2. **The consumption sensor is not configured.** The optimizer needs a consumption forecast; with no sensor it cannot build one.

Fix those two and you should get a schedule, independently of everything below.

**The `growatt_server read_ac_charge_times` 500 error** is real but harmless here — it is only used by the periodic health probe and the startup hardware sync, both of which already tolerate failure. It is noisy, not blocking. It happens because detection picked the wrong device (your portal-level "Malax Total" entry rather than the inverter).

[...]

Issue #252, opened 2026-07-09
VPP Control don't affects the inverter as planned


zlakes01:
My settings is:

VPP Control

Power Control Mode: select.solax_remotecontrol_power_control
Active Power Target: number.solax_remotecontrol_active_power
Autorepeat Duration: number.solax_remotecontrol_autorepeat_duration
Power Control Trigger: button.solax_powercontrol_trigger
Battery Minimum SOC: number.solax_selfuse_discharge_min_soc
Charger Use Mode: select.solax_charger_use_mode

This select Power -400 W for export period and the battery continue charges.

johanzander:
Can you export debug logs and attach them here?

zlakes01:
[faaac926_bess_manager_2026-07-10T08-02-50.539Z.log](https://github.com/user-attachments/files/29880686/faaac926_bess_manager_2026-07-10T08-02-50.539Z.log)

zlakes01:
I'm also wondering where are all BESS entities?

bess-agent:
Thanks for the report and the log file — I looked into it, but I need a different export to confirm what's going on.

**What was attached is the BESS Manager add-on's raw application log**, not the structured debug bundle. It's genuine BESS output (I can see the optimizer's `core.bess.*` log lines), but it only shows what BESS *decided and sent* — it has no sensor readback afterward, so I can't confirm what the inverter actually did in response.

From the log, BESS did compute and send the command you described:
- Period 40 (10:00): `Intent=BATTERY_EXPORT, -400 W, DischargeRate=4%`
- It issued the correct SolaX VPP sequence (`select_option → "Enabled Battery Control"`, `set_value → -400`, autorepeat, `button.press`) as a real (non-test-mode) call, no errors.

That's as far as I can go without seeing `battery_power` / `battery_soc` / `pv_power` around 10:00–10:02, and without knowing your `Charger Use Mode` (`solax_charger_use_mode`) setting — this looks like it could be a firmware/config-level interaction with the SolaX inverter rather than a BESS bug, but I can't confirm that from a log alone.

**Could you generate and attach the debug bundle instead?** In the BESS Manager web UI:

1. Go to **Settings → Diagnostics**, click **"Export Debug Data"** (downloads a markdown file), or
2. Click the **"Report a problem"** icon in the header, which downloads the same bundle and opens a pre-filled GitHub issue for you.

[...]

PR #655, opened 2026-08-18
Stop Stage 2 reading the sub-agent's own system prompt, and give step 4 a check that fires

### LINKED ISSUE #654: Stage 2's main agent reads the sub-agent's own system prompt, and nothing uses it
## Problem

Stage 2's main agent reads `.claude/agents/bess-analyst.md` (25,681 B) as
REQUIRED READING item 3. That file **is** the `bess-analyst` sub-agent's own
system prompt, and no step in the workflow prompt tells the main agent to use
it for anything.

Follow-up to #650, which cut `CLAUDE.md` from the same floor. This is the
second-largest item left.

## Measured

**It is the sub-agent's system prompt.** Frontmatter: `name: bess-analyst`,
`tools: Read, Grep, Glob, Bash, WebFetch`. So the main agent's read is a second
copy in a different context window, not a shared one.

**No instruction consumes it.** The six PROCESS steps are: get issue context →
identify the current problem → delegate to `bess-analyst` → verify the cited
`file:line` → publish → label. Item 3 is described as a "domain expertise
checklist", but nothing directs the main agent to apply it. Its one distinctive
task — step 4, "reading the file:line locations it cited … quote the actual
code" — needs the code, not a triage checklist.

**The prompt already inlines the critical content.** The task the prompt passes
to the sub-agent is a near-verbatim restatement of the file's
`## CRITICAL: Separate Evidence from Claims` — the section most plausibly
serving as a judging standard:

| `bess-analyst.md` | already inline in the prompt |
|---|---|
| "The reporter's description is a **hypothesis**, not a diagnosis" | "The reporter's explanation is a HYPOTHESIS — do not assume it is correct" |
| "Never start from 'where in the code could this bug be?' … Start from what the debug bundle shows" | Step 2 "Triage the LATEST debug bundle FIRST"; Step 3 "Only after triage, read the design docs" |
| "verify that BESS Manager actually uses code path Y for this user's setup (inverter type, integration…)" | Step 2 "whether the reporter's HA integration matches the code path BESS Manager uses. Check inverter type (MIN vs SPH vs SolaX)" |
| "If the debug bundle shows fundamental issues … flag those FIRST" | Step 2 "If sensors are unavailable … that is likely the real problem" |
| "A design choice that is intentional and documented is not a bug" | Step 4(b) "the behavior isn't intentional per the design docs" |

That content is currently loaded **three times**: sub-agent system prompt,
main-agent file read, and paraphrased inline.

**The artifacts agree.** All four landed Stage 2 analyses (#118, #252, #624,
#627) use the *workflow's* four headings — `## Root cause` / `## Evidence` /
`## Proposed fix` / `## Risks / open questions`. None uses `bess-analyst.md`'s
seven-item Output Format or its six-item type-B shape. #118's Evidence section
is exactly what step 4 asks for: quoted code at cited lines.

## What is NOT duplicated

`## Root-Cause Decomposition` — the P-optimality / forecast-error /
control-noise split, with "do not average them into one blended verdict … that
has repeatedly produced wrong conclusions on this repo." That is a real
standard and it is not in the prompt. But it is a standard for the
*investigator*, which the sub-agent already has; the main agent holding an
unread copy changes nothing while no step asks it to audit against it.

## Proposal

1. Drop item 3 from Stage 2's REQUIRED READING.
2. Because the decomposition standard is worth holding at the main-agent level,
   turn it into an instruction that actually fires — extend step 4 to reject a
   sub-agent report that blends the three categories into one verdict, or that
   cites no `file:line`.

Net: dead weight becomes a check. Saving is **up to 25,681 B off the main agent
per turn** (floor 92,438 → 66,757 B, −28%). "Up to", because whether a given
run complied with the read instruction is not observable — #646 showed
non-compliance happens.

## Limits of this evidence

Inspection plus artifacts, not a controlled experiment. What is shown is that
nothing consumes the read and its critical content is already inline — **not**
that behaviour is provably identical without it. That is also the argument for
adding the step-4 check rather than only deleting: it leaves the main agent
with a stated standard instead of an unread file.

## Explicitly out of scope

- Editing `.claude/agents/bess-analyst.md` itself. The sub-agent needs it whole.
- The other floor items (`rules.md`, `architecture.md`, the inline prompt) —
  all three are consumed by named steps.

## Acceptance

- Stage 2 still delegates to `bess-analyst` and still publishes (#649's
  verification step must stay green).
- Step 4 names a concrete rejection criterion rather than "independently verify".
- `scripts/quality-check.sh` passes, including #649's publish-contract gate.

johanzander:
## Summary
- Stage 2's main agent no longer reads `.claude/agents/bess-analyst.md` (25,681 B) — that file is the sub-agent's own system prompt
- Step 4's "independently verify" gains concrete rejection criteria, so the removal trades dead weight for an instruction that fires
- `quality-check.sh` gains a context-contract gate (written RED first, both assertions mutation-tested)
- Net **≈ −24.9 KB** off the main agent per turn

## Root cause

`.claude/agents/bess-analyst.md` **is** the `bess-analyst` sub-agent's system prompt — frontmatter `name: bess-analyst`, `tools: Read, Grep, Glob, Bash, WebFetch`. Listing it as the main agent's REQUIRED READING loaded a second copy of another agent's instructions into a different context window, on every turn.

**Nothing consumed it.** The six PROCESS steps: get issue context → identify the current problem → delegate → verify the cited `file:line` → publish → label. No step applies a "domain expertise checklist"; step 4 needs the cited *code*, not a triage procedure the main agent never runs.

**It was loaded three times.** The section most plausibly serving as a judging standard, `## CRITICAL: Separate Evidence from Claims`, is already restated almost verbatim in the sub-agent task the same prompt passes:

| `bess-analyst.md` | already inline in the prompt |
|---|---|
| "The reporter's description is a **hypothesis**, not a diagnosis" | "The reporter's explanation is a HYPOTHESIS" |
| "Never start from 'where in the code could this bug be?'" | Step 2 "Triage the LATEST debug bundle FIRST" |
| "verify that BESS Manager actually uses code path Y for this user's setup" | Step 2 "whether the reporter's HA integration matches the code path BESS Manager uses" |
| "If the debug bundle shows fundamental issues … flag those FIRST" | Step 2 "If sensors are unavailable … that is likely the real problem" |
| "A design choice that is intentional and documented is not a bug" | Step 4(b) "the behavior isn't intentional per the design docs" |

**The artifacts agree.** All four landed Stage 2 analyses (#118, #252, #624, #627) use the *workflow's* four headings — never `bess-analyst.md`'s seven-item Output Format or its six-item type-B shape. The main agent follows the prompt, not the file.

## Fix

Remove item 3, with an explicit note saying why, and give step 4 the check that was missing. A report fails verification if it claims a **code** bug with no `file:line`, if a cited location doesn't say what's claimed, or if it blends P-optimality / forecast error / control noise into one verdict when actual behavior diverged from plan. A failed report routes to the existing inconclusive path rather than being dressed up as a root cause — no new mechanism.

## Review findings fixed

Three were defects in my own first draft:

| Finding | Why it mattered |
|---|---|
| "cites no `file:line`" would reject **correct** diagnoses | The same prompt tells the analyst that unavailable sensors / wrong inverter type / HA integration mismatch are likely the real cause — and those cite no code. Now qualified to code-bug claims only. |
| Inconclusive path never removed `ready-for-analysis` | `backlog-digest.sh:275` checks that label **before** `needs-human-review`, so the issue reads as un-analysed and gets re-dispatched — **re-billing the stage**. Widening that path without fixing it would have made a cost bug worse. Verified against the source; now clears the label. |
| New criterion used vocabulary defined only in the file it stops the agent reading | P-optimality had a gloss; `P≠R` never expanded `P` or `R`; control noise had none. All three glossed inline. |

Two hardened the gate: `WF.glob("*.yml")` missed `.yaml` workflows, and the delegation pin was a bare substring over the whole file — a benign reformat would raise a hard ERROR falsely claiming the delegation was gone, while a leftover mention in a comment would satisfy it after the real instruction was deleted.

## Test plan
- [x] `./scripts/quality-check.sh` — Errors: 0, Warnings: 0
- [x] `.venv/bin/pytest -m slow` — 554 passed, 8 skipped (419 s)
- [x] Workflow YAML re-parsed after the block-scalar edits; prompt extracted and asserted on 9 properties (delegation present, no numbered agent-file read, all three glosses, publish command intact, inconclusive path clears the label)

No mock-HA surface — this is a prompt change with no runtime behaviour. What was observed instead is the gate's RED→GREEN transition, the mutation results below, and the parsed prompt.

Both assertions were written before the fix and mutation-tested:

**Read-list assertion**
- Pre-fix file: **1 hit** — `3. .claude/agents/bess-analyst.md — domain expertise checklist` → gate RED
- Post-fix file: **0 hits** → GREEN

**Delegation assertion**
- `subagent_type: bess-analyst` → `general-purpose`: gate RED, `restored: True`
- Backtick reformat `` subagent_type: `bess-analyst` ``: **passes** (no false ERROR)
- Quoted reformat `subagent_type: "bess-analyst"`: **passes**
- String present only *outside* the `prompt:` block: **fails** (a comment can't satisfy it)

The last three are the ones the first bare-substring version got wrong.

## Outcome-level coverage

- `quality-check.sh`'s context-contract gate pins both halves: no sub-agent definition as a numbered required read in any workflow, and Stage 2 still delegating. No optimizer fixture or golden applies — the diff touches no runtime code.
- Deliberately narrow: the read-list matcher keys on the *numbered-list* shape. An unnumbered "read `<path>`" instruction would slip through. That's accepted — the prompt now carries an explicit "Do NOT read" note, and a matcher broad enough to catch every phrasing would flag that note too, i.e. fight its own documentation.

## Documentation check

Neither `docs/agents/bess-knowledge.md` nor `docs/SOFTWARE_DESIGN.md` documents the P-optimality / forecast-error / control-noise taxonomy — it lives only in `.claude/agents/bess-analyst.md`, so there was nothing to update. Worth flagging: this diff creates a second, abbreviated definition site in the workflow prompt, which can drift from the analyst file. Consolidating the taxonomy into `bess-knowledge.md` and referencing it from both would fix that, and is out of scope here.

## Size

Saving is **up to 25,681 B** off the main agent per turn. The prompt itself grew 7,098 → 7,821 B, so net is ≈ **−24.9 KB**. "Up to", because whether a given run obeyed the read instruction isn't observable — #646 showed it need not.

## Scope assessment

Adds no parameter, flag, default-fallback, second construction site or extra trigger. The failed-verification path reuses the existing inconclusive route rather than inventing a retry loop. Scope is **local** — Stage 2's prompt and one gate.

Closes #654

### DIFF
--- a/.github/workflows/issue-analyze.yml
+++ b/.github/workflows/issue-analyze.yml
@@ -56,7 +56,11 @@ jobs:
             ──────────────────────────────────────────────────────────────────
             1. docs/agents/rules.md           — hard constraints
             2. docs/agents/architecture.md    — codebase layout & data flow
-            3. .claude/agents/bess-analyst.md — domain expertise checklist
+
+            Do NOT read `.claude/agents/bess-analyst.md`. It is the
+            `bess-analyst` sub-agent's own system prompt — it already has it in
+            full, and you never run the procedure it describes. Reading it costs
+            25 KB of your context on every turn and tells you nothing you act on.
 
             ──────────────────────────────────────────────────────────────────
             PROCESS
@@ -120,6 +124,32 @@ jobs:
                reading the file:line locations it cited. Do not just trust
                the summary — quote the actual code.
 
+               The report does NOT pass verification if:
+                 - it claims a CODE bug but cites no `file:line` — there is
+                   nothing to check. A diagnosis that the real problem is an
+                   unavailable sensor, the wrong inverter type, an HA
+                   integration mismatch or a stale add-on version is a
+                   legitimate root cause and needs no code citation;
+                 - a cited location does not say what the report claims it says;
+                 - actual behavior diverged from the plan (unplanned import,
+                   floor breach, spike) and the report gives ONE blended verdict
+                   instead of a separate verdict for each of:
+                     - **P-optimality** — was the allocation optimal against the
+                       inputs the DP actually had at the decision point, judged
+                       WITHOUT hindsight?
+                     - **Forecast error (P≠R)** — the Plan was optimal for its
+                       own forecast, but that forecast differed from what was
+                       Realized.
+                     - **Control/execution noise** — the plan's commanded rate
+                       was never actually achieved by the inverter.
+                   Blending these has repeatedly produced wrong conclusions in
+                   this repo.
+
+               A report that fails verification is an inconclusive result, not a
+               diagnosis. Publish what you found and what is still missing via
+               the IF YOU CAN'T REACH A CONCLUSION path below — do not dress it
+               up as a root cause.
+
             5. **Publish the analysis by RUNNING these two commands.** Writing
                the diagnosis as your reply does NOT publish it — you are running
                headless, nothing is watching your output, and a final message
@@ -169,7 +199,7 @@ jobs:
                  ANALYSIS_EOF
 
                  gh issue comment ${{ github.event.issue.number }} --body-file /tmp/analysis.md
-                 gh issue edit ${{ github.event.issue.number }} --add-label needs-human-review
+                 gh issue edit ${{ github.event.issue.number }} --add-label needs-human-review --remove-label ready-for-analysis
 
             Be honest — "I couldn't find the bug in the current code" is a
             valid outcome. Do not invent a root cause to fill the template.
--- a/scripts/quality-check.sh
+++ b/scripts/quality-check.sh
@@ -516,6 +516,68 @@ then
 fi
 
 echo ""
+echo "📋 Checking bot workflow context contract..."
+echo "-------------------------------------------"
+
+# A workflow prompt must not tell its MAIN agent to read a `.claude/agents/*.md`
+# file. Those files are sub-agent system prompts: the sub-agent already has one
+# in full, so naming it as the main agent's REQUIRED READING loads a second copy
+# into a different context window, on every turn, for an agent that never runs
+# the procedure it describes.
+#
+# Stage 2 did exactly that with bess-analyst.md (25,681 B) and nothing consumed
+# it (#654). Its six steps are: get context, identify the current problem,
+# delegate, verify the cited file:line, publish, label -- no step applies a
+# "domain expertise checklist". Worse, the section of that file which could
+# plausibly serve as a judging standard (Separate Evidence from Claims) was
+# already restated inline in the sub-agent task the same prompt passes, so the
+# content was loaded three times over.
+#
+# The paired assertion matters as much as the removal: Stage 2 must still
+# DELEGATE. Dropping the read is the saving; dropping the delegation would gut
+# the stage, and the same "trim the prompt" instinct reaches for both.
+if ! python3 - <<'PY_CTX'
+import re, sys, pathlib
+
+WF = pathlib.Path(".github/workflows")
+bad = []
+
+for f in sorted(set(WF.glob("*.yml")) | set(WF.glob("*.yaml"))):
+    text = f.read_text()
+    for i, line in enumerate(text.splitlines(), 1):
+        # Match the shape the regression actually takes: a NUMBERED entry in a
+        # reading list, e.g. "3. .claude/agents/bess-analyst.md — checklist".
+        # Deliberately narrow. A prose line that merely names the path is not
+        # flagged, because the prompt now carries an explicit "Do NOT read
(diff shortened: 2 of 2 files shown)


bess-agent:
@claude-bot review

PR #656, opened 2026-08-18
feat: add the board bootstrap script, with the Priority tiers the board actually has

johanzander:
## Summary
- Split out of #614 per that PR's review: the board changes were undescribed and orthogonal to its mypy/review-bot fixes, and made up more than half the diff by file count.
- Adds `scripts/backlog-board-init.sh` and the two Product Owner board-memory files.
- Fixes the stale `P0` Priority tier the review flagged.

## What was left to move, and what was not

The original bundle was mostly overtaken by main while #614 sat. Checked against current `origin/main`:

| Board change in #614 | State on main |
|---|---|
| `Ready for Dev` / `In Progress` column names in `backlog-digest.sh` | already landed |
| same rename in `.claude/skills/backlog/SKILL.md` | already landed |
| `test_backlog_digest.py` column updates | already landed |
| `awaiting` checked before `analyzed` (the #96 dead-end) | already landed |
| `SKILL.md` documenting that no `P0` exists | already landed, correctly |
| `scripts/backlog-board-init.sh` | **not on main — moved here** |
| `.claude/agent-memory/product-owner/` (2 memories + index) | **not on main — moved here** |

So this PR carries only the two unlanded pieces. Nothing from #614's board half is dropped: its full state is preserved at `eb4d00e9`.

## The `P0` fix

`backlog-board-init.sh` created the Priority field with options `P0,P1,P2`. The live board has `P1`–`P4` and no `P0` — confirmed by this PR's own memory file (the GraphQL option ids), and by `SKILL.md` on main, which states "**there is no `P0`**". A board bootstrapped by the old script would have produced a tier no consumer reads, and would have been missing `P3`/`P4`, which the digest ranks on. Now `P1,P2,P3,P4`, with the reason recorded in the script so it cannot drift back.

The memory file's note that the skill text was "stale" is also corrected — that was true when it was written, and is not any more.

## Test plan
- [x] Verified against current main that each already-landed item really is landed (table above), rather than assuming the bundle still applied

`backend/tests/test_backlog_board_init.py` shims `gh` on PATH and asserts on the arguments the script passes — the pattern `test_backlog_digest.py` already uses for this shape of script. Review round 1 correctly pointed out that this PR's original "cannot be tested without a live board" claim did not survive contact with the repo's own precedent, so that claim is withdrawn and replaced by the tests.

Each test reddens under the mutation it guards:
- Reverted: `--single-select-options "P1,P2,P3,P4"` back to `"P0,P1,P2"`
- Result: `test_priority_field_is_created_with_the_tiers_the_board_has` FAILED, 1 failed / 2 passed
- Reverted: the lookup back to `... || true`
- Result: `test_a_failed_lookup_does_not_create_a_second_board` FAILED, 1 failed / 2 passed
- Restored: tree clean, 3 passed

## Also from review round 1
The duplicate-board risk is fixed, not merely noted. `gh project list` ended in `|| true`, which makes "no such board" and "the lookup failed" indistinguishable — a rate limit or transient GraphQL error would have produced a second "BESS Manager Backlog", contradicting the idempotence the script's own header promises. It now stops with a message, and the third test pins that.

The PO board-state memory was trimmed to its durable lesson: the per-issue triage snapshot it carried was mirrored board state, which `.claude/skills/backlog/SKILL.md` ("State lives on GitHub, nowhere else") forbids, and it was already stale two days on.

## Outcome-level coverage
The assertions are on what the script *does* — which `gh` calls it makes, with which arguments — not on its log text. For a bootstrap whose entire effect is those calls, that is the outcome. No optimizer fixture or golden applies; this diff touches no DP, intent or control path.

## Documentation check
`docs/agents/bess-knowledge.md` and `docs/SOFTWARE_DESIGN.md` describe optimizer behaviour and system design; this PR touches neither — no mechanism, formula or threshold either file documents is affected. `.claude/skills/backlog/SKILL.md` is the doc for this area and already describes the board schema correctly on main, including the absent `P0`.

No `Closes` — this is a split, not an issue fix.

### DIFF
--- a/.claude/agent-memory/product-owner/MEMORY.md
+++ b/.claude/agent-memory/product-owner/MEMORY.md
@@ -1,2 +1,2 @@
 - [Board field IDs](project_board_field_ids.md) — GraphQL node/field/option IDs for Status/Priority/Awaiting; Priority has NO P0 option, only P1-P4
-- [Board state 2026-08-16](project_backlog_board_state_2026_08_16.md) — board had only 1 of 36 open issues before this pass; digest column ≠ board presence
+- [Digest column ≠ board presence](project_backlog_board_state_2026_08_16.md) — the digest derives a column whether or not a card exists; diff `item-list` against `issue list` before trusting sync
--- a/.claude/agent-memory/product-owner/MEMORY.md
+++ b/.claude/agent-memory/product-owner/MEMORY.md
@@ -1,0 +1,2 @@
+- [Board field IDs](project_board_field_ids.md) — GraphQL node/field/option IDs for Status/Priority/Awaiting; Priority has NO P0 option, only P1-P4
+- [Board state 2026-08-16](project_backlog_board_state_2026_08_16.md) — board had only 1 of 36 open issues before this pass; digest column ≠ board presence
--- a/.claude/agent-memory/product-owner/project_backlog_board_state_2026_08_16.md
+++ b/.claude/agent-memory/product-owner/project_backlog_board_state_2026_08_16.md
@@ -1,6 +1,6 @@
 ---
 name: project_backlog_board_state_2026_08_16
-description: State of the backlog board before/after the first board-writes-only PO pass on 2026-08-16 — only #611 was on the board, 35 issues missing entirely
+description: The digest deriving a column for an issue does not mean the issue is on the board — learned when a 2026-08-16 pass found 1 card against 36 open issues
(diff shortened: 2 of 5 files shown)


bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

PR #657, opened 2026-08-18
fix: let implement-issue run end to end without permission prompts

johanzander:
## Summary

- `implement-issue` stalled mid-run waiting for approvals nobody was awake to give. The profile blocked the one command Step 11 prescribes and permitted several it never runs.
- Every command the skill runs from fetch to a ready PR now resolves to `allow`; the maintainer's calls still `ask`.
- Adds a test pinning the command inventory extracted from the skills, so a skill gaining a new command fails CI instead of stalling a run at 2am.

## Root cause

`Bash(gh api *)` sat in `ask`. Reading inline review comments (`pulls/<n>/comments`) has no `gh pr view` equivalent — that endpoint is the only way to get the feed, and it is a plain GET. So the review loop's own prescribed command (SKILL.md:698) was the one thing gated, while `gh auth token` and `gh issue close` were permitted outright.

Separately, the four scripts the skills actually drive — `request-pr-review.sh`, `gh-agent.sh`, `worktree-setup.sh`, `quality-check.sh` — matched no rule in any category. Permission rules match only the top-level command string, so each fell through to the auto-mode classifier and behaved unpredictably run to run.

## Fix

| Category | Change |
|---|---|
| `allow` | the four scripts; `gh pr checks/ready/comment`, `gh run list/view/watch`, `gh api` reads; `git *` (Step 4's prune uses `git -C <path> …`, which a narrow `git branch *` would miss); build/test commands |
| `ask` | `gh issue close`; `gh api` write-forms in **both** flag orderings — `gh api * -X *` alone cannot match `gh api -X POST <url>`, since the leading `*` cannot span the empty prefix |
| `deny` | `git reset*`, `git push --force*`/`-f*` (SKILL.md:176), `gh auth token*` |
| removed | blanket `Bash(gh api)` / `Bash(gh api *)` from `ask` |

`git push` is deliberately **not** gated: `main`, `beta` and tags are protected by rulesets with empty bypass lists, so a client-side prompt could only ever fire on the routine feature-branch case.

`gh pr merge`, `gh issue close` and the release verbs stay in `ask` for the opposite reason — the server cannot distinguish the agent calling `gh` with the maintainer's credentials from the maintainer clicking the button.

## Test plan

`backend/tests/test_agent_permissions.py` — 46 assertions over three tables (runs-unattended / needs-approval / forbidden), plus the flag-ordering case.

```
46 passed in 0.03s
2138 passed, 50 skipped, 562 deselected   # -m "not slow"
```

Run against the pre-fix settings it fails 29 of 46 — including every one of the four scripts, the Step 11 `gh api` read, and `gh auth token` resolving to `allow`.

## Scope assessment

Config plus one test file; no production code. No CHANGELOG entry — agent tooling has no user-visible effect.

## Known limitation

The test models the matcher as `deny > ask > allow` with `fnmatch`, so it asserts *the rule set expresses the intent*, not that the harness enforces it that way. The one behaviour verified against the real matcher by hand: mid-string wildcards do match (`git -c core.pager=cat stash --help` is caught by `Bash(git -* stash --*)`), which the `gh api` write-form rules depend on.

## Review round 1

`CHANGES_REQUESTED` correctly caught `Bash(git reset*)` in `deny` as a blocker rather than a follow-up. `deny` never prompts, and `rules.md:36-40` prescribes `git reset --soft HEAD~1` as the WIP-commit recovery pattern now that `git stash` is denied repo-wide — so the rule removed a documented hard-constraint workflow repo-wide, for humans too. Narrowed to `Bash(git reset --hard*)`, the only form that discards committed work.

Also added a `MUST_NOT_BE_DENIED` table: the `FORBIDDEN` table only exercised `--hard`, so the suite could not have caught the regression it introduced. A `deny` pattern broader than the danger it targets now fails CI.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -52,7 +52,22 @@
       "Bash(gh pr diff *)",
       "Bash(gh pr list *)",
       "Bash(gh pr review *)",
-      "Bash(gh pr create *)"
+      "Bash(gh pr create *)",
+      "Bash(git *)",
+      "Bash(gh pr checks *)",
+      "Bash(gh pr ready *)",
+      "Bash(gh pr comment *)",
+      "Bash(gh run list *)",
+      "Bash(gh run view *)",
+      "Bash(gh run watch *)",
+      "Bash(gh api *)",
+      "Bash(scripts/request-pr-review.sh *)",
+      "Bash(scripts/gh-agent.sh *)",
+      "Bash(./scripts/worktree-setup.sh*)",
+      "Bash(./scripts/quality-check.sh*)",
+      "Bash(.venv/bin/python scripts/*)",
+      "Bash(npm run build*)",
+      "Bash(npx *)"
     ],
     "deny": [
       "Bash(podman machine rm *)",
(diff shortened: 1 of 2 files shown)


bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

PR #658, opened 2026-08-19
feat: give issues and PRs an explicit state machine, so the backlog loop can progress work

johanzander:
Gives issues and PRs an explicit state machine, so every state has one cheap next action and the Product Owner loop can actually progress work.

## Why

The fleet had stalled completely: 8 open PRs, all drafts, 6 CONFLICTING, none ever approved. Four causes, each measurable:

1. **The PR lifecycle was collapsed into one column.** The digest derived an issue's column as `if $pr != null then "In Review"`, tested first, outranking everything. A never-reviewed draft, a conflicted draft and an approved-and-waiting PR all reported the same state.
2. **The only exit from draft cost a full session.** Every unfinished draft resolved to `/implement-issue` at $1-4. Nobody spends five of those in a tick, so the queue only grew.
3. **The action list sorted alphabetically.** `sort_by(.action, ...)` put `dispatchable` above `resume_implementation` because `d < m < r` -- every pass led with "start new work" and buried "finish started work".
4. **Five of eight PRs edited the same three files.** The skill already said to queue same-file work, but it was prose with no gate, so it never fired.

A fifth defect surfaced during design: `pr_for` returned `$matches[0]`, discarding every PR after the first -- while the project's own no-auto-close rule guarantees an issue has several.

## What changed

**State model.** `Status` is the phase, `Awaiting` is the wait, and they are now orthogonal. `In Verification` names the merged-to-main-but-unreleased period that previously had no representation. `Awaiting` becomes *signed*: `reporter`/`upstream`/`discussion` stay quiet because someone else owes us; `maintainer` is loud and ranks first, inverting a field that previously made an item blocked on you *quieter*.

**`advance-pr` (new skill).** The single copy of the review loop, extracted from `implement-issue` Step 11 -- not duplicated, moved. It advances exactly one PR by one state and exits, never re-reading the implementation context. Three callers: `implement-issue`, the rhythm pass, and you.

**Flow policy.** Actions rank right-to-left across the board (escalations first, then finish before start). WIP limit of 3 counting `In Progress` + `In Review` together. A collision gate that refuses to dispatch work touching a file already in flight, and clusters two Ready items that would fight over one file into a single unit.

**One issue, many PRs.** Every PR carries `Refs #N`; only the graduation PR carries `Closes #N`. The issue's column derives from *all* its open PRs.

## Test plan

- 138 targeted tests across `test_backlog_digest.py`, `test_backlog_rhythm.py`, `test_request_pr_review.py`
- 591 in the fast suite; `./scripts/quality-check.sh` green after merging `origin/main`
- Every guard added was verified by mutation -- inverted or removed, test confirmed failing, restored, confirmed passing. Two guards were found untested this way and strengthened.
- Not exercised against the live board: the two board-field additions are deliberately left to the maintainer (see below).

## Known gaps, deliberately not fixed here

- **The collision gate is inert until `predicted_files` has a producer.** The consumption side is specified and tested; nothing populates it yet, so `needs_touch_set` fires for every Ready item and `dispatchable` never appears from the script alone. It fails *shut* -- it blocks dispatch rather than waving collisions through. The tests fabricate that field, so their passing is not coverage of the missing half.
- **`pr_board` is still read by the scripts** although the design retires PR cards. Deferral now reads *both* card types, so this is safe today and safe to remove later.
- **The duplicate-review race is narrowed, not closed.** `request-pr-review.sh` now checks for an in-flight run before posting, but GitHub creates the run seconds after the comment lands, so two callers inside that window can still double-trigger. Closing it needs a check on recent review *comments*, not runs.

## Ordering after merge

1. Add the `In Verification` Status option and the `maintainer` Awaiting option to Project #1, then delete the now-stale note in `advance-pr/SKILL.md` that says `maintainer` does not exist yet.
2. Script-side `pr_board` removal.
3. Only then delete PR cards from the board -- doing it earlier empties `$pr_board` and un-parks every deferred PR.

Until step 1, the digest computes `In Verification` and `move_card` fires for it every tick with no valid target.

## Unrelated

`backend/tests/test_dashboard_api.py::TestHistoricalDataStatus::test_dismiss_persists_across_requests` is order-dependent and fails in some suite orderings. Proven unrelated to this branch: the file is byte-identical to the branch base and it fails with this branch's test files excluded entirely. It passes on this branch as it stands.

https://claude.ai/code/session_017uvZsYtTvvyhGFxjGc29kM


bess-agent:
@claude-bot review

PR #664, opened 2026-08-21
fix: make the mypy gate a ratchet, so it can actually be passed

johanzander:
## Scope assessment

**Local**, not structural. The gate already had one owner — the mypy step — invoked from two places with duplicated inline logic. This consolidates that logic into a single script both callers delegate to, rather than giving it a new owner.

The diff adds no parameter, flag, default-fallback, second construction site, or extra trigger whose job is to route around an ordering/timing/dependency problem. Two alternatives were considered and rejected with reasons recorded in the script header: a committed baseline file (drifts, needs regenerating) and `git worktree add` for baseline extraction (side effects on a repo shared with live agent sessions).

Every failure path fails closed — unresolvable merge-base, unextractable baseline, and a crashed baseline mypy run all exit 2 rather than reporting a pass.

## Summary

- `--follow-imports=silent` so mypy errors are attributed to the files under test, not to everything they import
- A merge-base baseline, so only errors this branch **introduces** fail the gate
- CI and `quality-check.sh` now share one script, so a green local gate and a green CI gate mean the same thing

## Root cause

#614 scoped mypy to changed files, intending that "the legacy backlog burns down as files get edited". As written the gate could not be satisfied at all: **mypy reports errors in the modules a checked file imports**, not only in the file named on the command line. Handing it one `core/bess` file pulls in `ha_api_controller.py` and 30 others — so the gate demanded exactly the repo-wide cleanup its own comment ruled out ("Repo-wide is not an option (2914 errors across 191 files)").

Measured on #643's six-line fix: **404 errors across 31 files, none introduced by that branch.** No PR touching `core/bess` could pass. It went unnoticed because #614 merged Aug 19, after the last `core/bess` PRs (#618, #619) had already landed — #643 was simply the first to hit it.

## Fix

`scripts/mypy-changed.sh` owns the logic; `ci.yml` and `quality-check.sh` both call it.

The baseline is **computed, not committed** — the merge-base tree is extracted with `git archive` and type-checked on the fly. No generated file to drift or regenerate, and deliberately not `git worktree add`: this repo is shared with other checkouts and live agent sessions, and a lint gate should not register and remove worktrees as a side effect.

Error signatures drop line numbers, so inserting a line above an existing error does not read as a new one. A file added on this branch has no baseline entry, so new files are still expected to be clean.

Two bugs found while building it, both now pinned by tests:

- The baseline pass runs from inside the extracted tree, where a relative `.venv/bin/mypy` no longer resolves. It silently produced an **empty** baseline, turning all 61 pre-existing errors in one file into "newly introduced". Fixed by resolving mypy to an absolute path, and by checking the baseline run's exit status rather than discarding it.
- `quality-check.sh` invoked the script by a cwd-relative path — correct from the repo root, but it meant the gate silently never ran from anywhere else. Now resolved against `quality-check.sh`'s own location.

## Test plan

- [x] `./scripts/quality-check.sh` — fast suite 2190 passed / 50 skipped, frontend 125 passed, mypy gate green on its own diff
- [x] Slow suite: 554 passed, 8 skipped
- [x] **Run against the real #643 diff**: `✅ mypy OK (no new errors; 85 pre-existing in touched files)`, exit 0 — where the old gate failed it on 404 errors
- [x] A genuinely new error still fails with exactly one finding
- [x] Inserting lines at the top of a legacy file (shifting every error's line number) does not manufacture new errors
- [x] A brand-new untracked file carrying an error fails
- [ ] `shellcheck` not run — not installed locally

- Reverted: the baseline arm, by pointing the `git archive` shim at a well-typed source while HEAD keeps the ill-typed one
- Result: `test_a_preexisting_error_does_not_fail_the_gate` and `test_newly_introduced_type_error_fails_the_gate` use the **identical HEAD source** and differ only in the baseline — so a pass cannot come from mypy never running
- Separately: the empty-baseline bug was caught by a real run reporting 61 new errors for a comment-only edit, before any test existed for it; `test_an_uncomputable_baseline_fails_the_gate` now pins the fail-closed behaviour
- Restored: tree clean

## Outcome-level coverage

`backend/tests/test_quality_check_mypy_gate.py`, rewritten rather than extended — its cases pinned the old "changed files must be clean" semantics. Every guard it had is kept (vacuity, unresolvable ref, spaced paths); the two new ones are the halves that matter now: a pre-existing error must pass, and an uncomputable baseline must fail closed. 6/6.

## Documentation

`docs/agents/rules.md` claimed all code must pass mypy "with zero errors/warnings". That was never true — there is a ~2900-error backlog — and it is now explicitly not the rule. It documents the ratchet instead.

No `CHANGELOG.md` entry: internal tooling with no user-visible effect, consistent with #656/#654/#651.

## Note

This PR proves itself — CI runs the new gate on the diff that changes it. It also unblocks `fix/issue-643-grid-charge-read`, pushed and waiting on this to land.

https://claude.ai/code/session_015EswRHZDV575RZYPBq7L7o

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -476,17 +476,11 @@ jobs:
         run: |
           set -euo pipefail
           git fetch --no-tags origin main
-          base=$(git merge-base FETCH_HEAD HEAD)
-          # --diff-filter=d drops deletions: a file removed on this branch is
-          # still named by the diff and mypy cannot check what is not there.
-          mapfile -t changed < <(git diff --name-only --diff-filter=d "$base" HEAD -- '*.py')
-          if [ ${#changed[@]} -eq 0 ]; then
-            echo "No changed Python files — nothing to type-check."
-            exit 0
-          fi
-          printf 'Checking %s file(s):\n' "${#changed[@]}"
-          printf '  %s\n' "${changed[@]}"
-          mypy --explicit-package-bases --ignore-missing-imports "${changed[@]}"
+          # Same script quality-check.sh runs, so a green local gate and a
+          # green CI gate mean the same thing. Without --include-worktree:
+          # CI has only the commit, and an untracked-file scan here would
+          # pick up build artefacts rather than authored code.
+          scripts/mypy-changed.sh mypy
 
   # ── Docker build & boot: verify production image starts ──────────
   docker-build-test:
--- a/backend/tests/test_quality_check_mypy_gate.py
+++ b/backend/tests/test_quality_check_mypy_gate.py
@@ -1,10 +1,16 @@
-"""Tests for the mypy gate in scripts/quality-check.sh.
+"""Tests for the mypy gate in scripts/quality-check.sh and scripts/mypy-changed.sh.
 
-The gate scopes mypy to files changed against `origin/main`. Two things have
-to hold for that to be worth anything: a real type error in a changed file
-must fail the run, and a run that could not resolve the ref must not be able
-to report success — warnings exit 0, so the severity of that branch IS the
-behaviour.
+The gate scopes mypy to files changed against `origin/main` and fails only on
+errors the branch INTRODUCED, measured against the same files at the
+merge-base. Four things have to hold for that to be worth anything:
+
+- a newly introduced type error must fail the run;
+- an error that already existed at the merge-base must NOT;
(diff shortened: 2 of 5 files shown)


bess-agent:
@claude-bot review

PR #690, opened 2026-08-24
fix: make the PR reviewer approve when nothing blocks

bess-agent:
## What

Cuts the PR reviewer's verdict menu from three options to two, so that "I found nits but nothing blocks" comes back as an **APPROVE with the nits in the body** instead of a `COMMENTED` review that leaves the merge blocked.

## Why

A `COMMENTED` review does not set GitHub's `reviewDecision`. It cannot clear an earlier `CHANGES_REQUESTED` — that block stands until a human dismisses the stale review by hand.

`pr-review.yml` nonetheless offered `COMMENT — questions/observations only` alongside `APPROVE — no blockers, only nits or none`. With findings in hand but none of them blocking, the bot reached for the option that felt honest rather than the one its own contract specified. Measured on #687:

| round | state | the bot's own words |
|---|---|---|
| 2 | `CHANGES_REQUESTED` | real findings, since fixed |
| 3 | `COMMENTED` | "Neither finding blocks merge on its own" |
| 4 | `COMMENTED` | "One finding, not blocking merge but worth a look" |

Two approvals submitted as comments. Four rounds in, every substantive finding fixed, and `reviewDecision` was still `CHANGES_REQUESTED` from round 2.

## What changed

- **`.github/workflows/pr-review.yml`** — two verdicts, chosen mechanically. `REQUEST_CHANGES` requires a nameable blocker: a `rules.md` violation, or a correctness bug stated as a concrete failure scenario (inputs/state → wrong output or crash) with `file:line` and the real code quoted. Everything else — nits, style, design questions, undemonstrable asymmetries, "worth a look before this lands" — is `APPROVE` with the finding written into the body; the author decides what to act on, and that call is theirs. `COMMENT` is banned outright, with the tell spelled out: if you are writing "not blocking" or "worth a look", you have already decided.
- **`scripts/request-pr-review.sh`** — **behaviour unchanged.** A `COMMENTED` is still held while the run is live and still returned once the run ends, so no caller hangs and the placeholder race stays closed. Only the documentation moves: a terminal `COMMENTED` is now a protocol violation to report, not a legal verdict.
- **`.claude/skills/advance-pr/SKILL.md`** — same reframing. Still treated exactly like `CHANGES_REQUESTED` (collect findings, do not flip ready), and now says to report that the bot broke its own contract, because the stale `reviewDecision` will still be blocking.
- **`backend/tests/test_request_pr_review.py`** — one docstring that asserted the old three-verdict contract. Assertions unchanged.

## Verification

- `pytest -m "not slow"`: 2281 passed, 50 skipped. (The 6 `test_quality_check_mypy_gate.py` failures in the sandboxed run are the known `mktemp`-under-`/var/folders` denial; all 6 pass unsandboxed.)
- `black --check` / `ruff check` clean on the one touched Python file; the workflow YAML parses.
- The script's own suite (`test_request_pr_review.py`, `test_agent_permissions.py`): 61 passed — the unchanged behaviour is what those tests pin.

## Note

This does not unblock #687 by itself. That PR's stale round-2 review still needs dismissing by hand; this is about the next PR, not that one.

https://claude.ai/code/session_017kSsdTKDSrvnDB6ZnMb8b3

### DIFF
--- a/.claude/skills/advance-pr/SKILL.md
+++ b/.claude/skills/advance-pr/SKILL.md
@@ -114,18 +114,24 @@ reopens the failure it fixed.
 - **CI must be green, not merely unconflicted, before `gh pr ready`.** A
   `MERGEABLE` PR with checks still running is not ready — `mergeable` only
   says the diff applies cleanly, it says nothing about whether it passes.
-- **`COMMENTED` is ambiguous, and `request-pr-review.sh` resolves it for
-  you — do not second-guess it.** The bot posts an early placeholder review
-  as `COMMENTED` before its real summary, so a `COMMENTED` seen while the
-  review run is still going decides nothing. The script asks whether the
+- **`COMMENTED` is not a verdict, and `request-pr-review.sh` resolves it for
+  you — do not second-guess it.** `pr-review.yml` has exactly two legal
+  verdicts, `APPROVE` and `REQUEST_CHANGES`; findings that do not block merge
+  are an approval with the nits in the body, because a `COMMENTED` review
+  leaves GitHub's `reviewDecision` showing the previous round's
+  `CHANGES_REQUESTED`. The bot also posts an early placeholder review as
+  `COMMENTED` before its real summary, so a `COMMENTED` seen while the review
+  run is still going decides nothing. The script asks whether the
   workflow run is still live, not how long has elapsed, because no fixed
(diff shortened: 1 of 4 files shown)


johanzander:
@claude-bot review

Issue #692, opened 2026-08-24
Terminal head_rate is priced over the whole horizon, not the pre-dawn window its knee integrates (#602)


bess-product-owner:
#602's concave terminal row prices its head segment with `median(buy)` taken over
the **whole optimization horizon**, while the segment that rate applies to is
defined by a different, much narrower window — the pre-dawn gap that
`knee_kwh_from_forecast` integrates over.

The two are not the same window, and they are not the same price.

## The inconsistency

`curve_from_knee` (`core/bess/terminal_value.py`) builds the head segment as "the
avoided purchase, `median(buy) * efficiency_discharge`: what the household does
not have to buy tomorrow because it carried the energy." The quantity it applies
to, `knee_kwh`, is derived by walking the next day's net load forward from the
terminal boundary and stopping at the first PV crossover — i.e. strictly the
pre-dawn hours.

So the *quantity* is an integral over the pre-dawn gap, and the *rate* is a
median over 112 periods spanning a full day including both evening peaks. The
head segment's own definition is an avoided-cost integral; its rate should be
measured over the hours that integral covers, and load-weighted, since that is
what "the purchase avoided" means.

## Why it biases upward

Pre-dawn hours are structurally the cheap part of the day. A median over the full
horizon is pulled up by evening peaks that the carried energy will never serve,
so `head_rate` will normally *overstate* the avoided cost — and `head_rate` is
exactly what decides whether to carry at all (`head_rate > max(sell)*eta_d`).

On `bess-debug-2026-08-24-200357` the two agree almost exactly, by luck:

```
median(buy) over 112 periods           2.8933
load-weighted pre-dawn buy (00:00-07:00) 2.8812      <- what the carry actually avoids
                                        0.4% apart
```

That closeness is a property of this day, not of the formula. The concern is the
systematic direction, not this bundle's error.

## Why it matters here

The margin this sits on is thin. On the same bundle:

```
load-weighted pre-dawn buy   2.8812
best terminal-day sell       2.6082
headroom to break-even        9.5%     (1.0% at the single cheapest pre-dawn hour)
```

A systematic upward bias in `head_rate` on top of a 9.5% margin is worth removing.
Measured blast radius: 19 of 27 knee-bound corpus fixtures have
`head_rate > max(sell_terminal_day)*eta_d` (18 with a non-zero knee), so this is
the common path, not an edge case.

## Proposed fix

Price the head segment over the same window `knee_kwh_from_forecast` walked,
[...]

PR #693, opened 2026-08-24
test: pin the #602 terminal-shape trade on real 2026-08-24 inputs

bess-agent:
## Summary
- Adds `regression_2026_08_24_200357` to the scenario corpus: the first fixture where #602's concave terminal row and the pre-#602 capped scalar disagree about a whole evening peak rather than about the last few kWh.

## Why
A v10.2 export and a `main` (f82ece60) export were taken 98 seconds apart on 2026-08-24 with **byte-identical optimizer inputs**, and their plans diverged by 26 SEK of in-horizon savings. Every existing #602 fixture sits where the knee only trims the tail, so the whole corpus stayed green through that change. This one does not.

## Configuration it captures
Knee-bound, and `head_rate` = median(buy) × η_d = **2.7486** sits above the best in-horizon export, max(sell) × η_d = **2.4778**. Because #602 drops the arbitrage-consistency cap in this regime, the concave row carries energy past the boundary where the capped scalar empties the pack to the floor.

Replayed against the pre-#602 terminal row (`TerminalValueCurve.flat(terminal_value_breakdown(...))`), every pinned field moves four to five orders of magnitude past its 0.001 tolerance:

```
field                            pinned (main)     pre-#602      delta
battery_solar_cost                    11.66296    -18.79713    -30.460
base_to_battery_solar_savings        161.33135    191.79143    +30.460
total_discharged                      33.27350     43.71907    -10.446
final SOE                                15.37         3.00
grid exported                             0.02        10.84 kWh
```

- Mutation run: fed the fixture the pre-#602 capped scalar (2.0778 — the cap does bind here: `buy_based` 2.3486 vs `sell_cap` 2.0778) instead of the concave curve.
- Result: the plan inverts — pack emptied to the floor, 10.84 kWh exported at the evening peak.
- Restored: tree clean, `test_scenarios.py` 52 passed.

## Outcome-level coverage
`expected_results` on the new fixture, via the auto-discovered `test_scenarios.py::test_all_scenarios` harness — plus a selector golden and a VPP baseline entry (`--add-new`, not a full re-baseline).

## Provenance — read before citing these numbers
**The figures above are not what the reporting system did.** The corpus derives a fixture's knee with `knee_kwh_from_trailing_darkness`, because a fixture has no tomorrow: **15.18 kWh** here. Production derived **10.67 kWh** forward from tomorrow's forecast, and ended at 13.65 kWh having exported 0.90.

So this fixture pins the **shape comparison**, not the production decision, and must not be cited as observed field behaviour. That caveat is written into the fixture's own `description` as well as the commit message, so it travels with the data.

## Test plan
- [x] `pytest core/bess/tests/unit/test_scenarios.py` — 52 passed
- [x] Full fast suite — 2281 passed (the 6 `test_quality_check_mypy_gate` failures are the known sandbox/`mktemp` issue; they pass unsandboxed)
- [x] Mutation verified as above
- [x] No existing golden or `expected_results` moved — the fixture is purely additive

## Not included
No `CHANGELOG.md` entry: this adds regression coverage only, with zero user-visible behaviour change.

No documentation change: the diff touches no mechanism, formula or threshold described in `docs/agents/bess-knowledge.md` or `docs/SOFTWARE_DESIGN.md` — it adds a fixture and its pins. Both files were checked for the symbols involved (`head_rate`, `knee_kwh`, `terminal_value`) and their existing descriptions remain accurate.

## Context
The investigation behind this fixture concluded **not** to revert #602 — the forgone export is 23.4 h out and gets re-decided on published prices before it executes, and the tariff relation `buy = 1.25·sell + 0.6245` (exact, derivable from the recorded price settings) makes self-consumption structurally beat export. The real weakness found is #692, the head-rate window defect. This fixture pins the behaviour that any fix for #692 would move.

Refs #602, #692


bess-agent:
@claude-bot review

Issue #698, opened 2026-08-25
Optimizer failure should not take down the control loop


bess-product-owner:
Split out of #697, whose priority 3 cannot be fixed inside the optimizer.

#697 fixed the allocation that OOM-killed the add-on eleven times on 2026-08-23.
That closes the specific pathology, but not the coupling that turned it into a
38-hour outage:

- The DP runs in-process with the API and the control loop (`backend/app.py:195`,
  `BackgroundScheduler` inside the uvicorn process). A memory pathology anywhere
  in the optimizer takes the whole add-on down, not just the solve.
- SIGKILL is uncatchable, so **no in-process guard can cover this class**. Every
  accuracy budget in `pwl_window_dp.py` raises `PWLWindowUnderRefinedError` and
  gets answered by #624's bisection — but only if the process is still alive to
  raise. Bounding allocation (as #697 does) is necessary and cannot be sufficient.
- Supervisor's restart backoff then converts a fast crash loop into an indefinite
  stop. That is correct behaviour on its part, but it means a deterministic
  optimizer bug becomes an outage with no further attempts and no alert.
- Throughout the outage the last-written inverter command (`battery_discharge_power_limit = 0`)
  stayed in force for ~48 hours. The battery could not serve load at any price —
  ~35 SEK missed in a single hour at 3.550 SEK/kWh on 08-24. See #694, #695.

## What this needs

1. **Run the optimization in a subprocess with its own `RLIMIT_AS`**, so
   exhaustion surfaces in the parent as a catchable failure instead of killing
   the API. This is the part that makes the failure mode expressible at all.
2. **Serve the last known-good schedule while a solve is failing**, so a
   repeatedly-failing optimizer degrades to a stale-but-sane plan rather than to
   whatever register value happened to be written last.
3. **Surface a persistently failing solve** rather than letting it be visible only
   as a silent crash loop in the add-on log.

## Why separate

Structural, and it touches process/lifecycle rather than the solver. Bundling it
into #697's memory fix would have hidden a small, provably output-identical
change inside a much larger one. `docs/agents/optimizer-architecture.md` records
the constraint under P6 and points here.

Issue #697, opened 2026-08-25
OOM crash loop in the #450 exact PWL re-solve took the add-on down for 38 hours


bess-product-owner:
The add-on was OOM-killed (SIGKILL) 11 times in 10 minutes on 2026-08-23, always at
the same point: the #450 exact PWL re-solve. Home Assistant Supervisor then stopped
restarting it, and the system stayed down for **38 hours** with the battery stranded.

This is deterministic, not a transient memory spike: same inputs, same crash, every
attempt.

## The crash

Every one of the 11 kills has an identical prologue:

```
2026-08-23 06:00:53 | INFO | battery_system_manager:2093 - Running optimization for 72 periods from 06:00
2026-08-23 06:00:53 | INFO | battery_system_manager:1974 - Terminal value: 0.481/kWh (buy_based=0.481,
                             sell_cap=1.515, cap_applied=True, median_buy=0.928, max_sell=2.016,
                             efficiency=0.95, cycle_cost=0.400)
2026-08-23 06:00:53 | INFO | dp_battery_algorithm:1993 - Optimization using dt=0.25 hours for horizon=72 periods
2026-08-23 06:00:53 | INFO | dp_battery_algorithm:2017 - Starting direct optimization: horizon=72,
                             initial_soe=9.0, initial_cost_basis=1.002
2026-08-23 06:00:57 | INFO | dp_battery_algorithm:2205 - Near-tied DP decisions detected (#450):
                             re-solving 2 window(s) [(23, 28), (38, 43)] with the exact PWL DP
/app/run.sh: line 26:    97 Killed    python -m uvicorn app:app --host 0.0.0.0 --port 8080
s6-rc: info: service legacy-services: stopping
```

The grid DP completes in ~4 s. The process dies during the PWL re-solve that follows,
before it logs anything further. `Killed` on the uvicorn line is the kernel OOM killer.

Note the windows are small — `(23, 28)` and `(38, 43)`, five periods each — so whatever
allocates is not proportional to window width in any obvious way.

## Crash loop, then permanent stop

```
06:00:53 kill      06:07:06 start -> kill
06:01:41 start ->  06:08:08 start -> kill
06:03:03 start ->  06:09:06 start -> kill
06:03:56 start ->  06:10:16 start -> kill
06:05:04 start ->  06:11:23 start -> kill (06:11:30, last)
06:06:04 start ->  ...then nothing until 2026-08-24 19:49 (manual start)
```

[...]

PR #699, opened 2026-08-25
fix: bound the exact PWL re-solve's memory so it cannot OOM the add-on (#697)

bess-agent:
## Summary

- Evaluate the exact PWL objective in fixed-size row blocks, so peak memory is independent of `|X|` and the re-solve can no longer OOM-kill the add-on.
- Add `PWL_MAX_EVAL_CELLS`, checked **before** the evaluation and denominated in cells; exceeding it raises `PWLWindowUnderRefinedError`, which #624's bisection already answers by re-sizing the window.
- Correct the `PWL_MAX_PREIMAGE_SEED_POINTS` comment that caused the mis-denomination, and the normative claim in `optimizer-architecture.md` that this bug proved wrong.

## Root cause

`_pwl_candidate_values_at` built a dense `|X| × |actions|` candidate matrix plus about a dozen same-shape temporaries in a single allocation. `|actions|` is ~102 for **every** battery, because the discharge lattice is `max_discharge_power_kw / 100` — so battery size is not the axis. Measured cost is a flat **~10 kB per breakpoint**, and the only free variable is `|X|`, which compounds per backward stage: `_pwl_window_seed_points` seeds the discharge preimage as `|xs_next| × ~100` (`pwl_window_dp.py:612`) and that set is routinely ~10× larger than the pruned row it collapses into.

Neither budget could catch it, for structural reasons rather than tuning:

- `PWL_MAX_PREIMAGE_SEED_POINTS` counts seed **abscissae**, not evaluation cells. Its own comment justified itself as "bounded by memory (1e6 float64 ~ 8 MB)". The abscissae are 8 MB; evaluating on them is ~10 GB.
- `PWL_MAX_BREAKPOINTS` is checked at `:767`, one line *after* the `values_at(X)` at `:766` it is meant to bound, and against the **pruned** row. The raw seed set at `:731` was evaluated with no size check at all.

So the solve never raised `PWLWindowUnderRefinedError`, never reached #624's bisection at `dp_battery_algorithm.py:2320`, and was killed by the kernel instead. **SIGKILL is not an exception**, which is why no in-process guard could have covered this class.

## Fix

Blocked evaluation. Every reduction in the objective is over `axis=1` (actions), including the import-cap floor, so rows are independent and the blocked result is **bitwise identical** to one whole allocation. The objective itself moved unchanged into `_pwl_candidate_values_block`; `_pwl_candidate_values_at` is now a thin wrapper that slices, checks the cell budget, and concatenates. No call-site change, no new control flow, no change to the value table, no P6 exception.

**Deliberately not** the "fall back to the grid DP result when exceeded" the issue originally proposed: `optimizer-architecture.md` P6 forbids catching `PWLWindowUnderRefinedError` to keep the grid DP's answer, and bisection already delivers the same availability without violating it.

**Split out to #698:** process isolation for the optimizer — the coupling that turned a solver bug into a 38-hour outage. Not fixable inside the optimizer, since SIGKILL is uncatchable and the DP shares the uvicorn process with the control loop.

## Scope assessment

**Local.** Confined to two functions in `core/bess/pwl_window_dp.py` plus two constants. The owner is the solver because it is the only thing that knows `|X|` before spending memory on it, and `_pwl_candidate_values_block` the only thing that knows `|actions|`. Not the call site — it already has the correct response to "too big to solve at once". Not `tie_detection.py` — that would cap windows, i.e. the wrong quantity again, which is the same mistake the existing budgets made.

**Workaround check:** the diff adds no parameter, flag, default-fallback, second construction site, extra trigger or branch whose job is to route around an ordering/timing/dependency problem. The ordering problem — a guard placed after the allocation it bounds — is fixed directly by putting the new check upstream of the evaluation, not papered over.

## Test plan

- [x] `./scripts/quality-check.sh` passes locally — `2288 passed, 50 skipped`, 0 errors, Black/Ruff/mypy/TS/ESLint green.
- [x] `.venv/bin/pytest -m slow` — `564 passed, 10 skipped` in 351s. **No golden, `expected_results` pin, VPP baseline or cost number moved**, which is the load-bearing check for the bitwise-identity claim.
- [x] **Observed in a memory-capped container** (`podman run --memory=250m` on the real backend image, running the real `optimize_battery_schedule`), which is the environment that actually kills the add-on:

  | | exit code | outcome |
  |---|---|---|
  | pre-fix | **137** (SIGKILL) | log ends at `Near-tied DP decisions detected (#450): re-solving 1 window(s) ... with the exact PWL DP`, then nothing — byte-for-byte the reporter's signature across all 11 kills |
  | post-fix | **0** | solves; 75 MB peak RSS; identical economics (grid-only 26.28, optimized -47.81, savings 74.09 SEK) and identical table (6 rows, `V[0]` 5986 breakpoints) |

  Uncapped in the same container: **375 MB → 75 MB** peak RSS. Traced peak on the same window: **198.1 MB → 12.1 MB**. Runtime 3.0s → 3.2s.

  Two honest caveats. (1) The compose stack was not used, because no `docker-compose.ci.yml` scenario produces a near-tie window — it would have exercised none of this. The capped container runs the same image and the same entry path instead. (2) The 1134 MB RSS in the issue is a macOS allocator-fragmentation figure; on Linux the same window is 375 MB, so the cap was tightened proportionally to sit between pre- and post-fix. The per-breakpoint cost is the portable quantity, and it is flat.

Three separate mutations, all on the real fixture:

- **The primary test was written RED, before the fix existed.** `test_a_five_period_window_solves_within_a_bounded_memory_footprint` failed at `198.1 MB` against its 48 MB ceiling — `AssertionError: exact PWL re-solve of a 5-period window peaked at 198.1 MB`. 1 failed. Green after the fix at 12.1 MB.
- **Reverted the blocking stride** (`range(0, X.size, block)` → `block + 1`, so blocks skip rows): the corrupted table propagated into the DP and `test_blocked_evaluation_is_bitwise_identical_to_one_whole_allocation` went red at fixture setup. 1 error.
- **Injected a row-coupling candidate** (`+ 1e-9 * X.size` on both return paths of the block — exactly the class of change the test exists to catch, and numerically harmless so nothing else breaks): `AssertionError: blocked evaluation differs from the single whole-array evaluation -- some candidate now couples rows, and blocking is no longer exact`. 1 failed.
- Restored after each; tree clean, `3 passed`.

One note on the identity test, because it was nearly vacuous: at `X.size = 512` both sides take the single-block path (`block` is 961 here) and it would have compared the code to itself. It now uses `4 × block + 7` points — past the boundary, with a ragged tail — and compares the wrapper against `_pwl_candidate_values_block` on the whole array, which *is* the pre-fix code path. It also asserts its own premise (`X.size > block`), so a future raise of `PWL_EVAL_BLOCK_CELLS` fails the test instead of silently hollowing it out.

Also worth recording: the multi-block branch is genuinely exercised, not just present. Instrumenting the wrapper on the fixture's own solve shows **146 of 190 calls** take it, max `|X|` **19,772**.

## Outcome-level coverage

- **Memory** (the subject of this fix): traced-peak assertion on a real DP-produced window, `test_a_five_period_window_solves_within_a_bounded_memory_footprint`. Deliberately not a breakpoint-count assertion — that proxy is exactly what was already in place and already failed, because it bounded the retained row while the transient evaluation was ten times larger.
- **Exactness**: `test_blocked_evaluation_is_bitwise_identical_to_one_whole_allocation` pins the blocked result against the pre-fix single-allocation path, and the full slow suite's scenario/golden/VPP pins are unmoved. No `R == P` scenario test was added because the change is provably output-identical rather than economic — the existing plan-faithfulness pins cover the behaviour and stayed green.
- **Guard ordering**: `test_an_oversized_evaluation_raises_before_it_allocates` asserts both that the ceiling fires and that peak allocation stays bounded while it does — the property the old ceiling lacked.

No new `*.json` fixture, so no golden or VPP re-baseline was needed.

## Documentation

`docs/agents/optimizer-architecture.md` is normative and was **wrong in its binding constraint**: it stated the exactly-solvable horizon is ~8 periods and "no budget raise extends it", when the real limit was transient evaluation memory, biting at five periods — inside the range the doc treated as safe. Corrected, stated as two general rules rather than this instance (bound the resource where it is spent, in its own units; a ceiling checked after the allocation is not a ceiling), plus the note that no in-process net can cover unbounded allocation because SIGKILL cannot raise.

`docs/SOFTWARE_DESIGN.md` has no PWL mentions. `docs/agents/bess-knowledge.md` mentions only the #450 preference table and an unrelated interpolant. Both checked by grep; neither needed a change.

Refs #697

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
+- **The add-on no longer runs out of memory and stops restarting when the optimizer refines a near-tied decision** — the exact re-solve now runs within a bounded memory footprint. ([#697](https://github.com/johanzander/bess-manager/issues/697))
 - **The battery now keeps a sensible overnight reserve instead of all-or-nothing** — end-of-horizon energy is valued by how much the house needs before sunrise, so midnight charge no longer flips between empty and full. ([#602](https://github.com/johanzander/bess-manager/issues/602))
 - **Sub-period discharge authorization now values stored energy accurately** — the gate read the battery's marginal value from a single grid cell, which mis-priced it in both directions, so it both held and released the battery in the wrong periods. ([#683](https://github.com/johanzander/bess-manager/issues/683))
 - **A stale `Awaiting` field no longer hides a qualifying bug from autonomous analysis** — the tier-1 carve-out is now computed from the issue's own evidence instead of the board field. ([#681](https://github.com/johanzander/bess-manager/issues/681))
--- a/core/bess/pwl_window_dp.py
+++ b/core/bess/pwl_window_dp.py
@@ -113,16 +113,100 @@ def _pwl_candidate_values_at(
     import_cap_kwh: float | None = None,
     capabilities: PlatformCapabilities = DEFAULT_CAPABILITIES,
 ) -> np.ndarray:
-    """Best achievable value at each SOE in `X` for period `t`: max over the
-    shared action set (IDLE, STORE, discharge grid) plus the
-    SOLAR_EXPORT-below-max bypass (#313), with V[t+1] evaluated exactly at
-    each candidate's true (continuous) next_soe -- no state snapping.
+    """Best achievable value at each SOE in `X` for period `t`, evaluated in
+    bounded row blocks (#697).
+
+    The objective itself lives in `_pwl_candidate_values_block`; this wrapper
+    exists only to stop it being asked for the whole of `X` at once. That
+    matters because the block builds a dense `|X| x |actions|` matrix and a
+    dozen same-shape temporaries, `|actions|` is ~102 for every battery
+    (`discharge_rate_step_kw` is `max_discharge_power_kw / 100`, so the count
+    is fixed by the lattice, not the hardware), and `|X|` compounds per
+    backward stage via the discharge-preimage cross product. Measured cost is
+    a flat ~10 kB per breakpoint, so the tens of thousands of breakpoints a
+    five-period window reaches turned into ~1.1 GB RSS and an OOM kill --
+    a kill, not an exception, so none of the accuracy budgets ever got to
+    speak and #624's bisection was structurally unreachable.
+
+    Blocking is exact, not an approximation: every reduction in the block is
+    over `axis=1` (actions), including the import-cap floor, so rows are
+    independent and `concatenate(block(X_i)) == block(X)` bitwise. It also
+    removes the allocator fragmentation that made peak RSS ~4x the traced
+    peak, since no single allocation is large enough to fragment around.
 
     `import_cap_kwh` is the house fuse's per-period grid-import ceiling
     (#429), enforced exactly as `_run_dynamic_programming` enforces it on the
     grid DP: total import (load + grid charging) is a hard constraint,
     constraining rather than excluding a period whose load alone exceeds the
     cap."""
+    X = np.asarray(X)
+    # `_pwl_candidate_values_block` appends one SOLAR_EXPORT-bypass column and
+    # at most one residual-cover column, so this is the width's upper bound.
+    n_actions = np.asarray(power_row).size + 2
+
+    cells = X.size * n_actions
+    if cells > PWL_MAX_EVAL_CELLS:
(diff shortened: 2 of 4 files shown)


bess-agent:
@claude-bot review

PR #705, opened 2026-08-28
feat: let a consumption forecast be overlaid with declared changes (#428)

johanzander:
## Summary
- Adds **Planned Consumption Changes**: a sparse, timestamped `{start, end, energy_kwh, mode}` overlay applied on top of whichever consumption forecast is configured (`ha_statistics`, `influxdb_7d_avg`, `sensor`, `fixed`) — not a fifth strategy. No overlay entity configured, no change to behaviour.
- `mode: "add"` layers declared energy onto the forecast; `mode: "set"` replaces the forecast across the span (the answer to an absolute/one-off block, and the way to declare a known load without double-counting its own history).
- New `core/bess/consumption_overlay.py`: parsing, composition, and a real-time (UTC-stepped) period grid so DST fall-back days compose correctly.
- Missing/unavailable/malformed overlay entity fails loudly via the runtime failure tracker rather than silently freezing the schedule.

## Design background
This is a deliberate redesign, not the original approach. PR #437 (a hand-authored 96-value series) was closed after a maintainer redesign comment; extensive follow-up discussion on #428 (Frank-Leysen, valexi7, nilaz) converged on the overlay-as-post-processing-stage shape and the `add`/`set` split used here. The updated design, including the residual-baseline (Managed Loads) and battery→EV policy questions intentionally deferred to follow-up issues, is posted on the issue: https://github.com/johanzander/bess-manager/issues/428#issuecomment-5445224631

## Scope assessment
- **Placement**: `_apply_consumption_overlay` is called from `_gather_optimization_data`, *after* the daily prediction cache and *after* the tomorrow-horizon extension. Applying it earlier (e.g. inside `_get_consumption_forecast`) would trap entries behind the date cache on cached strategies, and duplicate/drop them when the horizon extends into tomorrow. Both failure modes are pinned by tests (`test_consumption_overlay.py`).
- **Workaround check**: nothing here routes around an ordering problem with a flag or second construction site — the overlay is a genuine new stage with one call site, positioned where the data is actually final.
- **Category**: local — the change lives entirely inside `BatterySystemManager`'s existing consumption-forecast pipeline; no optimizer-internals (DP, intent classification, control/rate mapping) are touched, confirmed by grep and by review.
- **New classes**: `OverlayBlock` and `OverlayResult` (frozen/plain dataclasses in `consumption_overlay.py`) — the block shape (start, end, energy_kwh, mode) is the design the linked issue discussion converged on and is what's posted at https://github.com/johanzander/bess-manager/issues/428#issuecomment-5445224631; flagging explicitly per the new-class approval checklist item.

## Fix
- `core/bess/consumption_overlay.py` (new): `OverlayBlock`, `parse_overlay_blocks`, `period_starts_from` (UTC-stepped grid), `apply_overlay`.
- `core/bess/battery_system_manager.py`: `_apply_consumption_overlay`, called from `_gather_optimization_data`; runtime-failure reporting via `record_failure_once` (coalesces repeat failures instead of freezing the banner on the first one).
- `core/bess/ha_api_controller.py`: `get_consumption_overlay_blocks()` — returns `[]` when unconfigured, raises `ConsumptionOverlayError` for missing/malformed data. Reads the entity's `blocks` attribute regardless of its `state`, since a template sensor's `state` commonly comes from an unrelated helper that can itself read `unknown` while `blocks` still renders correctly — gating on `state` alone would make the documented example a schedule-stopper.
- `core/bess/sensor_collector.py`, `core/bess/health_check.py`: overlay participates in the prediction health check; an empty declared-blocks list reports OK ("None declared"), not WARNING.
- `frontend/src/lib/sensorDefinitions.ts`, `SensorConfigSection.tsx`: new shared integration, registered in `SHARED_INTEGRATION_IDS` (an earlier omission here was the review's HIGH finding — an unregistered integration silently drops whatever the user configures).
- User-facing naming: **"Planned Consumption Changes"** throughout (backend messages, HA sensor display name, frontend labels, docs) — the name settled on during the design discussion; internal identifiers (`consumption_overlay` sensor key, module/file names) are unchanged.
- Docs: `docs/USER_GUIDE.md` (new section, incl. the midnight-rollover trap for overnight declarations), `docs/agents/bess-knowledge.md` (mechanism + the pre-existing, untouched DST flaw in `period_index_to_timestamp`), `CHANGELOG.md`.

## Test plan
- [x] `./scripts/quality-check.sh` passes locally — Black, Ruff, mypy (changed files), frontend tests, TypeScript, ESLint, permission surface, bot workflow contracts all green. **3 known pre-existing failures** in `backend/tests/test_quality_check_mypy_gate.py` are unrelated to this branch (confirmed: neither that test file nor `quality-check.sh` is touched here) — the test's synthetic stub project lacks `.github/workflows/` and `.claude/settings.json`, which a workflow-contract check added independently of this PR now expects to exist.
- [x] Fast suite: 2305 passed, 50 skipped (+ the 3 known-unrelated failures above). Slow suite: 564 passed, 10 skipped.
- [x] Local mock-HA E2E (`docker-compose.ci.yml`, `ci-normal-day` scenario, `libfaketime` pinned to the scenario's date): configured a `sensor.bess_consumption_overlay` template sensor with a 10 kWh EV block for 20:00–22:00, wired it into settings, and confirmed via container logs that a real `update_battery_schedule` run applied it: `Applied planned consumption changes: 1 entr(y/ies), 10.0 kWh net over the horizon`. The served schedule (`/api/growatt/detailed_schedule`) showed real DP-derived strategic intents afterward (not the empty-schedule fallback). System health reported `Planned Consumption Changes: OK` throughout. Verification-only scenario/settings edits were reverted before commit — not part of this diff.

The DST fix (`_overlap_seconds` measured each period's end via naive `period_start + PERIOD_DURATION`, disagreeing with the UTC-stepped grid across the 2026-10-25 fall-back hour):
- Reverted: the UTC-based `period_end` computation in `_overlap_seconds`, back to `period_start + PERIOD_DURATION`.
- Result: `test_apply_overlay_delivers_full_energy_across_the_dst_fall_back_hour` FAILED — an 8.0 kWh declared block delivered 10.0 kWh (`assert 10.0 == 8.0 ± 8e-06`), matching the exact bug measured during review (overlap inflated on the repeated hour, then zeroed for the following periods).
- Restored: fix reapplied, 17/17 tests in `test_consumption_overlay.py` pass.

State-gating fix (`get_consumption_overlay_blocks` previously required `state` to be a real value before reading `blocks`, which the documented template example — `state` sourced from an `input_boolean` — could itself defeat):
- Reverted: N/A (new test written directly against the pre-fix code path, confirmed RED first).
- Result: `TestConsumptionOverlayBlocksStateGating::test_blocks_are_read_even_when_state_is_unknown` / `..._is_unavailable` FAILED against the original state-first check (raised `ConsumptionOverlayError: ... is unavailable`/`is unknown` even though `blocks` was present and valid).
- Restored: fix applied (check `blocks` first, fall back to reporting `state` only when there's no data), all 4 new tests + full `test_ha_api_controller.py` (149 tests) pass.

## Outcome-level coverage
- `test_consumption_overlay.py` (17 tests): composition arithmetic (`add`/`set`, DST contiguity, clamping), parsing/validation, and four tests exercising `_gather_optimization_data` directly to confirm the overlay's output actually reaches `full_consumption` — not a synthetic-input unit test on the changed function in isolation.
- `test_ha_api_controller.py::TestConsumptionOverlayBlocksStateGating` (4 tests): the state-vs-blocks precedence fix.
- `test_prediction_health_check.py::TestConsumptionOverlay` (4 tests): health-check participation, including the empty-blocks-is-OK regression.
- No `R == P` plan-faithfulness scenario test: this change alters an *input vector* to the DP (`consumption_predictions`, before `_run_optimization`) and touches no DP internals, intent classification, or control/rate mapping. Given an identical input vector, everything downstream behaves exactly as before, so a full scenario pin would re-test unchanged code with no discriminating power here — confirmed independently by the background code-review agent's read of the diff.

Refs #428

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Added
+
+- **Tell BESS what consumption is coming with Planned Consumption Changes** — declare an EV session or a skipped pool pump in a template sensor, and it applies on top of whichever consumption forecast you already use. ([#428](https://github.com/johanzander/bess-manager/issues/428))
+
 ### Fixed
 
 - **The add-on no longer runs out of memory and stops restarting when the optimizer refines a near-tied decision** — the exact re-solve now runs within a bounded memory footprint. ([#697](https://github.com/johanzander/bess-manager/issues/697))
--- a/core/bess/battery_system_manager.py
+++ b/core/bess/battery_system_manager.py
@@ -13,6 +13,11 @@ from typing import Any, ClassVar
 import requests
 
 from . import time_utils
+from .consumption_overlay import (
+    ConsumptionOverlayError,
+    apply_overlay,
+    period_starts_from,
+)
 from .daily_view_builder import DailyView, DailyViewBuilder
 from .daily_view_store import DailyViewStore
 from .dp_battery_algorithm import (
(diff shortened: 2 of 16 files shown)


bess-agent:
@claude-bot review

bess-agent:
@claude-bot review

Issue #704, opened 2026-08-27
Tomorrow's price won't update


TomasSandstrom:
I have noticed (see screenshot) that for me the system won't update with the new prices for the next day. I am using Nord Pool (HACS). There seems however to be a price increase (I'm guessing my additional costs for import)

<img width="1627" height="573" alt="Image" src="https://github.com/user-attachments/assets/b2f6df58-54c6-4c15-a277-d6087921c56c" />

[bess-debug-2026-08-27-203305.md](https://github.com/user-attachments/files/31527223/bess-debug-2026-08-27-203305.md)

bess-product-owner:
Triage marked this bug ready for analysis (debug bundle received). Firing deep analysis per the backlog autonomous-spend rule.

@claude-bot analyze

bess-product-owner:
The Stage 2 run triggered on 2026-08-28 failed at its "verify the analysis reached the issue" step — no diagnosis was posted. Re-firing.

@claude-bot analyze

bess-product-owner:
Fix merged in #714 and now on `main`, so it's in the current **beta** build. On the HACS Nordpool source, BESS now only uses the `tomorrow` array when the sensor's `tomorrow_valid` attribute is true, so a pre-publish mirror of today's prices (inflated by the VAT multiplier) can no longer get latched for the day. If you're on the beta add-on, give it a run past 13:00 CEST. This issue closes automatically when the fix graduates to a stable release.

bess-product-owner:
Shipped in stable release [v11.0.0](https://github.com/johanzander/bess-manager/releases/tag/v11.0.0). Closing (release PR carried no closing keyword).

PR #714, opened 2026-08-29
fix: reject pre-publish tomorrow mirror on HACS Nordpool source (#704)

johanzander:
## Summary
- On the HACS Nordpool integration (`nordpool_hacs`), tomorrow's prices could get stuck all day at today's prices inflated by the VAT multiplier.
- `HomeAssistantSource._extract_prices_for_date` now uses the plain `tomorrow` array only when the sensor's `tomorrow_valid` attribute is true, and strips VAT on both plain-array fallback paths.

## Root cause
Quoting the Stage 2 diagnosis: BESS's cached "tomorrow" price data got permanently stuck at a wrong value fetched *before* Nordpool publishes next-day prices (~13:00 CEST). The stuck value was today's own already-VAT-adjusted price shape multiplied by the VAT multiplier again.

`_extract_prices_for_date` tried the timestamp-validated `raw_today`/`raw_tomorrow` path first, then fell back to the plain named arrays:

```python
elif target_date == tomorrow_date:
    prices = sensor_attributes.get("tomorrow")
    if prices:
        return self._handle_dst_transitions(prices)
```

Unlike `_parse_raw_data_for_date` (which does `price / self.vat_multiplier` and a per-array date check), this fallback never divided by `vat_multiplier` and never verified the array actually held tomorrow's data — it trusted the attribute name. Before publish, the HACS Nordpool sensor keeps `tomorrow_valid` false while its `tomorrow` array still mirrors today's VAT-inclusive values, so this fallback accepted the mirror. `PriceManager.get_price_data` then cached it for `_tomorrow_date` with no intraday invalidation, so one pre-13:00 cycle poisoned every later cycle that day.

The reporter's debug bundle confirms it: cached `tomorrow[0]=1.522`, cached `today[0]=1.2176`, ratio exactly `1.25` for all 96 periods; the live sensor had `tomorrow_valid: true` with real, much cheaper next-day prices by export time.

The other three price providers (`OfficialNordpoolSource`, `OctopusEnergySource`, `EntsoeSource`) all do per-entry timestamp date-filtering and raise `PriceDataUnavailableError` when tomorrow isn't verifiably available — the HACS named-array fallback was the lone outlier.

## Fix
- Plain `tomorrow` array is used only when `sensor_attributes.get("tomorrow_valid") is True` — the HACS component's own "next-day prices are real" signal, the equivalent of the validity checks the other providers already have. Strict `is True` (not truthiness) so a stringified `"false"` can't re-admit the mirror.
- Both plain-array fallback paths (`today` and `tomorrow`) now divide by `vat_multiplier`, matching `_parse_raw_data_for_date` and every other source.
- When tomorrow isn't confirmed, the method returns `None` → caller raises `PriceDataUnavailableError` → `get_tomorrow_prices()` returns `[]`, so nothing wrong is cached and a later cycle retries once real prices publish.
- The timestamp-validated `raw_today`/`raw_tomorrow` path is unchanged and still preferred.

Scope: local — entirely inside `HomeAssistantSource`, within the method's existing `-> list | None` contract. No new parameter/flag/trigger. The diagnosis's optional "add a tomorrow cache-invalidation trigger" is deliberately not included: once the source refuses unconfirmed data, nothing bad gets cached, so no new invalidation mechanism is needed.

## Test plan
- [x] `./scripts/quality-check.sh` passes locally (fast suite, black, ruff, mypy — clean)
- [x] `.venv/bin/pytest -m slow` — 564 passed, 10 skipped
- [x] Stood up the real mock-HA + backend stack (`docker-compose.ci.yml`, `nordpool_hacs` provider) and drove `HomeAssistantSource` in-container against the live mock sensor:
  - pre-market (`tomorrow_valid: false`, `tomorrow` mirrors `today`=0.5): `get_prices_for_date(tomorrow)` → `PriceDataUnavailableError`; `PriceManager.get_tomorrow_prices()` → `[]` (nothing cached, retries next cycle)
  - post-publish (`tomorrow_valid: true`, real prices=1.0): `get_prices_for_date(tomorrow)` → `[0.8, …]` × 96 (VAT stripped: 1.0/1.25), cached with `price: 0.8`
  - today, always: `[0.4, …]` (VAT stripped: 0.5/1.25)

- Reverted `core/bess/price_manager.py` to `HEAD`, ran the 4 new tests:
- Result: `test_home_assistant_source_rejects_premarket_tomorrow_mirror` FAILED (returned 96 periods at today×1.25 instead of raising), `test_home_assistant_source_tomorrow_plain_array_used_when_valid` FAILED (returned 1.5, not 1.5/1.25), `test_home_assistant_source_today_plain_array_fallback_strips_vat` FAILED (VAT not stripped); `test_home_assistant_source_tomorrow_from_valid_raw_data_still_works` passed (guard on the already-correct raw path). 3 failed, 1 passed.
- Restored: tree clean.

## Outcome-level coverage
- Unit tests on `HomeAssistantSource` via `get_prices_for_date` for: pre-market mirror rejected → `PriceDataUnavailableError`; plain `tomorrow` used and VAT-stripped once `tomorrow_valid` is true; valid `raw_tomorrow` still works; plain `today` fallback strips VAT. (`test_all_scenarios` / `R == P` not applicable — this is price-source parsing, upstream of the DP.)

## Docs
- `docs/SOFTWARE_DESIGN.md` — updated the one line describing how the `nordpool` (HACS) runtime path reads prices. No other section in `bess-knowledge.md` or `SOFTWARE_DESIGN.md` describes `HomeAssistantSource` internals, VAT stripping, or the tomorrow fallback.

## Note for the reviewer
No runtime mock-HA scenario is affected: `ci-normal-day` / `ci-growatt-vpp` E2E run under `nordpool_official` (service-call path), which doesn't touch this code. `ci-wizard-nordpool-hacs` is wizard/discovery-only and never builds a schedule.

Refs #704


bess-agent:
@claude-bot review

Issue #680, opened 2026-08-22
Late switching


ridax67:
## What happened
The VPP switching happens quite late, 1.5 minute into the period?

Plus I wonder about idle 17.45-18.00.  It doesn't seem justified by price, but maybe it became real idle because previous periods might have been justified?

## Debug bundle
Please drag-and-drop **bess-debug-2026-08-22-182134.md** from your Downloads folder into the editor below before submitting.
The file contains your version, settings, recent logs, and the raw HA discovery dump that the maintainers need to reproduce the issue.

[bess-debug-2026-08-22-182134.md](https://github.com/user-attachments/files/31336493/bess-debug-2026-08-22-182134.md)

johanzander:
@claude-bot analyze

bess-product-owner:
Stage 2 diagnosed this (comment above): symptom (b) — the 17:45–18:00 idle — is correct behaviour (battery no longer full, small surplus absorbed rather than exported), no fix needed. Symptom (a) — the switch landing ~1.5 min into the period — is pipeline latency (the hardware write is the last step of a synchronous price→sensors→DP→apply cycle), not a deliberate delay, and the analysis could not pin down *why* that latency grows through the afternoon from a bundle with no DEBUG-level logging.

To take this further we need a **DEBUG-level debug bundle** captured across an afternoon/evening window where you see the late switching. In the add-on config set the log level to `debug`, let it run over a few period boundaries in the afternoon, then export and attach a fresh bundle here. Marking this as waiting on that.

bess-product-owner:
Closing based on the analysis above: both symptoms here turn out to be correct-as-designed rather than bugs.

- The ~1.5 min post-boundary delay is architectural pipeline latency (price fetch → sensors → DP solve → apply runs synchronously before the hardware write), not a deliberate or buggy scheduling delay.
- The IDLE decision at 17:45–18:00 was correctly price-justified for that slot — the earlier SOLAR_EXPORT forecast changed because the battery's carried-forward SOE genuinely dropped from prior profitable discharges that day.

No code defect was found. Reopen if this recurs with evidence pointing at an actual regression.

PR #716, opened 2026-08-29
fix: rhythm pass flags a stale Awaiting=analysis after Stage 2

johanzander:
## Problem

`Awaiting: analysis` on the board is a placeholder meaning "this item still needs Stage 2". Once Stage 2 runs it stamps `analyzed` (or `needs-human-review`) and removes `ready-for-analysis` — but nothing clears the **board** field, and `backlog-rhythm.sh` only emits `set_awaiting` when `awaiting_source == "label"` (field empty). An explicitly-set board value is trusted as a deliberate wait, so the item is silently counted into the Analysis column and suppressed from every action, indefinitely.

Found via the live board: **#680** and **#704** were both `Awaiting: analysis` + `analyzed`, analysed weeks ago, invisible to every rhythm pass. (Both reconciled by hand today.)

## Change

`scripts/backlog-rhythm.sh` — new `set_awaiting` trigger:

```
awaiting == "analysis" AND awaiting_source == "board"
  AND (labels has "analyzed" OR labels has "needs-human-review")
```

Ranks with the other grooming-debt actions (rank 6). Does **not** fire before Stage 2 (no label) — that value is then an accurate wait.

## Tests

`backend/tests/test_backlog_rhythm.py` — `test_stale_analysis_awaiting_after_stage2_is_grooming_debt` (fires) and `test_analysis_awaiting_before_stage2_is_not_flagged` (does not). `pytest backend/tests/test_backlog_rhythm.py backend/tests/test_backlog_digest.py` — 185 passed.

Skill doc `set_awaiting` row updated to match.


bess-agent:
@claude-bot review

